PR #14, opened 2024-07-25
Add instruction to run hub and agent as system service

MFYDev:
Hi, if you do not mind I added some instructions on how to run binaries in the background as system service. With this added it is more linux-newbie friendly.

Feel free to let me know if you think this is too much.

I can also add OIDC instruction for Authentik if you want me but that is mainly for PocketBase. Feel free to let me know.


MFYDev:
I also think maybe we can add this instruction as a button next to the docker compose generation.  The point is since some server or system is very light, installing mainly based on docker might be too heavy for those.

I can make it as a one line command including download, create service, start and enable. Should be very simple.

I can create a PR for this later this week as well if you agree with me

crypt0rr (approved):
Welcome instructions.

henrygd:
Good idea, I'll merge with a few updates a bit later.

It might be nice to add subcommands to the binaries for installing / uninstalling the service.

For example:

```bash
# hub
./beszel service install

# agent
./beszel-agent service install --port "..." --key "..."
```

Issue #105, opened 2024-08-08
Show agent versions on hub.


utlilb:
It would be nice to see the agent versions of each connected system.

henrygd:
Added in 0.2.0

hellofaduck:
I updated docker image of server-side and all agents on all hosts to 0.2.0, and still no agent version info...
![image](https://github.com/user-attachments/assets/cebb82d6-5c4e-4e01-ac5b-02684b1e8d80)

henrygd:
@hellofaduck Please double check that agents are updated to 0.2.0 and restart them.

If you use the binary and have a system service running, make sure to restart the service:

```bash
sudo systemctl restart beszel-agent
```

In the "systems" table in PocketBase you should see the agent version (`v`) being populated:

![image](https://github.com/user-attachments/assets/99a81692-8ab5-42b4-822e-c66f19bb5f88)

hellofaduck:
It seems that restarting of agents is worked, but strange that update command dont doing that automaticaly....

PR #109, opened 2024-08-10
Agent version display in dashboard

theRealBassist:
Creates a new column in the dashboard which displays the version of each running agent. As requested by #105 .

I would like to note that I have never used Go or contributed to any open source project before, so if there are any major faux pas which I have committed, I would love to know, so that I can learn more!


henrygd:
As with #111, unfortunately the timing is bad for a PR. But I will @ you when the restructuring is done if you want to submit this again.

theRealBassist:
@henrygd Read through both comments. Sounds good!

PR #110, opened 2024-08-10
Feature/change file structure

TOomaAh:


henrygd:
Thanks!

I only had a quick minute to look over it, but seems like a nice improvement! Should make it easier to add tests as well.

Just a couple things:

I'd like to avoid having lots of files pile up in the root directory, so perhaps we can make a folder called `beszel` or `app` in the root directory and move the application related code there.

Also, I noticed you accidentally committed an exe file. Sorry to ask you to do this, but I'd like to remove that from the git history. Not sure of the best way to do this. Maybe either `git reset --soft` and `git push --force`, or just making a fresh pull request.

Anyway, thanks again. I'll fiddle around with the workflows to make sure those will still work, then merge when I have a bit more time over the next few days.

TOomaAh:
> Thanks!
> 
> I only had a quick minute to look over it, but seems like a nice improvement! Should make it easier to add tests as well.
> 
> Just a couple things:
> 
> I'd like to avoid having lots of files pile up in the root directory, so perhaps we can make a folder called `beszel` or `app` in the root directory and move the application related code there.
> 
> Also, I noticed you accidentally committed an exe file. Sorry to ask you to do this, but I'd like to remove that from the git history. Not sure of the best way to do this. Maybe either `git reset --soft` and `git push --force`, or just making a fresh pull request.
> 
> Anyway, thanks again. I'll fiddle around with the workflows to make sure those will still work, then merge when I have a bit more time over the next few days.

I removed the executable from the git history.

I didn't understand the second point. Do you want me to remove the 'internal', 'cmd', etc... folders from the project root or do you want me to remove the 'Dockerfile...' files or both?

henrygd:
Sorry,  I mean both.

This is more personal preference than common practice, but I'd like for the root directory to have fewer files and clear separation of purpose, rather than having everything intermingled at the base.

So I'd prefer to have the same files at the root as the repo currently has, but instead of individual folders for hub / agent, we just have one folder named beszel, and the code for the app lives there.

We can keep the structure that you have currently, it just goes in the beszel folder instead of the root.

If this still isn't clear -- and I understand why it may not be -- I can make a commit tomorrow that organizes that way.

henrygd:
I made some updates and submitted a pr on your fork.

If you think of any other changes, you can merge and update. Otherwise I can just continue working on the branch I checked out already.

TOomaAh:
For my part, I won't go into any further modifications or code slicing to avoid creating bugs where none exist. 
Let me know when you want me to merge your modifications into the [https://github.com/TOomaAh/beszel/pull/1](PR) and I'll do it.

henrygd:
No worries, no need to merge if you're done. I'll close that pr and eventually merge this to main after I've had a bit more time to look it over and test workflows.

Issue #103, opened 2024-08-08
Docker Hosts - Agent reports system as down if any container is stuck in a boot loop


NightHawkATL:
Noticed this yesterday with 2 of my Docker VMs. Also got word from another Discord member that is running this as well that the same thing happened to him.

For me, I had an issue with the redis container for Tube Archivist that made it stick in a boot loop as it couldn't read the cache db after an update. Once I cleared the cache db and restarted the container, the Beszel agent started reporting the whole VM and all containers to be up. 

For my friend, he had an issue with cloudflared and was stuck in a boot loop. He stopped the container and the agent reported the VM as up again.

henrygd:
I'll look into it. Should be an easy enough fix.

theRealBassist:
> Noticed this yesterday with 2 of my Docker VMs. Also got word from another Discord member that is running this as well that the same thing happened to him.
> 
> For me, I had an issue with the redis container for Tube Archivist that made it stick in a boot loop as it couldn't read the cache db after an update. Once I cleared the cache db and restarted the container, the Beszel agent started reporting the whole VM and all containers to be up.
> 
> For my friend, he had an issue with cloudflared and was stuck in a boot loop. He stopped the container and the agent reported the VM as up again.

This was  exactly what was causing my problem in #57 . TubeArchivist had a Redis issue, and at the same time I was trying to get Beszel to work. When I fixed the TubeArchivist issue is when I restarted my VM and everything was working.

Issue #57, opened 2024-07-28
Failed to get server stats: unexpected end of JSON input


mrnico76:
Hello, 

I've a problem with beszel agent. I use it in a docker and I can't add the agent client to the dashboard, because of this error: Failed to get server stats: unexpected end of JSON input

I've also checked the agent's logs, and i've only noticed this error: Error getting container stats: mailcowdockerized-netfilter-mailcow-1 cpu pct greater than 100: 3.308239611496335e+10

henrygd:
What version of Docker on the agent system (`docker -v`)?

Try restarting Docker. `sudo systemctl restart docker` if on linux.

Is that error only with the one container? Approximately how many other containers are you running on the system?

mrnico76:
Docker version 27.1.1, build 6312585

I've restarted Docker and It worked until this container started "mailcowdockerized-netfilter-mailcow-1".

I think the error is only with this container and I've 55 containers running

henrygd:
Try 0.1.1 whenever you have a chance to upgrade and see if that fixes the problem. No rush.

mrnico76:
I've upgraded to 0.1.1, but It didn't fixed my problem. Now I have no error in the agent's docker logs, but I still have this error: Failed to get server stats: unexpected end of JSON input

henrygd:
Can you check that the API is returning valid JSON for that container?

Use `docker ps` to list containers, then copy the container's ID.

Then use curl to make a request to the API (replace `d07d19594b6e` with your container ID):
```
curl --unix-socket /var/run/docker.sock -H "Content-Type: application/json" "http://localhost/containers/d07d19594b6e/stats?stream=0&one-shot=1"
```

Post the response here please, or just compare with the response for another container ID to make sure everything is present.

theRealBassist:
Just want to pop in to say that I have the exact same issue. 

When I make an API request on the agent system in the manner you mentioned I get the following:

```
[...]

PR #111, opened 2024-08-10
Fix for edge case on checking age of docker container

theRealBassist:
There are a few small edge cases while checking if a container is possible restarting/has restarted where instead of the typical `Up 20 seconds` message in the status field, you will instead see something like `Up 20 seconds (health: starting)` or `Up 20 seconds (healthy)` where the strings.HasSuffix() method will not locate the "seconds" string even though it is present.

It might not be necessary to include the additional check for the actual "starting" string. However, in the event of a container boot looping like in issues #103 and #57, this might assist in filtering out errant data. However, I was not able to force a situation wherein I could actually test if this would work.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -175,8 +175,9 @@ func getDockerStats() ([]*ContainerStats, error) {
 		ctr.IdShort = ctr.Id[:12]
 		validIds[ctr.IdShort] = struct{}{}
 		// check if container is less than 1 minute old (possible restart)
+		// additionally check if container has status of starting
 		// note: can't use Created field because it's not updated on restart
-		if strings.HasSuffix(ctr.Status, "seconds") {
+		if strings.Contains(ctr.Status, "seconds") || strings.Contains(ctr.Status, "starting") {
 			// if so, remove old container data
 			deleteContainerStatsSync(ctr.IdShort)
 		}


henrygd:
Thanks for pointing this out!

Unfortunately it's the worst possible time for a pull request as we have an ongoing change in #110 to restructure the project.

It should be done in the next handful of days so I can ping you afterwards if you want to make another PR.

henrygd:
Restructuring is done now, so you can resubmit if you'd like.

To run locally you can do:

- `bun run dev` in `beszel/site`
- `KEY="..." go run .` in `beszel/cmd/agent`
- `go run . serve` in `beszel/cmd/hub` (you can also put an existing `beszel_data` folder in this directory)

If you don't want to go through the trouble, that's understandable and I can review / implement your changes from your closed PRs.

theRealBassist:
I should be able to get this issue fixed tomorrow while I'm at work. I can look back at adding the version display later tomorrow or maybe the next day!

Appreciate the ping back.

theRealBassist:
@henrygd Sorry to keep necroing this thread, but I'm running into an issue with running the hub with `bun run dev` wherein I just get locked onto a login screen? I'll include the logs from my browser and a screenshot below. I have the changes made, but I don't want to commit them until I can test it out first obviously.

I can put in a new issue if you need.

```
Uncaught (in promise) ClientResponseError 0: The request was autocancelled. You can find more info in https://github.com/pocketbase/js-sdk#auto-cancellation.
    _ClientResponseError ClientResponseError.ts:13
    send Client.ts:376
2 ClientResponseError.ts:13:8
```


Screenshot_20240819_140731

henrygd:
Not on a computer to test, but this should only happen if DISABLE_PASSWORD_AUTH env var is set and no auth providers are enabled.

That error is unrelated, it's just because React double mounts components in dev mode which causes the second fetch request to auto cancel.

I'll look into it later if you don't figure it out 👍

PR #124, opened 2024-08-19
Fixes edge cases where container health message does not have expected suffix

### LINKED ISSUE #103: Docker Hosts - Agent reports system as down if any container is stuck in a boot loop

theRealBassist:
Resubmit for #111

Once again, I'm not 100% sure the addition of the check for the "starting" string will necessarily fix #103, but it shouldn't hurt. The only downside is that containers which do not have a properly configured health check will not load data.

### DIFF
--- a/beszel/internal/agent/agent.go
+++ b/beszel/internal/agent/agent.go
@@ -190,8 +190,9 @@ func (a *Agent) getDockerStats() ([]*container.Stats, error) {
 		ctr.IdShort = ctr.Id[:12]
 		validIds[ctr.IdShort] = struct{}{}
 		// check if container is less than 1 minute old (possible restart)
+		// also check if container health is listed as starting to catch potential boot loops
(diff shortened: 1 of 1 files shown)


henrygd:
Thank you!

I fixed 103 earlier and it's unrelated to this section of code, so I changed the check to be only `strings.Contains(ctr.Status, "second")`.

This does fix an issue with the way I was doing it before, as I didn't account for the health status appended to the end. I appreciate you pointing that out!

PR #135, opened 2024-08-24
feat: style scrollbar

steveiliop56:
I simply changed the default scrollbar to have transparent background and use the border color to match beszel's theme. Here is a preview with 175% zoom:

image

And in light mode:

image

### DIFF
--- a/beszel/site/src/index.css
+++ b/beszel/site/src/index.css
@@ -2,85 +2,85 @@
 @tailwind components;
 @tailwind utilities;
 @layer base {
-	:root {
-		--background: 30 8% 98.5%;
-		--foreground: 30 0% 0%;
-		--card: 30 0% 100%;
-		--card-foreground: 240 6.67% 2.94%;
-		--popover: 30 0% 100%;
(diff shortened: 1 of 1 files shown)


henrygd:
Good idea, thanks. I may tweak it a bit for contrast and give it a min-width to prevent it overriding mobile scrollbars.

steveiliop56:
Yeah no problem! My ocd was just dying with the default scrollbar lol

Simonh2o:
I'd recommend only changing the webkit scrollbars for windows/chrome OS combo and not for mac/linux/phones which utilizes floating scrollbars (prevents scrollbar from taking up space in dom). Setting the color-scheme to dark would be enough for windows OS so you get rid of the light scrollbar during dark mode. Just my opinion.

steveiliop56:
It's yours man do whatever you want lol.

henrygd:
@Simonh2o Good point.

After doing some research, it looks like setting the property below when dark mode is enabled will apply a darker theme to the native scrollbars.

```css
color-scheme: dark;
```

So I think I'll go this route instead of overriding with css.

Issue #71, opened 2024-07-29
[Feature Request] Webhook alerts / push notifications


henrygd:
Please see #49 to discuss webhook alerts.

Keywords for search: Ntfy, Discord, Pushover, Slack, Mattermost, Telegram, Apprise, Gotify, Whatsapp

henrygd:
Added in 0.4.0

PR #145, opened 2024-09-01
Add support for alerts via shoutrrr

parnic:
This provides users the ability to use a wide variety of notification platforms instead of just email. If there's a problem sending a notification via shoutrrr, an error is logged and email is attempted as a fallback.

Since this uses Viper, users can set a notification type and URL via either a config file or environment variable. In the beszel_data folder (where the sqlite dbs reside), create an alerts.env file to set values that way.

Values:
* NOTIFICATION_TYPE
  * If this is `shoutrrr`, then the shoutrrr library is used. Any other value, including not being set, uses the fallback email behavior.
* NOTIFICATION_URL
  * If NOTIFICATION_TYPE is shoutrrr, this is the URL given to shoutrrr to send the alert. See list of supported services: https://containrrr.dev/shoutrrr/services/overview/

Note: there's currently a bug in viper v1.18.2+ where environment variable overrides aren't functioning when no config file exists, so this library should remain pinned to 1.18.1 until that's fixed. See: https://github.com/spf13/viper/issues/1895

Fixes https://github.com/henrygd/beszel/issues/71


henrygd:
Thanks!

shoutrrr does look like a great option for this, assuming it doesn't add much overhead.

However, I want to allow users to add their own notification URLs. So I'd rather build the config into the web UI than pull global values from a file.

I'm working on adding support for multiple disks and should have that release out in the next few days. As soon as that's finished, I'll pull your changes and test it out.

If it looks like it's going to work, I'll merge this and adapt it for the following release.

parnic:
Sounds perfect. I'd love to have this support in my current docker installation of Beszel, which has been awesome so far. I agree some UI configuration would be preferred, but this felt like a first step toward that.

Thanks for taking a look!

henrygd:
Finally pushed this out. There's a settings page now where you can add shoutrrr URLs.

Thanks again for your contribution :+1:

PR #209, opened 2024-10-17
fix: path escape for system name in alert message

Yukino16:
characters like space would be escaped to `+` instead of `%20`, this makes the url in alert message invalid if system name contains space


henrygd:
Thanks :+1:

PR #345, opened 2024-12-27
fix: json.Unmarshal not resetting struct fields

chuangbo:
### Description

`json.Unmarshal` was reusing the same `Stats` struct, leading to incorrect values when unmarshaling multiple JSON strings. This commit creates a new `Stats` instance in each loop iteration to resolve the issue.

A reproducible example is available on Go Playground: https://go.dev/play/p/HnZ8-z7hgXq

### DIFF
--- a/beszel/internal/records/records.go
+++ b/beszel/internal/records/records.go
@@ -155,6 +155,7 @@ func (rm *RecordManager) AverageSystemStats(records RecordStats) system.Stats {
 
 	var stats system.Stats
 	for i := range records {
+		stats = system.Stats{} // Zero the struct before unmarshalling
 		json.Unmarshal(records[i].Stats, &stats)
 		sum.Cpu += stats.Cpu
 		sum.Mem += stats.Mem


henrygd:
Thanks for catching this problem!

Correct me if I'm wrong, but this only affects fields that use `omitempty` - so `MemZfsArc`, `Swap`, and `SwapUsed`. And only if those values exist, then decrease to zero, right?

I agree with your solution, but I'd still like to reuse the `Stats` struct for memory efficiency. Instead of creating a new variable in the loop, we can zero the Stats struct instead.

Example: https://go.dev/play/p/K6FnQgWmuJU

Can you please update the PR to use this method instead?

chuangbo:
Thank you for the reply! I tested it again, and the problem happens even without `omitempty`. It seems `omitempty` doesn't affect this issue.

Here's an example: https://go.dev/play/p/xriooE6fFf2

As the documentation says, `emitempty` only affects how json is encoded:

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.
https://pkg.go.dev/encoding/json

About memory performance, I created a benchmark here https://gist.github.com/chuangbo/85c4baf315f79f60a96594b682e661fd. It shows there's little difference between reusing or creating a new Stats variable in the loop.

Please let me know if I'm mistaken.

chuangbo:
PR has been updated as suggested.

henrygd:
Merged, thanks :+1:

PR #412, opened 2025-01-15
adding reading agent ssh key from file

beny2000:
This PR adds support for reading the agent SSH key from a file. This allows storing the agent key as a docker secret. 

I'm not a Go dev so let me know if there's a better way to do this but it's pretty straightforward

### DIFF
--- a/beszel/cmd/agent/agent.go
+++ b/beszel/cmd/agent/agent.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"strings"
+	"io/ioutil"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


beszel/cmd/agent/agent.go
```diff
@@ -25,7 +26,16 @@ func main() {
...
-		log.Fatal("KEY environment variable is not set")
+		keyFile := os.Getenv("KEY_FILE")
+		if keyFile != "" {
+			if keyData, err := ioutil.ReadFile(keyFile); err == nil {
+				pubKey = keyData
+			} else {
+				log.Fatalf("Failed to read key from file '%s': %v", keyFile, err)
+			}
+		} else {
+			log.Fatal("KEY environment variable is not set, and KEY_FILE environment variable is not set")
```
devil1234, line 37: I think should be changed to : KEY environment variable is not set or KEY_FILE environment variable is not set

henrygd:
Merged, thanks :+1:

beszel/cmd/agent/agent.go
henrygd: I changed it to "Must set KEY or KEY_FILE environment variable." Slightly more concise.

Issue #232, opened 2024-10-28
agent ssh server starts on ipv6


MarkusGnigler:
i have a docker deployment of the agent in a regular ipv4 network and the ssh server starts only on ipv6.
i does not have anything configure for ipv6 and i have set the the `PORT` variable to `0.0.0.0:45876`

```bash
╰─λ netstat -tulpn | grep 45876
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
tcp6       0      0 :::45876                :::*                    LISTEN      -
```

This obviously means i cant reach my agent.

```bash
╰─λ telnet 10.0.0.15 45876
Trying 10.0.0.15...
telnet: connect to address 10.0.0.15: No route to host
```

Is it somewhat related to the start process?

https://github.com/henrygd/beszel/blob/ec7cb53d935b7d07ed1fbf8fd0893856e75c3981/beszel/internal/agent/server.go#L15

henrygd:
Maybe try `10.0.0.15:45876`, or just `45876`.

If that doesn't work, please paste your docker compose here.

MarkusGnigler:
Thanks for the fast reply.

I have started with the `PORT` variable set only to 45876 as from the example docker compose file.

Kind of fixed it for now by removing the `network_mode: host` and using regular port mapping.

This is my docker compose file:

```yml
services:
  beszel-agent:
    image: "henrygd/beszel-agent"
    container_name: "beszel-agent"
    restart: unless-stopped
    ports:
      - 45876:45876
#    network_mode: host
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      # monitor other disks / partitions by mounting a folder in /extra-filesystems
      # - /mnt/disk1/.beszel:/extra-filesystems/disk1:ro
    environment:
      PORT: 0.0.0.0:45876
      KEY: "XXX"
      # FILESYSTEM: /dev/sda1 # override the root partition /

```

Any guess whats the problem here? Network monitoring would be nice.

henrygd:
I'm not really sure. This is the first time this behavior has been mentioned. My guess is that it has to do with network configuration on the host machine.

The address:port is passed directly to the [gliderlabs/ssh](https://github.com/gliderlabs/ssh) [`ListenAndServe`](https://github.com/gliderlabs/ssh/blob/master/server.go#L342) method, as you found earlier.

[...]

PR #433, opened 2025-01-18
Added proc/net fake

MarkusGnigler:
Added fakenet to have a copy of `proc/net` inside the agent container.
Used the concept from [netdata](https://scene-si.org/2016/11/22/bypassing-docker-network-isolation-hack/) to simulate `proc/net` within `/dev/shm/fakenet`.

#232

I copied the bash script for the copy function directly from [Script]( https://raw.githubusercontent.com/titpetric/netdata/master/fakenet.sh).

In the docker file i have mounted a shared named volume between the agent and the copy container.

Added:
- Fakenet dockerfile and script to `/beszel` folder
- Docker image build in the `docker-images` action
- Compose file extended in `add-system.tsx` and `/supplemental` standalone agent

Missing:
- Extending compose file in `/supplemental` for `same-system` agent


henrygd:
Thanks, I'll get back to you on this tomorrow.

MarkusGnigler:
Please take your time. Other issues are more important i think :)

MarkusGnigler:
If we can trim it down to just */proc/net/dev*, no extra docker image is necessary. But that's a matter of taste.

What do you think?

```yml
---
volumes:
    procnet: {}

services:
  fakenet:
    image: alpine
    restart: unless-stopped
    network_mode: host
    entrypoint: /bin/sh
    command:
      - -c
      - |
        export OUTPUT="/dev/shm/fakenet/"
        export SOURCE="/proc/net/dev"
        mkdir -p $OUTPUT/dev
        while [ true ]; do
            OUTFILE="$${SOURCE:10}"
            echo "$$(<SOURCE)" > $$OUTPUT$$OUTFILE
            cat $$SOURCE
            sleep 0.23
        done
    volumes:
      - procnet:/dev/shm/fakenet
```

henrygd:
> If we can trim it down to just /proc/net/dev, no extra docker image is necessary. But that's a matter of taste.

I love this, good idea! It allows people see what's happening rather than just running a black box process. 

My general idea was to write a little C program to do this. The benefit there is that it would use practically no resources. Probably something like 500 KB memory or less.

I want to play around with both approaches, and maybe offer both depending on user preference. Didn't have time to get into it today, but will as soon as I can.

kevinschweikert:
For the alpine image there's one env variable ($OUTPUT) still unescaped. The correct spec would be:

```yaml
services:
  fakenet:
    image: alpine
    restart: unless-stopped
    network_mode: host
    entrypoint: /bin/sh
    command:
      - -c
      - |
        export OUTPUT="/dev/shm/fakenet/"
        export SOURCE="/proc/net/dev"
        mkdir -p $$OUTPUT/dev # <<<<< this neeed to be double escaped
        while [ true ]; do
        OUTFILE="$${SOURCE:10}"
        echo "$$(<SOURCE)" > $$OUTPUT$$OUTFILE
        cat $$SOURCE
        sleep 0.23
        done
    volumes:
      - procnet:/dev/shm/fakenet

volumes:
  procnet:
```

My setup looks like this with a [netbird](https://netbird.io/) container to connect them to a VPN network. That way i cannot use the `network_mode: host`. 

```yaml
services:
  beszel:
    image: henrygd/beszel:latest
    container_name: beszel
    restart: unless-stopped
    volumes:
      - ./beszel_data:/beszel_data
    network_mode: service:netbird

  netbird:
    image: 'netbirdio/netbird:latest'
    container_name: beszel-netbird
    restart: unless-stopped
    cap_add:
      - NET_ADMIN
    environment:
      - NB_SETUP_KEY=...
    volumes:
      - ./netbird-data:/etc/netbird

  beszel-agent:
    image: henrygd/beszel-agent:latest
    container_name: beszel-agent
    restart: unless-stopped
    network_mode: service:netbird
    volumes:
      - procnet:/dev/shm/fakenet
    environment:
      PORT: 45876
      # Do not remove quotes around the key
      KEY: ...

  fakenet:
    image: alpine
    restart: unless-stopped
    network_mode: host
    entrypoint: /bin/sh
    command:
      - -c
      - |
        export OUTPUT="/dev/shm/fakenet/"
        export SOURCE="/proc/net/dev"
        mkdir -p $$OUTPUT/dev
        while [ true ]; do
        OUTFILE="$${SOURCE:10}"
        echo "$$(<SOURCE)" > $$OUTPUT$$OUTFILE
        cat $$SOURCE
        sleep 0.23
        done
    volumes:
      - procnet:/dev/shm/fakenet

volumes:
  procnet:
```

henrygd:
I've been experimenting with this over the last few days.

The copy idea from Netdata works, and I love that everything can be included in the `command` option, but unfortunately it's not efficient. I'm seeing constant CPU usage of around 1.5% on a Ryzen 7 5700U, and 99% of the work the process does is unnecessary since we only need the data once per minute.

Instead I think we should do IPC via unix domain socket or pipes to get the content on demand. I'll try to get it working this evening.

MarkusGnigler:
Thanks, for investing your time!

Sad to hear that's too resource hungry.
I'm not really comfortable with IPC sry.

Should i abandoned this PR?

henrygd:
Yeah I'll just close it. But don't worry I'll have this working soon.

MarkusGnigler:
Take your time.
Thank you for your kindness and hard work!

PR #445, opened 2025-01-18
Add OpenWRT support for binary quick script installation

Nicols0Mart:
Modified the install-agent.sh script for binary installation to support OpenWRT.
I more or less tried to match your coding style, checking the OS type like you did with Alpine. 

To add the service, instead of systemd a procd init script is used. The service can be enabled, started, stoped, etc. by using 
`service beszel-agent ...`. 
An update command is also included to update and restart the service. For automatic updates, I opted to use a crontab job that uses `service beszel-agent update` once a day. 

To get the OS type variable, `sed`  is used instead of  `tr`, since most busybox systems have implmentations of `tr`  that do not support classes, as mentioned in [this issue](https://github.com/dehydrated-io/dehydrated/issues/285). This implementation is not the pretiest, since all 27 alphabet characters have to be included, but the upside is that this should support almost any linux system.

### DIFF
--- a/supplemental/scripts/install-agent.sh
+++ b/supplemental/scripts/install-agent.sh
@@ -5,6 +5,10 @@ is_alpine() {
   [ -f /etc/alpine-release ]
 }
 
+is_openwrt() {
+  cat /etc/os-release | grep -q "OpenWrt"
+}
+
 version=0.0.1
 # Define default values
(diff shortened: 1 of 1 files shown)


henrygd:
Looks good, thank you! :+1:

PR #464, opened 2025-01-21
add mac os binary agent install

sntxrr:
- adds mac binary agent install  (untested)

### DIFF
--- a/supplemental/scripts/install-agent.sh
+++ b/supplemental/scripts/install-agent.sh
@@ -8,6 +8,10 @@ is_openwrt() {
   cat /etc/os-release | grep -q "OpenWrt"
 }
 
+is_macos() {
+  [ "$(uname -s)" = "Darwin" ]
+}
+
 # Define default values
 PORT=45876
 UNINSTALL=false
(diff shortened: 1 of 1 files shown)


henrygd:
Thanks, unfortunately I can't test this because I don't have a mac.

I'll see if there's a cloud option I can try.

svenvg93:
@henrygd I got a M1 Macbook, if you want I can give it a try?

henrygd:
@svenvg93 Sure, that would be great!

The changes look fine to me, but I'm not a Mac expert, so you might want to check the diff yourself before running on your machine.

By the way, do you have an agent running on your Macbook currently? I'm curious because I'd really like to figure out the most reliable way to set it up.

I made a discussion post (#276) and got some helpful responses, but nothing definitive yet.

owldown:
Homebrew is a popular package manager for MacOS, and also runs on Linux. It is possible to create packages that install easily and don't clobber existing software. The packages can be binary or source code, in which case it still compiles locally, but without so much typing or cloning, and the binary is then installed in a way that it shows up in the PATH. 
I don't know all the details, but there are instructions here: https://docs.brew.sh/Formula-Cookbook on how to create a formula for the agent that could be installed easily on MacOS by the user typing "brew install beszel". I think you could test the formula on a linux system.

henrygd:
@owldown Interesting, that may be a better option. If we can set it up properly with brew then we could just run the necessary brew commands in the script.

svenvg93:
> @svenvg93 Sure, that would be great!
> 
> The changes look fine to me, but I'm not a Mac expert, so you might want to check the diff yourself before running on your machine.
> 
> By the way, do you have an agent running on your Macbook currently? I'm curious because I'd really like to figure out the most reliable way to set it up.
> 
> I made a discussion post (#276) and got some helpful responses, but nothing definitive yet.

I'm not an expert either in Mac development  😅 but the scripts fails.

```
curl -sL https://raw.githubusercontent.com/sntxrr/beszel/refs/heads/main/supplemental/scripts/install-agent.sh -o install-agent.sh && chmod +x install-agent.sh && ./install-agent.sh -p 45876 -k "ssh-ed25519 REMOVED"
./install-agent.sh: line 663: syntax error: unexpected end of file
```

Weirdly there are only 662 lines in the script 🤔

henrygd:
Thanks for trying anyway!

The info that @owldown and @zph provided about brew is really helpful and it definitely looks like the best option.

I'll experiment when I have time. It would be great if we can set it up directly through our GoReleaser config.

sntxrr:
Truth be told, I'm not expert either. I'm closing this in favor of a `homebrew` style install.  Thanks for considering it at least!

henrygd:
@sntxrr No worries, at least it put us on the right track! :+1:

PR #480, opened 2025-01-23
feat(release): add riscv agent builds to goreleaser config

silven:
This change adds a riscv build config to for the agent to goreleaser. As far as I am aware, only freebsd and linux supports this, so ignore mac and windows.

Verified to work on a [VisionFive 2 SBC](https://www.starfivetech.com/en/site/boards).

I have not updated any installation scripts as that looks like witchcraft to me.

### DIFF
--- a/beszel/.goreleaser.yml
+++ b/beszel/.goreleaser.yml
@@ -35,11 +35,16 @@ builds:
       - arm64
       - arm
       - mips64
+      - riscv64
     ignore:
       - goos: freebsd
         goarch: arm
       - goos: windows
         goarch: arm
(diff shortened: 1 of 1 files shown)


silven:
BTW: I really like this software. Nice work, I just needed a RISC-V binary for one of my SBCs.

henrygd:
Thanks! :+1: 

I do need to refactor the install script at some point. It's gotten a bit messy as support for more platforms has been added.

If you use systemd and `uname -m` returns `riscv64`, I think the script will be compatible. If not, let me know.

Issue #449, opened 2025-01-19
Display GPU in agent overview/summery


TomerGamerTV:
Can you add the option to view GPU load in the overview too? Because for me it's more important than my disk filling up, also while at it making the view customizable might be a good idea.

![image](https://github.com/user-attachments/assets/ddce5a88-0924-497e-89d2-99e18dd35417)

SonGokussj4:
That would be awesome. But there is an interesting design choice if there is more than 1 gpu... For example few of out servers have 3 GPUs. How would you shaw that in a table. As a 3 new custom rows behind each instance? Or as a subcard in the card view? I would be interested in this but I too would like to have this info in the general page :-)

TomerGamerTV:
> That would be awesome. But there is an interesting design choice if there is more than 1 gpu... For example few of out servers have 3 GPUs. How would you shaw that in a table. As a 3 new custom rows behind each instance? Or as a subcard in the card view? I would be interested in this but I too would like to have this info in the general page :-)

I this the better option will be hiding all the other GPUs behind the first one, and if the user wants to see everything he will click an arrow to pull them down.

The website will remember the setting and next the the person refreshes the website it will already be pulled down

PR #485, opened 2025-01-24
Adding GPU overview to the Overview Page

PTriebold:
I've added an Overview for the utilisation and Memory usage to the Overview page. Similar to #449   

The modification is only on the Hub side.  

In the info object there is now the latest stats object as well, there is a GPU number field and the GPUData object contains a memory percentage field.  

Preview:
Screenshot 2025-01-24 at 13 50 07


henrygd:
Thanks! I appreciate the work you did here, but I already have a local branch where I added temp and gpu to the table. It should be in the next release.

Right now it will only show the highest gpu load if you have multiple gpus, but maybe I'll take some inspiration from your screenshot and change to include multiple gpus in the grid view.

PR #510, opened 2025-01-31
Add SLURM cgroups support to system.go

MichaelAronsen:


henrygd:
Thanks! Adding support for cgroups is a big help, but we might want to separate this functionality into its own package and use it only if the system supports reading from cgroup.

Looks like a lot of stuff was removed here that we need to keep (sensors, gpu, etc.).

I need to work on some high priority things for the next release but I'll get back to you after that. Probably next week.

MichaelAronsen:
Yeah, this code was a proof of concept to see if it would even work.

If I can do more to help, do let me know, I want to use this to run per job node monitoring in hpc clusters.

tecosaur:
I wonder if this could be used to add support for monitoring Systemd services as well as containers?

tecosaur:
I wonder if this could be used to add support for monitoring Systemd services as well as containers?

MichaelAronsen:
> I wonder if this could be used to add support for monitoring Systemd services as well as containers?

I think so yes, as those are generally started within their own cgroup.
So that would enable beszel to monitor just the resources assigned to such services and containers.

tecosaur:
For what it's worth, this would catapult Beszel from "almost useless" to "really good" for me — since I use a NixOS server and most applications are managed as nix-provided systemd services instead of containers.

tecosaur:
Friendly poke @henrygd :slightly_smiling_face: 

As mentioned above, this looks really cool to me due to the capability to monitor systemd services. If you are able to spare the time to give this a good look, that would be great!

henrygd:
Thanks, if the feature can be more isolated (like in its own package) then I should be able to get to it more quickly.

I've had less spare time this year and more requests for support, so I'm behind where I hoped I'd be as far as development. We will get there eventually.

tecosaur:
> I made a new version that i'm testing out, if it works well, i'll make another Pull Request.

Hey @MichaelAronsen, just a gentle poke, it would be cool to hear how your revised effort went :slightly_smiling_face:

PR #522, opened 2025-02-03
feat: update password frontend validation

### LINKED ISSUE #521: [Bug]: Not able to login into backend (PocketBase) after Password change
### Description

I've got problems with creating a Superuser on first login, I think i chose a too long password with my 1password (100 chars) which was chopped off somewhere. So I reinstalled and chose a stupidly short password to change it later in the backend.

This worked (i used the password change feature) and i changed it to a 25 char password. After that, I was able to Login to the normal Beszel frontend with the new password. But when trying to enter the backend, neither the short nor 25-char password are working.

### Expected Behavior

- [ ] After a password change, the password should work in the Front- and Backend.
- [ ] When trying to register with an invalid password, the registration should fail.

### OS / Architecture

linux x64 via Coolify

### Beszel version

0.9.1

### Installation method

Coolify

ompathak2004:
fixes  #521

### DIFF
--- a/beszel/site/src/components/login/auth-form.tsx
+++ b/beszel/site/src/components/login/auth-form.tsx
@@ -14,7 +14,7 @@ import { Trans, t } from "@lingui/macro"
 
 const honeypot = v.literal("")
 const emailSchema = v.pipe(v.string(), v.email(t`Invalid email address.`))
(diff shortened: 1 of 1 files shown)


henrygd:
Thanks :+1:

PR #534, opened 2025-02-06
feat(AddAgent): added docker run command copy

Venipa:
added an option to use `docker run` command instead of docker compose.

image


henrygd:
Thanks! :+1:

PR #537, opened 2025-02-06
Initial beszel-hub k8s chart

dnikoloski:
Beszel-hub k8s chart is ready to deploy :rocket:


henrygd:
Thank you! :pray: 

Is there anything else you recommend I do for this?

Should we set up a domain to serve the charts?

Should we try setting up a github action to update them automatically on new releases?

Can this go in `/supplemental`? Like `supplemental/kubernetes/charts`? Or do we need a new root directory?

Sorry for so many questions,  I don't know much about how this works.

henrygd:
Thanks, before I merge would you mind adding another commit to move this inside `/supplemental`?

Like instead of `beszel-kubernetes/beszel-hub` we could do `supplemental/kubernetes/beszel-hub` or whatever structure you think works best.

I'll look into the github action when I have time, probably after the next release.

PR #539, opened 2025-02-08
Add Proxmox VE stats for VMs and LXCs

morgendagen:
Fixes #281  

Some environment variables must be added to the agent (I will create a separate PR for beszel-docs):

- `PROXMOX_URL`: url to PVE api, defaults to https://localhost:8006/api2/json
- `PROXMOX_NODE`: name of proxmox node to monitor
- `PROXMOX_TOKENID`: API token id, e.g. `root@pam!beszel`
- `PROXMOX_SECRET`: API token secret

I have not looked in to minimum requirements for the API token permissions, and am currently using one with admin access to `/`. I will add minimum requirements to the documentation when I have figured them out.
 
Proxmox server certificate vertification is currently **skipped**. Maybe it would make sense to add this as an option using an environment variable?

PR #542, opened 2025-02-09
Rename Go module to `github.com/henrygd/beszel`

IOAyman:
I've noticed that Beszel isn't currently installable using `go get`. This PR proposes renaming the Go module from `beszel` to `github.com/henrygd/beszel`. This would make it much easier for developers to use Beszel as a library within their own Go projects with a simple `go get github.com/henrygd/beszel`.
  
I've made the necessary changes, but I'm open to feedback and happy to refine them further.

### DIFF
--- a/beszel/Makefile
+++ b/beszel/Makefile
@@ -27,10 +27,10 @@ build-web-ui:
 	fi
 
 build-agent: tidy
-	GOOS=$(OS) GOARCH=$(ARCH) go build -o ./build/beszel-agent_$(OS)_$(ARCH) -ldflags "-w -s" beszel/cmd/agent
+	GOOS=$(OS) GOARCH=$(ARCH) go build -o ./build/beszel-agent_$(OS)_$(ARCH) -ldflags "-w -s" github.com/henrygd/beszel/cmd/agent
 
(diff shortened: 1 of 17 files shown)


henrygd:
I agree, but unfortunately the Go code is in the `/beszel` directory instead of the root, so we'd need to use `github.com/henrygd/beszel/beszel` instead of `github.com/henrygd/beszel`. Or maybe I could set up a vanity URL like `pkg.beszel.dev` that points to the subdirectory.

Not the greatest decision on my part. I didn't have a good understanding of how this worked when I started the project.

It's also not properly configured for use as a package since the agent and hub are internal.

I'll fix this, but it will be after the next release and after #539 is sorted out because I don't want to wreck that PR by changing the directory structure.

henrygd:
That's probably what we'll do, I'd just like get the next release out and the other PR merged before changing the structure.

This PR can stay open and I'll likely merge it as part of that process later in the week or next week.

IOAyman:
Thank you both for your input 🙏🏼 I didn't have a good understanding of how this worked as well, but I'm learning and am happy to help.

henrygd:
Thanks, sorry it took so long to get to this. I did merge, but cherry picked the commit, so github didn't track it. I'll close the PR but thanks for your efforts :+1:

PR #565, opened 2025-02-13
Update add-system.tsx docker-compose.yml template.

kmcfate:
Adding `version: "3"` makes this compatible with more recent docker tools.

Otherwise get error:

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
Unsupported config option for services: 'beszel-agent'
```

### DIFF
--- a/beszel/site/src/components/add-system.tsx
+++ b/beszel/site/src/components/add-system.tsx
@@ -61,6 +61,7 @@ export const SystemDialog = memo(({ setOpen, system }: { setOpen: (open: boolean
 
 	function copyDockerCompose(port: string) {
 		copyToClipboard(`services:
+  version: "3"
   beszel-agent:
     image: "henrygd/beszel-agent"
     container_name: "beszel-agent"


henrygd:
Thanks :+1:

jpbaril:
Funny because "version" is now obsolete in recent docker releases: https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

henrygd:
Yeah, I didn't realize that happened. Docker now gives this message: "the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion"

So I'll probably remove it.

jpbaril:
> So I'll probably remove it.

Well, in the foreseeable future you could keep it as it seems that per kmcfate in some cases it will not work at all if absent whereas in our case it will simple give a warning in the logs as you saw.

henrygd:
Kind of a lose/lose situation. If we include it then people will likely make PRs or bug reports asking to remove it because Docker prints a warning that says "please remove it."

From a quick look at some larger projects it seems most don't include the `version` attribute. 

If it's not in the spec anymore then I'm not sure that we should include it. From Docker's perspective it seems clear that they want it gone. Including it will continue to make people think it does something.

Issue #570, opened 2025-02-14
Beszel Agent 32Bit - Windows 7 32BIT - Beszel Agent will not Start using NSSM


kenrmayfield:
@henrygd @ArsFy @svenvg93 @mathyvds @a-mnich 

### Description

I switched to Ubuntu 21.04 and was able to Install GO v1.24.

I was able to Create the Agent for Windows.  

I Created a 64Bit at First and then went back and Created a 32Bit since I was using a 32Bit Windows OS.

When I Test I always start at Windows 7 and Move Up Windows OS Versions.

**Compile Command:** GOOS=windows GOARCH=386 CGO_ENABLED=0 go build -ldflags "-w -s" .

**I have some what Setup the Agent in Windows 7 however I am getting a Error once I Run:**
nssm start beszelgagent

**The Error:**
beszelagent: START: The service cannot be started, either because it is disabled or because it has no enabled devices associated with it.

**I used the Article:**
https://blog.ktz.me/using-beszel-to-monitor-windows/

**Windows Beszel Agent 32BIT:**
[agent32bit.zip](https://github.com/user-attachments/files/18793174/agent32bit.zip)

### Expected Behavior

nssm start beszelgagent to Start the Beszel Agent.

### Steps to Reproduce

**I used the Article:**
https://blog.ktz.me/using-beszel-to-monitor-windows/

### OS / Architecture

Windows 7 x86

### Beszel version

0.9.1

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

kenrmayfield:
@henrygd 

I see you Add My Issue to the Roadmap.   Thank You.

However I do have a Specific Question when I Compiled for for 32Bit and Ran the Agent on Windows 7.

nssm start beszelgagent

**The Error:**
beszelagent: START: The service cannot be started, either because it is disabled or because it has no enabled devices associated with it.

**What do I need to try to get the NSSM Service to Work with the 32Bit Agent?**

**Why do I have to wait for a Roadmap and Backlog?**

**Windows 32Bit Agent:**

[agent32bit.zip](https://github.com/user-attachments/files/18824471/agent32bit.zip)

henrygd:
Please don't tag contributors for support. I'm the only maintainer right now.

I'll try to get some of the Windows issues ironed out over the next few minor releases, as I said via email.

[...]

PR #581, opened 2025-02-15
add beszel-agent startup arguments

a-mnich:
I'm running the beszel-agent on Windows as a permanently running scheduled task in system context to avoid needing alternative programs like NSSM for running it in the background. As the Windows task scheduler provides support for supplying startup arguments I prefer to directly supply key and port via the commandline to be able to manage everything at one place.  

Changes:

* Added support for `-key` and `-port` flags to allow users to specify the public key and port number directly from the command line.
* Updated the usage information to provide a clear guide on available options and subcommands.
* Modified the logic to prioritize the `-key` flag over the `KEY` environment variable and the `-port` flag over the `PORT` environment variable.
* Enhanced error messaging to include instructions for using the `help` subcommand when the key is not provided.
* changed the version command to `version` instead of `-v` to allign with the other command like `help` and `update`

### DIFF
--- a/beszel/cmd/agent/agent.go
+++ b/beszel/cmd/agent/agent.go
@@ -3,6 +3,7 @@ package main
 import (
 	"beszel"
 	"beszel/internal/agent"
+	"flag"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


henrygd:
Thanks!

I need to refactor this section anyway as we're deprecating `PORT` in favor of `ADDR`. I'll merge this before I do that, probably tomorrow.

We'll likely go with `-addr` instead of `-port`, but no further changes are necessary to the PR :+1:

henrygd:
@gedw99 This has been added. In the next release you can optionally prefix all env vars with `BESZEL_HUB_` or `BESZEL_AGENT_`.

henrygd:
Merged, thanks :+1:

kenrmayfield:
@a-mnich 

Can you provide the Steps for Settting Up the Beszel Agent in the Windows Task Scheduler?

a-mnich:
> @a-mnich
> 
> Can you provide the Steps for Settting Up the Beszel Agent in the Windows Task Scheduler?

See #570

Chinoman10:
@a-mnich could I bother you to submit a new PR that does the same thing, but for:
- `EXTRA_FILESYSTEMS`?
  - Although I guess the 'disk stats' part (_executing a diagnostics program/dependency and reading/interpreting the output_) is most likely different on Linux and Windows machines... so a deeper integration that detects which type of OS it's running on would be needed...
- `LOG_LEVEL`? This one hopefully isn't harder than what was already done on this PR.
- `KEY_FILE`? Same comment as the one above.

@henrygd would be great if we could specify an .env file path as an arg. perhaps, instead of `KEY_FILE` solely being used for keys... That way we don't need to keep manually exposing 'expected env. vars' as command args, and can simply point to a file that's on the same folder as the `agent.exe`?

henrygd:
@Chinoman10 @a-mnich  I'll take care of it, no need for a PR.

I also saw your comment about env vars not being picked up through nssm. I will look into that soon.

I may add an option for everything to be configured though yaml as that will be easier to extend for future features like service monitoring.

kenrmayfield:
@henrygd 

`I may add an option for everything to be configured though yaml as that will be easier to extend for future features like service monitoring.`

Some User don't want to use Docker to Setup Beszel they Prefer the Binary way.

Issue #84, opened 2024-08-02
[Feature Request] pwa version


rolikeusch:
any plans to build a pwa version of the website? more info about pwa can be found [here](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/What_is_a_progressive_web_app).

i‘m a heavy mobile user and it would be amazing to have a fullscreen webpage on mobile.

henrygd:
There's now a PR for this: #651

As I said there, I'm a bit paranoid about footguns when it comes to PWAs, so any input from people who know what they're doing would be appreciated.

svenvg93:
It seems this is already possible looking at https://github.com/henrygd/beszel/pull/651

PR #651, opened 2025-03-06
Add barebones serviceworker setup to allow installing the app

dhvcc:
There's a better way to really do a nice PWA for this app especially with react, but for now this works nice for me. The only thing it needs to trigger is the "install" dialog, especially "adding to home" with safari no longer treats this as a link

### DIFF
--- a/beszel/site/index.html
+++ b/beszel/site/index.html
@@ -7,6 +7,16 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<title>Beszel</title>
 		<script>window.BASE_PATH = "%BASE_URL%"</script>
+    <script>
+      if ("serviceWorker" in navigator) {
+        navigator.serviceWorker.register("./static/serviceWorker.js")
+          .then(function(registration) {
+            console.log("ServiceWorker registration successful with scope: ", registration.scope)
+          }).catch(function(err) {
+            console.log("ServiceWorker registration failed: ", err);
+          })
+      }
+    </script>
 	</head>
 	<body>
 		<div id="app"></div>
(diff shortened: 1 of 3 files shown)


henrygd:
Thanks!

Unfortunately I haven't implemented a service worker in many years after a bad experience that broke updates on a website.

I'm extremely wary of that happening again, so I'd like to get more feedback on how to approach this.

I'll link this PR in the open issue (#84) in case any PWA experts find their way in there.

dhvcc:
> Thanks!
> 
> Unfortunately I haven't implemented a service worker in many years after a bad experience that broke updates on a website.
> 
> I'm extremely wary of that happening again, so I'd like to get more feedback on how to approach this.
> 
> I'll link this PR in the open issue (#84) in case any PWA experts find their way in there.

I think you're talking about using PWA libraries where the default behavior is to start caching stuff, which, without you knowing could break stuff? This implementation doesn't do anything, doesn't even create the cache. I use this kind of worker on my deployment and checking dev tools I don't see any cache at all for the page. Not even manifest.json is cached (which is the default behavior for most tools)

image

image

That said, I'm just giving context and no way a master of PWA stuff, but this is coming from what I've read in the docs, might link them if needed, or we can test different scenarios on my deployment to be 100% sure

henrygd:
Thanks for clarifying! Like I said it's been a very long time since I looked into this.

So the benefit to this is approach is that it automatically prompts the user to install to the home screen rather than requiring them to go into options and add it manually?

dhvcc:
Nevermind, it seems like the manifest alone is enough for the app to install as standalone on iOS with Safari, although I didn't test it on android I don't see reason how it'd be different, I'll close the PR, since manifest already exists in main branch (although not released yet)

PR #653, opened 2025-03-07
compute cpu and memory stats for windows docker daemon

ViryBe:
The Docker daemon's API returns different values on Windows and non-Windows systems. This impacts the calculation of the `cpuPct` and `usedMemory` statistics. The systems are disciminate on the `Server` header send by the server. Uses the unix stats calculation in case the header is not set.

`docker stats` implementation for reference:
https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L100

### DIFF
--- a/beszel/internal/agent/docker.go
+++ b/beszel/internal/agent/docker.go
@@ -14,8 +14,12 @@ import (
 	"sync"
 	"time"
 
+	"regexp"
+
 	"github.com/blang/semver"
 )
+
+var headerRegexp = regexp.MustCompile(`\ADocker/.+\s\((.+)\)\z`)
 
 type dockerManager struct {
 	client              *http.Client                // Client to query Docker API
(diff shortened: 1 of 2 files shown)


henrygd:
Thank you! I probably won't have time to test this today but I'll get back to you this weekend or next week.

henrygd:
Sorry for the delay, just got to this.

I installed Docker Desktop in Windows but the `Server` header says `linux`. Assuming because of WSL. Do I need to run Docker in Hyper-V?

henrygd:
Hyper-V didn't do it, but I finally figured out there's an option to switch to Windows Containers via the tray icon.

Unfortunately this isn't supported on my Windows version. I'll research what my options are when I have a bit more time.

beszel/internal/agent/docker.go
```diff
@@ -324,3 +337,31 @@ func getDockerHost() string {
...
+	}
+	return osType
+}
+
+// from: https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L185
+func calculateCPUPercentWindows(v container.ApiStats, prevCPUSsage uint64, prevRead time.Time) float64 {
+	// Max number of 100ns intervals between the previous time read and now
+	possIntervals := uint64(v.Read.Sub(prevRead).Nanoseconds())
+	possIntervals /= 100                // Convert to number of 100ns intervals
+	possIntervals *= uint64(v.NumProcs) // Multiple by the number of processors
```
henrygd, line 357: I think this line (357) needs to be removed. Docker's calculation multiplies usage by the number of cores, which can give values over 100%.

See here for more info: https://github.com/henrygd/beszel/issues/730#issuecomment-2773561880

ViryBe: Your comment is valid for Linux. For Windows, `docker stats` gives you the total between 0 and 100.

On line 357, `possIntervals` is multiplied by `NumProcs` but `possIntervals` is used to divide intervalsUsed.

The return value is essentially:

```
ret = (delta_used_cpu_total) / (delta_time * n_proc) * 100
```

`delta_time` being the real time elapsed, so `delta_time * n_proc` being the total time available for applications.

I hope this makes sense and that I understood the calculation correctly.

I tested it on a Windows machine with a Docker container running on multiple cores, and `docker stats` always gives a value in the [0,100] range.

It's odd to have different scales for Linux and Windows, but this is how I understand [calculateCPUPercentWindows](https://github.com/docker/cli/blob/c80675bfe18ccf9a8ae5f2885d888c86a15f8b8d/cli/command/container/stats_helpers.go#L185).

henrygd:
Merged, thanks!

I did a tiny bit of refactoring. Functionality should be exactly the same, but it would be super helpful if you could pull from main and confirm :+1:

ViryBe:
i've tested v0.11 today, and it seems to work as expected. 
thanks !

henrygd:
Awesome, thank you!

PR #792, opened 2025-04-30
Custom data support

pitbox46:
This is a work in progress!

# Description

Adds support for arbitrary numeric data in the form of "extra data" (`eData`). The functionality is as follows

1) Store the extra data in `system_stats.stats.eData` as a key-value pair. This is easily done by watching for new records and updating them.
2) Create a record under the `extra_data_configs` collection. Each record is 1 chart for 1 system. Depending on the parameters, we can display several different extra variables on the chart. 

We can utilize the Pocketbase API to add what we want using a separate process. This way, minimal changes have to be made to Beszel's code and the system is easily modifiable. 

# Battery Charge Example

[beszel_bat.mjs](https://github.com/pitbox46/beszel-extra-agent/blob/main/beszel_bat.mjs)

(I have the laptop to keep the battery at ~80%, so I expect the graph to be flat)

# Todo 

- [ ]  Implement the `extra_data_configs` collection in Beszel (unsure how how to do this)
- [x] Add example battery charge example
- [ ] Documentation

### DIFF
--- a/beszel/site/src/components/charts/extra-data-chart.tsx
+++ b/beszel/site/src/components/charts/extra-data-chart.tsx
@@ -11,7 +11,7 @@ import {
 	chartMargin,
 } from "@/lib/utils"
 // import Spinner from '../spinner'
-import { ChartData, EDataConfig, EDataKey } from "@/types"
+import { ChartData, EDataConfig } from "@/types"
 import { memo, useMemo } from "react"
 import { useLingui } from "@lingui/react/macro"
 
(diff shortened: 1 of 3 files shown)


henrygd:
Interesting idea, thanks. I'll think it over and try to get back to you tomorrow but it may be a few days.

pitbox46:
I'm not sure of how to add the container `extra_data_configs` via the agent (I'm not very familiar with go), so I had to implement it using my JS script for now. You can see my script here [beszel_bat.mjs](https://github.com/pitbox46/beszel-extra-agent/blob/main/beszel_bat.mjs). 

I will also update the description with some new info and an example image.

henrygd:
Sorry, forgot about this and it was closed automatically. I'll reopen and try to get to it relatively soon but I'm absolutely slammed at the moment :-/

GamerClassN7:
Any news this would be heally handy i want to switch to beszel from ZBX and this is only think i canot overcome since i use tone of custom metric walue loaded from many API

PR #816, opened 2025-05-14
feat: LA, migration, float alert value, manual alert value input

Allespro:
Hello, I made some changes that I really need for my servers monitoring:

- Load Average 1m 5m 15m (that's very important metrics I'm surprised they haven't been);
- Alerts for LA;
- Float alert value for LA, fields configured in `site/src/components/alerts/alerts-system.tsx` L217;
- Manual alert value input, click on number and edit;
- Migrations like PocketBase provide it;

idk why .po translation files was edited, but I committed it

beszel

la


Allespro:
Just updated the migration files, the migration will only be applied if the collection does not exist, allowing the old Beszel to be updated

henrygd:
Thanks, just from a quick look I'm not sure migrations are necessary. The values are included in the existing json object, right?

I can't review this PR immediately because I have limited dev time and really need to finish a major feature for the next release, but I'll get to it as soon as I can.

henrygd:
Thanks, this is in main now. I removed 1 minute load from the table just because we have no room left. I'll add a chart with all three values later to compensate for that.

I like what you did with the manual alert input, but I removed it for the time being and kept it in another branch for future reference. I'll add it later but want to make sure you can tab into it, so maybe a hidden input or contenteditible attribute.

PR #858, opened 2025-06-04
Update viewport meta to prevent input zoom on iOS

52Lxcloud:
- Prevents page zoom when tapping input fields on iPhone


henrygd:
Thanks! I don't have an iPhone to test.

Does this completely disable the ability of the user to zoom as well? I'd like to avoid that if possible.

52Lxcloud:
> Thanks! I don't have an iPhone to test.
> 
> 
> 
> Does this completely disable the ability of the user to zoom as well? I'd like to avoid that if possible.

In Safari, users can still zoom the webpage at will.

henrygd:
Thanks, I will test and merge this before the next release.

PR #867, opened 2025-06-07
Create service user as system user

TobiX:
This clearly marks the account as a "service account" to the administrator. Additionally, it's hidden from desktop login managers.

I used long options where possible, since that makes the script easier to read to viewers who aren't familiar with the called tools. Since the useradd used is from `shadow-utils` on all distros except Alpine (where busybox's adduser is used), these options should be available everywhere.

I was thinking about using systemd's `DynamicUser` feature, but wasn't sure if the user isn't used for permission management, so I left this out for now.

Using `/nonexistent` is a marker used in Debian to clearly signal the system administrator that the home directory is missing on purpose.

### DIFF
--- a/supplemental/scripts/install-agent.sh
+++ b/supplemental/scripts/install-agent.sh
@@ -317,14 +317,14 @@ fi
 if is_alpine; then
   if ! id -u beszel >/dev/null 2>&1; then
     echo "Creating a dedicated user for the Beszel Agent service..."
-    adduser -D -H -s /sbin/nologin beszel
+    adduser -S -D -H -s /sbin/nologin beszel
   fi
   # Add the user to the docker group to allow access to the Docker socket
   addgroup beszel docker
 else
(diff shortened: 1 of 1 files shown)


henrygd:
Thanks!

Using `DynamicUser` is a great idea. I don't think it will break anything.

We do add the user to the `docker` group, but we can use `SupplementaryGroups=docker` instead. Might need to check if the group exists before adding that. Or just `groupadd docker` so it works if Docker is installed in the future. 

Off the top of my head I don't know of anything else that would need to be changed. We have `StateDirectory` defined, so we should always be able to write there if needed.

Do you want to try adding `DynamicUser`? If not I can merge this now and add it later.

TobiX:
I think this change is still useful for non-systemd distros? So maybe merge this now and I'll take a stab at DynamicUser support later...

henrygd:
Sounds good, thanks :+1:

PR #922, opened 2025-06-26
opening systems in browser tabs

### LINKED ISSUE #865: [Feature]: opening systems in browser tabs

would love to be able to easily open each system in different browser tabs :)

ctrl/cmd + click to open a new browser tab

Anish-Chanda:
Support ctrl/cmd+click and shift+click for row and grid views. fixes #865 
also adds a `aria-label` to the `SystemCard` and `SystemTableRow`

### DIFF
--- a/beszel/site/src/components/systems-table/systems-table.tsx
+++ b/beszel/site/src/components/systems-table/systems-table.tsx
@@ -531,13 +531,19 @@ const SystemTableRow = memo(
 			return (
 				<TableRow
 					// data-state={row.getIsSelected() && "selected"}
+					aria-label={t`View details for ${system.name}`}
 					className={cn("cursor-pointer transition-opacity", {
 						"opacity-50": system.status === "paused",
 					})}
 					onClick={(e) => {
 						const target = e.target as HTMLElement
+						const systemPath = getPagePath($router, "system", { name: system.name })
 						if (!target.closest("[data-nolink]") && e.currentTarget.contains(target)) {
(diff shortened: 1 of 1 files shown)


Anish-Chanda:
Oh, the link component in the router could be updated to return early if ctrl/cmd/shift was pressed to allow the browser to handle the event if this behavior is okay for all the places `Link` was used.

henrygd:
Thanks!

The `Link` component is an `<a>` element and should be handled by the browser. So I don't think any changes are needed to the grid layout cards.

The `onclick` table rows were supposed to be temporary hack, but I never got around to updating them. I'm wondering if it's possible to do something like the cards and use an absolute positioned `<a>` in a `td`, then bring the other clickable elements up with `z-index` / `pointer-events`.

Then maybe something like this to stop the link from triggering when a nested element is clicked:

```js
if (e.target.closest("[data-nolink]")) {
  e.stopPropagation()
}
```

It's possible I already tried this and it didn't work, so I can merge your fix if you don't mind verifying whether the changes to the grid layout are necessary.

Anish-Chanda:
yea, it is required for the grid layout, i just tested it with the Link, didn't work

henrygd:
I was tinkering with this again and realized we needed to update the Link component to handle new tabs properly. I did that and it seems to fix this, as well as all other places the Links are used. So I'll close the PR but thank you again for your work :+1: 

Let me know if it's not fixed in the next release.

Issue #791, opened 2025-04-30
[Feature]: Display device uptime on the System table


me-2-you:
### Describe the feature you would like to see

Add device uptime information to main the dashboard view.

### Describe how you would like to see this feature implemented

Display options:

1. Add a column containing a cell with the device uptime.

2. Add the uptime as a tooltip when hovering over the device name.

Thank you.

henrygd:
This will be in the next release.

Issue #612, opened 2025-02-21
[Feature]: Dashboard filter on device Up or Down


cjmair:
### Describe the feature you would like to see

Is there any way to filter the dashboard, so it only displays devices that are Up or Down ?
If not it would be handy to have this option

### Describe how you would like to see this feature implemented

Don't mind

henrygd:
In the latest release you can search by "up", "down", or "paused" in the filter bar.

PR #927, opened 2025-06-26
[Feature] Add Uptime and status dorpdown to Dashboard

svenvg93:
This PR adds system uptime to the Dashboard and introduces a dropdown filter to view systems by status: Up, Down, or Paused.

closes https://github.com/henrygd/beszel/issues/612
closes https://github.com/henrygd/beszel/issues/791
Scherm­afbeelding 2025-06-27 om 00 17 36

### DIFF
--- a/beszel/site/src/components/systems-table/systems-table.tsx
+++ b/beszel/site/src/components/systems-table/systems-table.tsx
@@ -61,6 +61,8 @@ import {
 	Settings2Icon,
 	EyeIcon,
 	PenBoxIcon,
+	ClockIcon,
+	FilterIcon,
 } from "lucide-react"
 import { memo, useEffect, useMemo, useRef, useState } from "react"
 import { $systems, pb } from "@/lib/stores"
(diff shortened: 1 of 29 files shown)


henrygd:
Do you think this could work as a column in the existing 'View' options?

Maybe four columns: Layout / Status / Sort By / Visible Fields?

svenvg93:
> Do you think this could work as a column in the existing 'View' options?
> 
> 
> 
> Maybe four columns: Layout / Status / Sort By / Visible Fields?

Sure we can somehow. 
Let me have a look at it and come back to you.

svenvg93:
@henrygd moved it to the "View" option

henrygd:
Sorry for the long wait on these PRs.

The table at this point is about at maximum horizontal capacity, so I want to avoid adding new things for the time being. Otherwise it overflows and hides the action buttons which can be confusing for new users.

Do you think uptime and last seen could work in a tooltip while hovering over the agent cell?

In the near future I'd like to redesign slightly with a sidebar and possibly remove the width constraints for the table so we can start adding things in again.

svenvg93:
I do agree the table is getting polluted with all the information, at the same time the user can select what they want to seem themself by selection the ones they want. For example I have the GPU one hidden as I don't have an GPU to monitor. Option could be to save these preferences in the user settings. 

We could make an tooltip as well of course 
Scherm­afbeelding 2025-08-28 om 21 34 11

Let me know if you want this, otherwiseI will remove them for now and only keep the filtering changes.

henrygd:
Good points, let's just remove for now and we'll add them back into the table after redesign :+1: 

I'll get this in the next release.

svenvg93:
Sounds good I removed them 😄

henrygd:
Thanks so much!

Impact123:
> I'll get this in the next release.

Was this forgotten or am I blind? Seeing the last refresh time would be very helpful.

Issue #741, opened 2025-04-09
[Feature]: Always show SWAP graph


Tom5421:
### Describe the feature you would like to see

How would you feel about showing the graph if SWAP is enabled on the system, even if its not being actively used?

### Describe how you would like to see this feature implemented

The SWAP graph would show if SWAP has been enabled on the machine, even if it is not using it currently.

ncrosty58:
Apparently swap is hidden whenever usage is presently 0 even if past usage (within the charts timeframe window) was above 0.

nalpo:
For me It would be helpful to always have that info for me, even if it is zero most of the time. Right now I am debugging a VM which randomly stops and I would like to see if and how swap was used at the time. What about integrating swap into the RAM Diagram? Then there would be no empty diagram.

BernhardBaumrock:
+1 — I’d love to see the swap chart always shown when swap is enabled, even if current usage is 0.

On my server (multi-service Docker stack: two Paperless instances, Celery, Tika, Hermes, etc.) I recently added a 4 GiB swap file as an OOM safety net. Swap is active and verified (`swapon --show`, `free -h`), but with `swappiness=10` it stays at 0 B under normal load — which is exactly what I want. The problem is the UI hides the swap chart in that case, so it looks like swap isn’t configured at all.

I’m not trying to monitor swap pressure day-to-day; I want visibility that swap *exists* and is sized correctly, without waiting for memory pressure to prove it’s there.

Thanks for considering this!

Issue #278, opened 2024-11-10
More details in details page for CPU, RAM, DISK


flo82:
I'm using different tools for monitoring my systems. What i'm missing in beszel are the following improvements:

CPU graph could be devided into:
* CPU
* IOWAIT
* Steal
* User
* System

RAM could be devided into:
* RAM
* SWAP
* Buffered
* Cached

Also an average, max value for the selected timerange would be appreciated.

Furthermore fstab could be used and monitored to get more devices for monitoring disk-space.

Total values for network for the timerange would also be interesting.

What do you think of the suggestions?
I would like to help you, but I have no idea of go-language :-(

henrygd:
More CPU metrics could be helpful.

RAM is already divided into used / buffer + cache, as well as ARC if you use ZFS. And there's a separate chart for swap if swap is being utilized by your system.

We have average / max toggle for various metrics like cpu and bandwidth. Not for memory though. We could add it, but generally memory doesn't fluctuate as rapidly as the others so I'm not sure it's needed.

Monitoring fstab is an interesting idea, but there are still limitations with docker where the disks need to be manually set up. And perhaps fstab includes things that people don't want or need to be monitored.

Total network traffic is definitely something I want to add at some point. I think there's one or two other issues about that.

olderthantheinternet:
Love the simplicity of this project, it installs easily and you are up and running very quickly.

I too would LOVE to see IOWait added

mrspartak:
+1 for steal time metric. As a lot of us most probably use VPS and noisy neighbor problem is not always easy to detect.

adrianmihalko:
It would be nice if the Disk I/O graph included information about which processes are responsible for the activity (like iotop).

Silun:
I would love CPU steal logging in Beszel!

svenvg93:
@henrygd this one is implemented now :)

PR #933, opened 2025-06-27
[Feature] Add CPU usage breakdown and expend SWAP #931

svenvg93:
This PR adds the Free/Used/Cache for the SWAP as well as breaksdown the CPU in System/Usser/IOWait/Steal

closes https://github.com/henrygd/beszel/issues/278
closes https://github.com/henrygd/beszel/issues/741

Scherm­afbeelding 2025-06-27 om 23 11 05
Scherm­afbeelding 2025-06-27 om 23 16 49

### DIFF
--- a/beszel/internal/agent/agent.go
+++ b/beszel/internal/agent/agent.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/shirou/gopsutil/v4/cpu"
 )
 
 type Agent struct {
@@ -25,6 +27,7 @@ type Agent struct {
 	systemInfo    system.Info                // Host system info
 	gpuManager    *GPUManager                // Manages GPU data
 	cache         *SessionCache              // Cache for system stats based on primary session ID
+	prevCpuTimes  []cpu.TimesStat            // Previous CPU times for calculating detailed metrics
 }
 
 func NewAgent() *Agent {
--- a/beszel/internal/agent/system.go
+++ b/beszel/internal/agent/system.go
@@ -77,11 +77,56 @@ func (a *Agent) getSystemStats() system.Stats {
 		systemStats.Cpu = twoDecimals(cpuPct[0])
 	}
 
+	// detailed cpu times
+	if cpuTimes, err := cpu.Times(false); err == nil && len(cpuTimes) > 0 {
+		// Get the first CPU (total across all cores)
+		currentCpu := cpuTimes[0]
+		
+		if len(a.prevCpuTimes) > 0 {
+			prevCpu := a.prevCpuTimes[0]
+			
+			// Calculate deltas
+			totalDelta := currentCpu.Total() - prevCpu.Total()
+			if totalDelta > 0 {
+				userDelta := currentCpu.User - prevCpu.User
+				systemDelta := currentCpu.System - prevCpu.System
+				iowaitDelta := currentCpu.Iowait - prevCpu.Iowait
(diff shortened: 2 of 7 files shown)


svenvg93:
@henrygd Taking your comment in the other PR about the Load Average into account, I’m guessing the Extra Disks section is also taking up too much space here, especially when there are many disks. I’ll look into creating a more compact view for it — maybe something where the user can select 2 or 3 disks they want to focus on.

Idea:
One there is one disk we just show it as normal. 
Scherm­afbeelding 2025-07-15 om 19 55 44

When there are more we show an `x Disks` with an tooltip to show them all. 
Scherm­afbeelding 2025-07-15 om 19 54 59

svenvg93:
@henrygd would it be possible to get this one in the 0.12 release as it would be great possibility to deliver some of the most requested features 😁

henrygd:
I'm trying to get the 0.12.0 release out today or tomorrow (dealing with a WinGet upgrade issue). So it won't be in that, but I can definitely get it in a patch release afterwards.

I like the look of the tooltip for the extra disks. Maybe we can add it to the existing disk column rather than adding the extra column?

svenvg93:
> I like the look of the tooltip for the extra disks. Maybe we can add it to the existing disk column rather than adding the extra column?

That is a good idea as well, the only doubt I have with the tooltip is that it might become to big when there is a lot of disks. Maybe we can add a "+" next to the current disk and have it expend a sheet with all the disks, it will give enough room of the disks.

svenvg93:
> I like the look of the tooltip for the extra disks. Maybe we can add it to the existing disk column rather than adding the extra column?

I move the tooltip to the Disk colunm 😄

jonashaag:
Hello! I can't tell from the issue description or screenshots: Does this add the ability to show max memory use when plots are aggregated? That'd be very helpful for detecting memory leaks.

henrygd:
FYI - I moved some API related stuff to its own module which may create more merge conflicts on multiple PRs. Please don't worry about this as I will sort it out when merging.

These changes are good but I'm not 100% sure about adding free memory and free disk space. To me it's already clear how much free memory exists by the empty space and the fact the Y axis is based on the total value.

The chart is also a little harder to read, at least for me. Maybe it would be easier if the order was always free at the top, cache / buff in the middle, and used at the bottom.

Just my two cents, and I'm happy to include free if it's the standard way that other software displays this. I'll look further into it :)

henrygd:
@jonashaag It does not but I'll add that to my list.

svenvg93:
I'm fine with removing the free part, if you prefer that. It might make the charts easier to read. 

edit: I removed them.

svenvg93:
@henrygd This one is up to date and ready as well

StephanRoemer:
I agree that it might be not so useful for the beszel graphs. But free disk space would definitely be a good thing for the API (like using it in Home Asisstant).

svenvg93:
@henrygd I fixed all the conflicts, hopefully without breaking anything ;)

henrygd:
I just added a way to pretty name devices with double underscores like `sda2__Jellyfin Media`.

This caused some minor conflicts again but it's no trouble for me to resolve these on my side :+1:

svenvg93:
@henrygd  Is there anything I can do to help getting this PR across the finish line? Don't mind breaking it up in smaller PRs If that makes it easier  for you.

henrygd:
If it's not too much trouble that would be great!

svenvg93:
> If it's not too much trouble that would be great!

Of course! I will close this one and will slowly work on separate PRs

PR #938, opened 2025-06-30
Adds display unit preference

### LINKED ISSUE #402: Ability to change the displayed units.
- Fahrenheit option for temperatures.
- Bits per second (bps) option on Network I/O and Disk I/O. 
  - As separate options since some people prefer B/s for Disk but bps for Net traffic.

Love Beszel. It's by far the cleanest and simplest solution I've found for system monitoring :)

Anish-Chanda:
Adding unit display preferences for Temperature, Networking, and Disk throughput. fixes #402


henrygd:
Cool, thanks. I should have a beta version of the next release out in the next few days and I'll follow up on the open PRs afterwards.

Anish-Chanda:
changes:
- updated settings UI to add new options
- updated `stores.ts` and `internal/users.go` to store the user's unit preferences 
- new conversion helpers in `utils.ts`
- updated area container (we check which charts to add units to from their name) and temperature charts to support units.

Added options to select unit preferences under general settings
image

screenshots with Fahrenheit for temp, bps for networking and MB/s for disk throughput.
image
image

henrygd:
Thanks, I do want to change the data coming in from the agent to use bytes at some point, rather than MB, which will make these conversions more accurate.

Anish-Chanda:
@henrygd first we could make a change in the hub that converts existing `dr`, `dw` etc in system stats from MB to bytes, and based on the agents version we either write it to db directly or for older agents we convert it to bytes. in the subsequent release for the agent we could start pushing bytes instead of MB. there will still be a bit of inaccuracy in the historic data but once the agents are updated we should have accurate data for these charts. this would make this change the least disruptive. I'm also open to alternative approaches or discussion if anyone has other ideas.

henrygd:
That would be a good way to do it. Or add an extra property like `bytes: true`.

0.12.0 beta is slightly delayed so I can get another feature in, but I'll be done very soon. We may be able to get this in the final 0.12.0 release. There might be conflicts when I merge my dev branch but don't worry about that.

henrygd:
Merged, thanks! I expanded it a bit so the choices are bits and bytes, and the bytes option will show the whole range (KB, MB, etc).

Issue #909, opened 2025-06-21
[Feature]: Collect and display operating system (os) version


viperk1:
### Describe the feature you would like to see

It would be great to have the OS version captured by the agent and then displayed on the dashboard as well as being able to trigger notifications about it. I know today we capture the kernel information.

### Describe how you would like to see this feature implemented

The agent can capture this information based on the operating system (linux/windows/etc) and make it available to the hub for further actions.

svenvg93:
Thanks for taking the time to suggest this!

This issue is a duplicate of an existing one, so we’ll be closing it to keep the discussion in a single place. You can follow progress and updates on the linked issue, and feel free to subscribe there or add any additional details that might help.

Thanks again for helping improve the project!

Issue #899, opened 2025-06-19
[Feature]: CPU core clock


Nassi88:
### Describe the feature you would like to see

Please add CPU core clock than i can delet grafana😉

### Describe how you would like to see this feature implemented

I want to see all cpu cores clock

svenvg93:
Can you please provide more details?
What would you like to see in terms of CPU core clocks. Do you want to see the speed for example?

Nassi88:
I was thinking of a display where each core shows how high the clock is. Of course, the update interval should be short there. Maybe that it doesn't need too much performance because of the refrash that it only runs when you call up the web. I don't know if that's feasible. I think if the refrash is short that it needs unnecessary power. But it would be cool to have a graphical view of the core clock with ghz each core

Issue #891, opened 2025-06-17
[Feature]: Collect OS Info


Justinzobel:
### Describe the feature you would like to see

I'd love to see OS info collected by the agent.

So this could be displayed both on the overview page and the individual server info.

This way, admins can see which of their devices need an OS upgrade.

### Describe how you would like to see this feature implemented

The same as any other metric.

svenvg93:
Thanks for taking the time to suggest this!

This issue is a duplicate of an existing one, so we’ll be closing it to keep the discussion in a single place. You can follow progress and updates on the linked issue, and feel free to subscribe there or add any additional details that might help.

Thanks again for helping improve the project!

Issue #205, opened 2024-10-14
[Feature Request] - More Detail On System Details Page + Mult Core CPU and GPU Graph


serialscriptr:
I think it would be useful to be able to see more detail relevant to each of the graphs on the system details page. Examples can be:
- Disk Space: show information about disk like disk name and manufacturer
- Bandwidth: show name of network adapter being monitored
- Total Memory Usage: manufacturer and total ram count
These could be displayed as a tool tip when you hover over the header of the relevant graph

It would also be nice to have a per core CPU graph and a GPU graph. I've been running jellyfin on an orange pi 5 and being able to monitor the GPU on it would be fantastic

henrygd:
@gander I'm not sure why you posted that in this issue. Looks like maybe your environment is restricted to one core? If you think the stats are wrong, please open a new issue with more info, thanks.

gander:
@henrygd yeah. I must have misunderstood something.

svenvg93:
This has been implemented as part of S.M.A.R.T. and other improvements.

PR #952, opened 2025-07-10
[Feature] Collect hardware and OS metadata (CPU, RAM, disk, manufacturer, etc.)

svenvg93:
This PR enhances the collection and display of hardware and OS metadata.

### Features

- Collect disk model and vendor information for display in the System Info bar
- Collect CPU architecture details, shown in the CPU tooltip
- Collect OS metadata (e.g. version) for the System Info bar
- Collect network interface details (name, vendor, model)
- Move kernel version display to the OS tooltip
- Display total memory in the System Info bar

https://github.com/user-attachments/assets/9399bfe3-a659-4e6c-8808-305dc91ec34a

### Chore
Restructure the system.info logic for improved clarity and maintainability

closes https://github.com/henrygd/beszel/issues/891
closes https://github.com/henrygd/beszel/issues/909
closes https://github.com/henrygd/beszel/issues/205
closes https://github.com/henrygd/beszel/issues/899


galkin-it:
When this pr is merged? I need this stuff :)

svenvg93:
> When this pr is merged? I need this stuff :)

I hope soon, but I know @henrygd has a lot on his plate now.

svenvg93:
@henrygd  Is there anything I can do to help getting this PR across the finish line? Don't mind breaking it up in smaller PRs If that makes it easier  for you.

henrygd:
That would be great if you can. One of the reasons I haven't gotten to this is because I've been thinking about how to split up what needs to be collected every minute from what doesn't.

For example, things like OS, CPU model, kernel, net interface capabilities, are not going to change within the lifetime of the process, and I don't necessarily want to keep adding more data to the operation that runs every minute.

We've added more flexibility recently in how we can collect from agents, so maybe we just collect most of this upfront when the agent connects, or maybe we can run a different collection operation every 10 or 15 min? Just throwing out ideas.

svenvg93:
Even though this information now is collected every minute, its only read from the system at the start of the agent and stored in the `system.info` its not part of the `system stats`.
But I do  agree with you,  it even could only be collected the first time or something like every 24 hours.

svenvg93:
@henrygd I added an `StaticInfo` part for this information that will be collected every 15 minutes

svenvg93:
@henrygd any update on this ?

svenvg93:
I updated the collection method to be only upon agent connection.

viperk1:
Waiting eagerly for this one 😺

svenvg93:
@henrygd Would it be possible to get this into the next release?

henrygd:
I will try to get it in.

henrygd:
I'm going to merge this to a separate branch for now without merging to main for a few reasons:

- While reviewing I thought of broader design changes I'd like to make regarding the static info collection that go beyond the scope of this PR. It would be easiest for me to start fresh, then integrate this functionality afterward. I will prioritize this.

- It looks we can't read disk or NIC metadata from inside a container, which is a big drawback. I hoped `network_mode: host` or using a `device` mapping would work, but it doesn't seem to. More work needs to be done here to figure out a possible solution.

- Running in a container also complicates how we get OS details. We would need to do something like check if we're in a container and get the host OS from the Docker API if available. Otherwise it shows the OS of the container.

svenvg93:
Fair point about the container agent. Did not realise that when doing the development as thats done with the binary agent in my case. Sorry for that.

viperk1:
> I'm going to merge this to a separate branch for now without merging to main for a few reasons:
> 
> * While reviewing I thought of broader design changes I'd like to make regarding the static info collection that go beyond the scope of this PR. It would be easiest for me to start fresh, then integrate this functionality afterward. I will prioritize this.
> * It looks we can't read disk or NIC metadata from inside a container, which is a big drawback. I hoped `network_mode: host` or using a `device` mapping would work, but it doesn't seem to. More work needs to be done here to figure out a possible solution.
> * Running in a container also complicates how we get OS details. We would need to do something like check if we're in a container and get the host OS from the Docker API if available. Otherwise it shows the OS of the container.

Hey Henry, as you're developing this I'd love if some of these (OS specifically) can make it as an optional column to the top level all systems view. Thank you!

PR #960, opened 2025-07-13
[Feature][0.12.0-Beta] Enhance Load Average Display, Charting & Alert Grouping

svenvg93:
### Summary
This PR enhances the recently added Load Averages feature by improving visibility, usability, and alert consolidation.

### System Table

- Added the 1-minute load average alongside the existing 5- and 15-minute values.
  - We use the core count to calculate the bar.
- Consolidated all three load averages into a single column, with a hover tooltip to reveal the 1m, 5m, and 15m figures.

Load averages in system table with tooltip

### Historical Chart

- Introduced a chart to track load average trends over time, giving users deeper insight into system performance.

Load average history chart

### Alerts

- Included the 1-minute load average in alert notifications for quicker spike detection.
- Grouped all Load Average alerts into a single collapsible section to reduce dashboard clutter.

Collapsible load average alerts

Please review and let me know if any tweaks are needed!

### DIFF
--- a/beszel/Makefile
+++ b/beszel/Makefile
@@ -56,7 +56,7 @@ dev-hub: export ENV=dev
 dev-hub:
 	mkdir -p ./site/dist && touch ./site/dist/index.html
 	@if command -v entr >/dev/null 2>&1; then \
-		find ./cmd/hub ./internal/{alerts,hub,records,users} -name "*.go" | entr -r -s "cd ./cmd/hub && go run . serve"; \
+		find ./cmd/hub/*.go ./internal/{alerts,hub,records,users}/*.go | entr -r -s "cd ./cmd/hub && go run . serve --http 0.0.0.0:8090"; \
 	else \
 		cd ./cmd/hub && go run . serve --http 0.0.0.0:8090; \
 	fi
--- a/beszel/internal/agent/system.go
+++ b/beszel/internal/agent/system.go
@@ -251,6 +251,7 @@ func (a *Agent) getSystemStats() system.Stats {
 
 	// update base system info
 	a.systemInfo.Cpu = systemStats.Cpu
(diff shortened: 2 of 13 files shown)


henrygd:
Thanks, the chart looks great.

I like the concept for load avg in the table, but increasing the row height decreases the number of systems in view, so I'd like keep things on one line.

We could show them inline like `0.53, 1.23, 0.86` since that's how they're usually presented (in the `uptime` command for example). The drawback is sorting. I'm not sure we can have three separate sort states. But we could just add them together and sort by total value.

I also want to add a setting to choose between displaying raw values or normalizing by cores. Then maybe an indicator dot like the one next to the agent version based on the normalized value.

What do you think?

svenvg93:
I see what you mean, next to each other would be better in terms of space.
Quick adjustment; 

https://github.com/user-attachments/assets/83d3700c-f6fb-4782-be22-49aaae4d2c4e

The dots change based on the normalized value. Let me know if you like it, then I will commit it.

For the sorting; to simplify things, we can disable sorting for this column. The CPU Load column already reflects the load and is  sortable.

henrygd:
Thanks, that looks good. I'll also tweak it a bit after you commit to see what works.

Maybe we get rid of the slashes and just use a space in between like `htop`. And maybe even use a single dot indicator based on the highest value, just to shrink it a bit.

Initially I was on board with grouping the load average alerts, but on further reflection I think we should leave them at top level. What bothers me is having to click an extra time, which could be annoying if you want to go through the systems to check if a particular load average alert is active.

I know the dialog is getting pretty big, so I'll probably switch it to a sheet / sidebar component at some point. Besides that we're good to go :+1:

svenvg93:
I changed the alerts back to be ungrouped. Should be ready to go now! 

I had some ideas for the alert dialog, I will post them in an discussion when I have something usefull to pitch 😉

Issue #945, opened 2025-07-05
[Feature]: Adding hebrew localization


gabay:
### Describe the feature you would like to see

Hey - I'd like to translate this project to hebrew, if that's of any interest :)

### Describe how you would like to see this feature implemented

Adding hebrew to crowdin and leaving the translation for me.

henrygd:
Sorry for the delay, I've just added Hebrew. Thanks for your help translating.

Let me know if you want me to prefill machine translations to cut down on some of the work.

https://crowdin.com/editor/beszel/all/en-he?filter=basic&value=0

gabay:
Thanks!
I'd love to get some machine translation to start from, if it's not too much hassle/cost.

henrygd:
No problem, just added them. Let me know once you've looked them over 👍

gabay:
I reviewed all translations and siggested a few modifications

It looks good :) thanks.

henrygd:
Thanks, it should be in the next release.

PR #961, opened 2025-07-13
[Feature] Add Hebrew Localization Support

svenvg93:
## Summary
Add Hebrew to the list of available translation languages.

## Changes
- Introduce he locale and include the initial Hebrew translation files
- Update language selector to surface Hebrew as an option

closes https://github.com/henrygd/beszel/issues/945.

### DIFF
--- a/beszel/site/lingui.config.ts
+++ b/beszel/site/lingui.config.ts
@@ -11,6 +11,7 @@ export default defineConfig({
 		"es",
 		"fa",
 		"fr",
+		"he",
 		"hr",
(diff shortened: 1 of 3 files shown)


svenvg93:
Not stale, just patiently waiting for @henrygd

henrygd:
I just added Hebrew on the Crowdin side. We shouldn't add it to Beszel yet until translations are done. We've had a couple false starts on languages that were never completed.

Crowdin will generate the PO file, so we don't need that in the PR.

Issue #1006, opened 2025-07-28
[Bug]: Binary Hub/Agent not restarting after update


bradley-varol:
### Component

Hub & Agent

### Description

With the recent 0.12.x updates, I noticed that the hub and agents were not updating. I made sure that the daily update check was enabled for each agent.

Simply restarting the hub and agents with `systemctl restart` fixed this, so perhaps the update script isn't restarting the services properly?

### Expected Behavior

Hub is updated after running the install script, and Agents are automatically updated in the background - no need to SSH and run `systemctl restart`.

If this is the expected way to update/restart, perhaps it could be documented.

### Steps to Reproduce

1. Install Beszel agent/hub on Debian
2. Press `y` to confirm you'd like automatic updates
3. Wait for an update
4. Agent/hub not updated
5. Run `systemctl restart beszel-{agent|hub}`
6. Version is updated

### Category

Installation

### OS / Architecture

Proxmox CTs running Debian (Bookworm)

### Beszel version

0.11.1

### Installation method

Binary

svenvg93:
@bradley-varol  can you confirm that with the last release the agent was restarted?

bradley-varol:
> [@bradley-varol](https://github.com/bradley-varol) can you confirm that with the last release the agent was restarted?

Sorry for delay in responding.

I just manually updated the Hub from `0.12.3` to `0.12.7`. The Agents now say `0.12.6` so it looks like they might be updating daily!

Let's wait for one more release and I'll confirm if the Hub and Agents updated so we can close this.

bradley-varol:
Looks like the Agents updated to `0.12.9` but the Hub was still on `0.12.7` 🤔 

I had to run the install script AND restart the service for it to update!

henrygd:
The hub binary install script didn't include auto-update functionality, but I just added it.

You'll need to rerun the install script again. ~Uninstall first with the `-u` flag, then~ run with the `--auto-update` flag added.

You should see it print `Setting up daily automatic updates for beszel-hub...`

If you don't want to do this, simply running `/opt/beszel/beszel update` *should* work.

bradley-varol:
I re-ran the install script with `--auto-update` and I did see `Setting up daily automatic updates for beszel-hub...` 🚀 

I'm happy for this to be closed if you are.

Btw thanks a lot for your hard work - Beszel has been really useful so far! The alerting helped minimize downtime when a bug caused a server to quickly run out of disk space :)

henrygd:
[...]

Issue #469, opened 2025-01-22
[Bug]: install-agent.sh doesn't upgrade the agent, at least no change from hub.


Oaklight:
### Description

daily update doesn't seem to work. Rerun the `install-agent.sh` script doesn't upgrade the agent. At least from the hub, it's still 0.8.0.
I have to uninstall the agent using the install-agent.sh then reinstall it.

### Expected Behavior

update the agent without uninstall

### Steps to Reproduce

1. install an earlier version using script
2. enable the daily update. Or rerun the install script
3. check hub after a while, disable/resume the corresponding agent

### OS / Architecture

debian 12.9 & ubuntu 24.04

### Beszel version

0.9.1 script after china-mirror patch

### Installation method

Binary

### Configuration

```yaml
no special setting. all done by install-agent.sh script
```

### Hub Logs

```json

```

### Agent Logs

```shell

```

henrygd:
If you use `install-agent.sh` to update, try restarting the service afterward:

```bash
sudo systemctl restart beszel-agent.service
```

The auto update may not work if your machine is in China because the built-in update (`./beszel-agent update`) does not use a proxy for GitHub.

I'm using [rhysd/go-github-selfupdate](https://github.com/rhysd/go-github-selfupdate/) which does not appear to support a proxy URL. I will look into forking adding support for this.

Oaklight:
I have machine both in and outside china. The autoupdate doesn't work on either machines.

henrygd:
Okay, no worries. Let's figure this out.

What happens when you run these commands?

```bash
/opt/beszel-agent/beszel-agent -v
```

```bash
/bin/sh -c '/opt/beszel-agent/beszel-agent update | grep -q "Successfully updated" && (echo "Update found, restarting beszel-agent" && systemctl restart beszel-agent) || echo "No updates found"'
```

```bash
/opt/beszel-agent/beszel-agent -v
```

And can you please paste the content of `/etc/systemd/system/beszel-agent-update.service`?

Oaklight:
I run the three commands on the machine displayed as 0.8.0 from hub.
```
~# /opt/beszel-agent/beszel-agent -v
beszel-agent 0.9.1
~# /bin/sh -c '/opt/beszel-agent/beszel-agent update | grep -q "Successfully updated" && (echo "Update found, restarting beszel-agent" && systemctl restart beszel-agent) || echo "No updates found"'
No updates found
~# /opt/beszel-agent/beszel-agent -v
[...]

PR #1009, opened 2025-07-28
[Chore] Improve auto update mechanism 

svenvg93:
## 📃 Description

This PR (hopefully) improves the auto update mechanism.

closes https://github.com/henrygd/beszel/issues/1006
closes https://github.com/henrygd/beszel/issues/469

### ✏️ Changed

- Add the service restart to the `update.go` for the agent and hub
- Use the build-in update command to update the agent and hub on the timer

### 🗑️ Removed

- creation of the `run-update.sh`

### DIFF
--- a/beszel/internal/agent/update.go
+++ b/beszel/internal/agent/update.go
@@ -4,6 +4,7 @@ import (
 	"beszel"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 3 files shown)


beszel/internal/agent/update.go
```diff
@@ -53,4 +54,100 @@ func Update() {
...
+	// Try chcon first
+	if chconCmd, err := exec.LookPath("chcon"); err == nil {
+		binaryPath, _ := os.Executable()
+		chconExec := exec.Command(chconCmd, "-t", "bin_t", binaryPath)
+		if err := chconExec.Run(); err != nil {
+			fmt.Println("Warning: chcon command failed to apply context.")
+		}
+	}
+
+	// Try restorecon as well
```
trancee, line 143: @svenvg93 just wondering, but does this command not just revert the context back to the default?

> The `restorecon` command is used to restore the SELinux security context of files and directories to their default values as defined in the SELinux policy.

henrygd:
This is merged, thanks again for your work. :+1: 

I ended up rewriting the updater based on what PocketBase does, so we were able to get rid of the `rhysd/go-github-selfupdate` dependency as well. Fingers crossed this works for everyone :)

PR #1014, opened 2025-07-29
[Feature] Add support to monitor Tailscale nodes

svenvg93:
## 📃 Description

This PR adds support to monitor your Tailnet with Beszel! Using the Tailscale API we grab all the information  and display it on a different  tab. Monitoring is done via the Hub and the API is called every 5 minutes. 

| Variable | Description | Default | Required |
|----------|-------------|---------|----------|
| `TAILSCALE_ENABLED` | Enable Tailscale network monitoring | `false` | No |
| `TAILSCALE_TAILNET` | Tailscale tailnet name (e.g., `funny.ts.net`) | - | Yes (if enabled) |
| `TAILSCALE_API_KEY` | Tailscale API key for authentication | - | Yes* |
| `TAILSCALE_CLIENT_ID` | OAuth2 client ID for Tailscale API | - | Yes* |
| `TAILSCALE_CLIENT_SECRET` | OAuth2 client secret for Tailscale API | - | Yes* |

**Note**: For Tailscale authentication, you must provide either:
- `TAILSCALE_API_KEY` (API key method), OR
- Both `TAILSCALE_CLIENT_ID` and `TAILSCALE_CLIENT_SECRET` (OAuth2 method)

If accepted, I will add alerts for offline and out of date versions. 

## 📖 Documentation

Will update it when merged

### ➕ Added

- `Tailscale_stats` collection to store information 
- Add an Tailscale tab on the homepage, it will only show when `TAILSCALE_ENABLED`  is set to true.
- Data table to show the nodes with actions to copy the ip or magic dns name

## 📷 Screenshots

https://github.com/user-attachments/assets/2a090606-c1e3-43c3-88ff-9a9b02d55332


henrygd:
Very cool! We can definitely add this but I need to work through some older PRs first.

How long does it take to fetch the data from Tailscale? If it's not too long then maybe we can fetch it on demand (when the user starts a session or opens this page) and cache for a few minutes rather than fetching continuously in the background. Then we may not even need to store it in the DB either, just use the expiry map package I added for the universal tokens.

Also we should probably make this configurable at a user level rather than having only one global configuration.

svenvg93:
No worries there is no rush, I like to build my ideas and offer them to you 😄. 

Fetching the data takes less than a second, but this can vary depending on the number of nodes in your network.

I chose to fetch the data in the background so we can monitor changes over time and potentially trigger alerts. Some basic statistics are already stored, and we could even generate charts showing metrics like the number of online devices. If we only fetch the data on demand, we’re limited to a snapshot of the current state—essentially duplicating what’s already visible in the Tailscale admin console, which I think defeats the purpose.

We could make this behavior configurable at the user level. That said, How often users of the same Beszel instance would actually have different tailnets to monitor. Thats why a global settings made sense to me.

Btw, I also have a branch that integrates Tailscale in the Hub to have to show up on your tailnet as a node. Let me know if you want it.

TheLinuxGuy:
This PR is amazing.

Have you considered making these changes compatible also with netbird.io? They are like tailscale but have a few more "free" features and this feature could possibly handle both - their APIs seem similar to how you are retrieving stuff https://docs.netbird.io/api/resources/peers#list-all-peers

svenvg93:
> Have you considered making these changes compatible also with netbird.io? They are like tailscale but have a few more "free" features and this feature could possibly handle both - their APIs seem similar to how you are retrieving stuff https://docs.netbird.io/api/resources/peers#list-all-peers

I'm not using Netbrid myself so I have not looked at it. The API does seems to be similar but not fully, some fields are name differently. As well I use the tailscale Go client for the API, Netbrid seem to need an normal HTTP call. 

After this one is merged, we could add a Netbrid version as well by adapting the Tailscale Manager for Netbrid.

solracsf:
Just an appreciation post for this amazing work!
Thanks a lot for this contribution 💘

kleinweby:
This looks nice. Could you add support for changing the [base api url](https://github.com/tailscale/tailscale-client-go/blob/74c8fc3cb4d7/v2/client.go#L27)?

PR #1026, opened 2025-08-02
[Feature] Store/Edit agent config on the Hub

svenvg93:
## 📃 Description

This PR allows the agent to receive the config from the Hub upon changes. If configuration is set with environment values these will take president over the hub config. 

### ➕ Added

- Pushconfig from the Hub.
- agent_config colunm to the system collection
- UI to configure the agent config

## 📷 Screenshots
Scherm­afbeelding 2025-08-13 om 19 13 03


svenvg93:
@henrygd Do you want me to resolve the conflicts for this one?

henrygd:
No need right now, thanks for asking :)

I've had a couple other things come up, but should be finished with the refactoring for 0.13.0 within the next few days. It will allow adding new actions like this to be done a little more easily / cleanly. So I should be able to adapt it to the new flow without too much trouble.

PR #1010, opened 2025-07-28
[Feature] Move name copy to Action Dialog

svenvg93:
## 📃 Description

This PR moves the "Copy name button" to the Action Dialog not prevent from copying the name when trying to opening the system

Closes https://github.com/henrygd/beszel/issues/1007

### ✏️ Changed

- Move Copy name to Action Dialog

PR #1027, opened 2025-08-02
fix systems table names wrapping

bradley-varol:
## 📃 Description

The copy button was moved in #1010 but caused text to break and wrap which looks a jarring.

This PR adds `whitespace: nowrap` to the system table names.

## 📷 Screenshots

before
image

after
image


henrygd:
Gracias!

PR #1032, opened 2025-08-04
fix: allow system titles to expand

bwees:
## 📃 Description

It appears that the system titles column is maxed at 200px wide. This causes longer system titles to be word wrapped onto 2 columns (see picture). These titles that wrap are not very long and should still fit on one line to make the column readable.

image

## 📖 Documentation

N/A

### 🔧 Fixed

- word wrapping on system title column

## 📸 Screenshots

After changes:
image


bwees:
it appears this got fixed in 0.12.3. Maybe I should check the release notes next time lol

henrygd:
No worries, I appreciate the effort :smile:

PR #1053, opened 2025-08-16
Use British flag for English, and use Portuguese flag for Portuguese

bram-pkg:
## 📃 Description

English is clearly not from the US, English is from England, in the United Kingdom. It makes more sense to put the flag of the United Kingdom there.

Same with Portuguese, the language comes from Portugal, so the Portuguese flag should be used.

### ✏️ Changed

- US flag -> British flag
- Brazilian flag -> Portuguese flag

### DIFF
--- a/beszel/site/src/lib/languages.ts
+++ b/beszel/site/src/lib/languages.ts
@@ -87,7 +87,7 @@ export default [
 	{
 		lang: "pt",
 		label: "Português",
-		e: "🇧🇷",
+		e: "🇵🇹",
 	},
 	{
 		lang: "tr",


henrygd:
I'm just having a bit of fun with the flags and would rather not stick to any hard rules.

I have nothing against Portugal or Spain, I just like Brazil and Mexico and they likely make up a larger part of the user base.

Plus if we add Hebrew I'm going to use something like the scroll or temple emoji instead of a flag.

As far as England and USA, both are obscenely evil. But I'm an American with family in Northern Ireland, so the British flag is one of the few flags I find more objectionable than my own. No offense :neutral_face:. If you guys do something good for once in my lifetime, like vote in Corbyn, then I will happily make the switch.

bram-pkg:
Didn't mean to make any political statements with a PR about flags.. just trying to have them be more "correct" honestly, but you do you.

I, too, hope that the people in the UK will vote for a more left wing party in the next general elections.

henrygd:
Yeah no worries, it's a perfectly sensible PR.

Sorry I shit on England so much, I just don't have the opportunity very often :+1:

Issue #1090, opened 2025-08-27
[Bug]: Safari, clicking any system only returns the last one in the list


jcollas:
### Component

Hub

### Description

A Safari issue.  While using the beszel webapp, In any list of systems, clicking one will always return the stats for the last system in the list.  This works properly in Chrome.  Noticed it in 0.12.4.

### Expected Behavior

Clicking on a system should return the stats for the system, not for the last one in the list of systems.

### Steps to Reproduce

1. Use webui in Safari.
2. Click system link, (not the last one)
3. Page shows stats of last system in list

### Category

UI / UX

### Affected Metrics

CPU

### OS / Architecture

linux/amd64

### Beszel version

0.12.4

### Installation method

Docker

### Configuration

```yaml
beszel hub 0.12.4
```

### Hub Logs

```json

```

### Agent Logs

```shell

```

RCourtenay:
Also experiencing in 0.12.5

DYDARK:
I can confirm this behavior on Safari with Beszel Hub 0.12.4 and 0.12.5.
In the table view, clicking on any system always shows the stats of the last system in the list.

When switching to the grid view, everything works correctly – each system opens its own stats page.

So the bug seems limited to Safari + table view, while the grid view works as expected.

mk-73:
Same for me. Grid view works as aspected, table view has the same link (last in the list) for all system, also visible via inspect.

Z1mDMan:
Add me to the list of people noticing this since 0.12.4

henrygd:
Thanks for reporting. This will be fixed in the next release, likely out in the next few days.

Please change to grid view for now or use the command palette to navigate to your system pages.

henrygd:
Issue was closed automatically on merge. Reopening until this has been released to prevent duplicate issues 👍

henrygd:
Should be fixed now in 0.12.6. Let me know you still see it.

DYDARK:
> Should be fixed now in 0.12.6. Let me know you still see it.

Works great now, thanks for the fix!

PR #1092, opened 2025-08-27
[Bug] Fix system table in Safari

svenvg93:
## 📃 Description

This PR fixes the system selection in Safari, that was caused by commit [9d25181](https://github.com/henrygd/beszel/commit/9d25181d1d6a878a61f62f0e221b8fac48c29706)

closes https://github.com/henrygd/beszel/issues/1090


henrygd:
Thanks, I pushed an update to the PR because I think we can fix this with CSS. This also gives us a `safari:` tailwind variant which could be useful in the future.

I was able to reproduce in Gnome Web (Epiphany) and this resolves it. Could you take a look when you have time and verify that it also works in Safari?

svenvg93:
Yess this works in Safari as well 😄

henrygd:
Perfect, thanks!

PR #1103, opened 2025-08-28
Use older cuda image for increased compatibility

Impact123:
## 📃 Description

This changes the docker cuda image from `12.9.1` to `12.2.2` for increased compatibility.
For example, if you install [the `nvidia-driver` package from the debian trixie repos](https://packages.debian.org/en/trixie/nvidia-driver) from their repo [as recommended](https://wiki.debian.org/NvidiaGraphicsDrivers) you get cuda `12.4` and this error.

```
nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.9, please update your driver to a newer version, or use an earlier cuda container: unknown
```

I'm implementing the second suggestion here. [`cuda:12.2.2-base-ubuntu22.04` is supported](https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/supported-tags.md#cuda-1222) and as far as I know (please correct me if I'm wrong) this has no disadvantages for this project.

### ✏️ Changed
- Replace the docker cuda image from `12.9.1` to `12.2.2` for increased compatibility

### DIFF
--- a/beszel/dockerfile_agent_nvidia
+++ b/beszel/dockerfile_agent_nvidia
@@ -15,7 +15,7 @@ RUN CGO_ENABLED=0 GOGC=75 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -ldflags "-
 # --------------------------
 # Final image: GPU-enabled agent with nvidia-smi
 # --------------------------
-FROM nvidia/cuda:12.9.1-base-ubuntu22.04
+FROM nvidia/cuda:12.2.2-base-ubuntu22.04
 COPY --from=builder /agent /agent
 
 ENTRYPOINT ["/agent"]


henrygd:
Thanks, looks like it's a smaller image as well.

Do you know if it works properly with newer hardware since it was released in 2023? I don't have a ton of experience with Nvidia.

Impact123:
Unfortunately I do not know for sure. I was hoping someone else will chime in with that information, hence the `please correct me if I'm wrong`. My newest card that is used with linux is a RTX 3090. Considering that [other projects such as llama.cpp use versions such as `12.4.0`](https://github.com/ggml-org/llama.cpp/blob/a8bca68f727844e7dcf24a956003b3c2039ea563/.devops/cuda.Dockerfile#L3) I figure this should be fine. I'll see if I can find a way to test it with a RTX 4080.

Impact123:
I tested with a RTX 4080 Super and docker desktop on windows now and from what I can tell everything GPU related works fine.
I built an image with this `Dockerfile`
```bash
FROM henrygd/beszel-agent:latest AS beszel
FROM nvidia/cuda:12.2.2-base-ubuntu22.04
COPY --from=beszel /agent /agent

RUN apt-get update && \
    apt-get install -y --no-install-recommends nvidia-container-toolkit && \
    rm -rf /var/lib/apt/lists/*

ENTRYPOINT ["/agent"]
```

```bash
> docker exec beszel-agent apt list -qqi "*cuda*12*"
cuda-compat-12-2/unknown,now 535.129.03-1 amd64 [installed,upgradable to: 535.261.03-0ubuntu1]
cuda-cudart-12-2/unknown,now 12.2.140-1 amd64 [installed]
cuda-toolkit-12-2-config-common/unknown,now 12.2.140-1 all [installed,automatic]
cuda-toolkit-12-config-common/unknown,now 12.3.52-1 all [installed,upgradable to: 12.9.79-1]
```

henrygd:
Perfect, thanks for looking into that. Considering llama.cpp uses `12.4.0`, would it make sense for us to use that version as well?

I'm happy to merge either way. At worst it gives us broader compatibility than we have now. If new issues arise then we'll try to find a middle ground :+1:

Impact123:
I chose this tag because debian 12's driver would be supported as well. Using debian 13 as example was just a better way to show that even a very recently released version of it comes with unsupported drivers.

henrygd:
Sounds good, I appreciate your time and contribution!

Issue #1078, opened 2025-08-26
[Feature]: Show a count of systems at the top of the main page


Justinzobel:
### Component

Hub

### Describe the feature you would like to see

A count of online/total devices.

### Motivation / Use Case

A quick count of systems monitored that can be compared against a known quantity of devices.

For example, our hosting provider can show us we have 40 devices but if we see total monitored devices is 39 we know we're missing one in our monitoring and can easily figure out and deploy the monitor.

### Describe how you would like to see this feature implemented

After `All Systems` in the main page header show how many systems enrolled. This could also show a total of all systems online.

For example, `All Systems - 38/40 Online` with the online count showing as the alert colour if not the same as the total systems monitored.

### Category

Charts & Visualization

### Affected Metrics

Other

henrygd:
This has been added to the 'view' options in 0.12.7:

<img width="732" height="491" alt="Image" src="https://github.com/user-attachments/assets/d22e545e-7cc5-45c4-8748-d8fdc1f603e3" />

Justinzobel:
I was proposing it here: 

<img width="509" height="72" alt="Image" src="https://github.com/user-attachments/assets/63c3a02b-0b30-47c8-9a2f-a67307a82545" />

Justinzobel:
The filter is a great addition though.

teloryfrozy:
> I was proposing it here:
[...]

Issue #1024, opened 2025-08-01
[Feature]: When sorting a column, remember sorting after page refresh


Neoboffin:
### Component

Hub

### Describe the feature you would like to see

When viewing systems under "All Systems" (homepage), you can sort any column. I would like it to _remember_ which column I sorted upon page refresh.

### Motivation / Use Case

My tab is not always open and whenever I navigate away (within Beszel) or close the tab and navigate back to "All Systems" it loses that previous sorting setting forcing me to click-to-sort again.

### Describe how you would like to see this feature implemented

Add the last sorted column to localStorage?

### Category

UI / UX

### Affected Metrics

Other

pbaylas:
Looks like this was implemented by #1110 :tada:

henrygd:
Yes it was, thanks for the reminder! Forgot to close the issue 👍

PR #1110, opened 2025-08-30
Minor UI changes

teloryfrozy:
## 📃 Description

Add the online available systems count and save sorting mode across refreshes

### ➕ Added

- feat #1078 
- feat #1024 

## 📷 Screenshots

image

### DIFF
--- a/beszel/site/package.json
+++ b/beszel/site/package.json
@@ -4,7 +4,7 @@
 	"version": "0.12.5",
 	"type": "module",
 	"scripts": {
-		"dev": "vite",
+		"dev": "vite --host",
 		"build": "lingui extract --overwrite && lingui compile && vite build",
 		"preview": "vite preview",
 		"sync": "lingui extract --overwrite && lingui compile",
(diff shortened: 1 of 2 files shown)


henrygd:
Thank you!

I like the mention of online systems, but it's a bit to eye catching for me. Can we move it to the message below? Let's remove the first sentence and change it to "Click on a system to view information  - 1/5 online" and just leave it the muted color.

For remembering sort state, can we add a `useSessionStorage` hook and use that instead? This will remember it through the session instead of across sessions. For now I think this is the best default behavior. Later I can add an option to use local storage instead.

What do you think?

teloryfrozy:
Yes we can move the message bellow, I just followed the instructions given in the issue. I agree with the fact that only "online" should be colored, not the whole sentence otherwise it will be too eye catching. Particular case: should we display 0/0 online or not ("online" being green)

Between localStorage and sessionStorage to be honest, I would say its better to save it across sessions rather than the session itself and reset each time the page is closed. As it is like a very tiny setting and the people who need that feature have plenty of systems to monitor (which I do not (only 2 servers)). And the initial issuer mentioned closing the tab.

henrygd:
Thanks, this has been merged. I ended up tinkering around and moved the system numbers into the view pane. I also changed it to sessionstorage for now but will add a setting for this later. :+1:

PR #1112, opened 2025-08-30
[Feature] improved support for `mips` and `mipsle` architectures

### LINKED ISSUE #1036: [Bug]: Install script downloads incorrect mips binary on Openwrt (mipsel)
### Component

Agent

### Description

When using install sript it detect architecture using
`ARCH=$(uname -m,`
which returns
`mips` 
on my device (Mi Router 4a Gigabit). My knowlegde is very limited here, but with help of AI i was able to confirm, that `opkg print-architecture`
returns:
```
arch all 1
arch noarch 1
arch mipsel_24kc 10
```

I'm not sure if i needed, but i compiled beszel for mipsle, and it now works. I'm not sure what happens when it runs autoupdate.

### Expected Behavior

Script downloads mipsle binary

### Steps to Reproduce

1. Run the install script.
2. Observer errors related to incorrect binary

### Category

Installation

### Affected Metrics

Other

### OS / Architecture

mipsle

### Beszel version

0.12.3

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

a-mnich:
## 📃 Description
This PR should improve the support for mostly older devices with mips (little- and big endian) CPUs often used with OpenWRT.
Until now `mips` based devices installed binaries for the `mipsle` architecture  (= `mips little endian` sometimes also referred to as `mipsel`) which isn't compatible with `mips` as it uses the opposite byte significance order.
Additionaly quite a lot of the mips/miplse embedded devices don't come with with hardware floating point capabilities, therefore I switched to software based floating point for the `mipsle` and newly added `mips` architecture.

resolves #1036

Successfully tested on [AVM 300E (mips)](https://openwrt.org/toh/hwdata/avm/avm_fritz_wlan_repeater_300e) and [D-Link COVR X1860 (mipsle)](https://openwrt.org/inbox/toh/d-link/x1860_a1)

### ➕ Added

- support for `mips` architecture

### ✏️ Changed
/
### 🔧 Fixed

- switched `mipsle` to `softfloat` (#1036)

### DIFF
--- a/beszel/.goreleaser.yml
+++ b/beszel/.goreleaser.yml
@@ -38,6 +38,7 @@ builds:
       - mips64
       - riscv64
       - mipsle
+      - mips
       - ppc64le
     gomips:
       - hardfloat
@@ -52,6 +53,9 @@ builds:
         gomips: softfloat
       - goos: linux
         goarch: mipsle
+        gomips: hardfloat
+      - goos: linux
+        goarch: mips
         gomips: hardfloat
       - goos: windows
         goarch: arm
--- a/supplemental/scripts/install-agent.sh
+++ b/supplemental/scripts/install-agent.sh
@@ -398,7 +398,7 @@ fi
 echo "Downloading and installing the agent..."
 
 OS=$(uname -s | sed -e 'y/ABCDEFGHIJKLMNOPQRSTUVWXYZ/abcdefghijklmnopqrstuvwxyz/')
-ARCH=$(uname -m | sed -e 's/x86_64/amd64/' -e 's/armv6l/arm/' -e 's/armv7l/arm/' -e 's/aarch64/arm64/' -e 's/mips/mipsle/')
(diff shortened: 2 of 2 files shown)


a-mnich:
It looks like this change broke the auto installation for `mipsle` devices, as they are now downloading the mips binaries instead.
Although the binaries for both architectures work fine, it appears that `uname -m` returns `mips` even on actual `mipsle` devices.

While I tested the binaries on both architectures when creating the PR, I must have mixed something up. Either I tested the architecture detection logic on the wrong device or forgot to remove the regex replacement (-e 's/mips/mipsle/') on the correct one.

**mipsle device:**
```
root@OpenWrt-AP1:/tmp# uname -m
mips
root@OpenWrt-AP1:/tmp# opkg print-architecture
arch all 1
arch noarch 1
arch mipsel_24kc 10
```
**mips device:**
```
root@FritzRepeater300E:~# uname -m
mips
root@FritzRepeater300E:~# opkg print-architecture
arch all 1
arch noarch 1
arch mips_24kc 10
```
While it would be possible to use `opkg` for detection on OpenWRT devices, I will have to look up if there is a way to do this generally for all linux devices.

PR #1166, opened 2025-09-13
Fixing helm chart service/ingress

twentybit:
## 📃 Description

The helm chart doesn't work with an ingress as the service name hard coded `-web`, which can never match the ingress rule.

I've also made the default values less opinionated.

### 🔧 Fixed

- The service name in the helm chart could never match the ingress rule.
- Removed opinionated values in helm chart


henrygd:
Thank you! :+1:

Issue #1190, opened 2025-09-23
[Bug]: Too many network interfaces break UI layout


DmitryMigunov:
### Component

Hub

### Description

When running the Beszel agent on a server with Kubernetes, the agent reports a very large number of network interfaces. This causes the UI layout to break, making it difficult to read and navigate.

### Expected Behavior

The UI should gracefully handle cases where there are many interfaces (e.g., with pagination, collapsing/expanding sections, or some kind of filtering).

### Steps to Reproduce

1. Install Beszel agent on a server running Kubernetes.
2. Open the UI and check the list of network interfaces.
3. Notice that the large number of interfaces causes layout issues.

### Category

Charts & Visualization

### Affected Metrics

Network

### OS / Architecture

6.6.106-flatcar

### Beszel version

0.12.10

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

DmitryMigunov:
<img width="2313" height="1316" alt="Image" src="https://github.com/user-attachments/assets/9aafe6df-df51-4d8a-82d3-c40b677d81d4" />

henrygd:
These are internal virtual networks for Kubernetes, right? Is it okay if we exclude tracking `cali*` interfaces by default, or is this data useful in some way?

DmitryMigunov:
Yes, you are right — these are internal virtual networks created by Kubernetes.
Instead of hardcoding a filter only for `cali*` interfaces, it would be more flexible to support configurable pattern matching. Different CNI plugins (for example, Cilium) use different naming conventions for their virtual interfaces, so having a user-defined pattern filter would allow us to exclude the irrelevant ones while keeping the important data visible.

henrygd:
I changed `NICS` in the next release to have pattern matching and blacklist functionality. We've also added `cali*` to the default ignore list (let me know if you think we should still include it).

`NICS` is still treated as a whitelist by default. Can be used as a blacklist by prefixing with `-`.

| `NICS` value   | Mode      | Action                                                    |
| -------------- | --------- | --------------------------------------------------------- |
| `foo_*`        | Whitelist | Only interfaces matching `foo_*` are allowed.             |
| `foo_1,bar_*`  | Whitelist | Only `foo_1` and `bar_*` interfaces allowed.              |
| `-foo_*`       | Blacklist | Excludes interfaces matching `foo_*`; all others allowed. |
[...]

PR #1195, opened 2025-09-23
[Bug] Skip Cali interfaces for network collection

svenvg93:
## 📃 Description

This PR adds `cali` interfaces for Kubernetes  to the skip list, same as we do for docker interfaces 

fixes https://github.com/henrygd/beszel/issues/1190

### ✏️ Changed

- Add `cali` to the skip list

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -64,6 +64,14 @@ jobs:
             username: ${{ github.actor }}
             password_secret: GITHUB_TOKEN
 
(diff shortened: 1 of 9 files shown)


henrygd:
Thanks! I asked the OP if it's okay to exclude these or if they want that data for some reason. I'll merge if they're okay with excluding them :+1:

svenvg93:
Sounds good! I think we should exclude them to be inline with Docker interfaces. We can add them later in an seperate Kubernetes monitoring

PR #1216, opened 2025-09-29
Modify GPU usage section in readme

oregonpillow:
- Synced section as currently written in https://beszel.dev/guide/what-is-beszel#supported-metrics 
- might want to consider re-using the markdown in future to prevent them deviating. Or personally, I would remove all documentation from README, and use README only for screenshots and "marketing" to funnel people to the docs site.


henrygd:
Thanks a lot, totally overlooked that :+1:

PR #1235, opened 2025-10-04
Fix #1214: Add VOLUME ["/beszel_data"] to preserve config across container recreations

### LINKED ISSUE #1214: [Feature]: Add VOLUME to Dockerfile for Persistent Probe Data
### Component

Hub

In the remote control domain—such as with Sunlogin, TeamViewer, etc.—device IDs are typically derived from hardware fingerprints. This ensures that the device ID remains nearly unchanged even after a system reinstallation.

### Motivation / Use Case

Currently, Beszel does not run in privileged mode, and because users often omit the -v volume mount (e.g., -v beszel_data:/beszel_data) when deploying the probe, the environment configuration is lost every time the container is upgraded or recreated.

To avoid this unnecessary inconvenience—especially during automated upgrades or when users forget to manually specify a volume mount—I’d like to add the following line in the Dockerfile (note: not a .yaml file) before building the next version.

```dockerfile
VOLUME ["/beszel_data"]

This will ensure Docker automatically creates a persistent volume for /beszel_data, preserving critical configuration across container recreations.

### Screenshots

_No response_

### Category

Settings & Configuration

### Affected Metrics

CPU

### PR DESCRIPTION
**Pull request by Marrrrrrrrry**
## Description

This PR addresses issue #1214 by adding built-in data persistence to Beszel's agent containers.

Many users deploy Beszel with automated container update tools like Watchtower. If they forget (or don't know) to mount a volume for `/beszel_data`, all configuration is lost every time the container is automatically replaced — leading to unexpected resets and a poor user experience.

To prevent this, i add the following line to all agent Dockerfiles:

```dockerfile
VOLUME ["/beszel_data"]

### DIFF
--- a/internal/dockerfile_agent
+++ b/internal/dockerfile_agent
@@ -23,4 +23,7 @@ COPY --from=builder /agent /agent
 # this is so we don't need to create the /tmp directory in the scratch container
 COPY --from=builder /tmp /tmp
 
+# Ensure data persistence across container recreations
+VOLUME ["/beszel_data"]
+
 ENTRYPOINT ["/agent"]
--- a/internal/dockerfile_agent_intel
+++ b/internal/dockerfile_agent_intel
@@ -22,4 +22,7 @@ COPY --from=builder /agent /agent
 
 RUN apk add --no-cache -X https://dl-cdn.alpinelinux.org/alpine/edge/testing igt-gpu-tools
 
+# Ensure data persistence across container recreations
+VOLUME ["/beszel_data"]
+
 ENTRYPOINT ["/agent"]
--- a/internal/dockerfile_agent_nvidia
+++ b/internal/dockerfile_agent_nvidia
@@ -24,4 +24,7 @@ COPY --from=builder /agent /agent
 # this is so we don't need to create the /tmp directory in the scratch container
 COPY --from=builder /tmp /tmp
 
+# Ensure data persistence across container recreations
+VOLUME ["/beszel_data"]
+
 ENTRYPOINT ["/agent"]
--- a/internal/dockerfile_hub
+++ b/internal/dockerfile_hub
@@ -25,6 +25,9 @@ FROM scratch
 COPY --from=builder /beszel /
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 
+# Ensure data persistence across container recreations
+VOLUME ["/beszel_data"]
+
 EXPOSE 8090
 
 ENTRYPOINT [ "/beszel" ]

### CODE REVIEW

**Comment by henrygd** [comment]
Good idea, thanks!

`VOLUME ["/beszel_data"]` should work for the hub, but the agent uses `/var/lib/beszel-agent` by default, so I think we need to change to that.

Just to clarify, you're not losing your env vars on the agents, right? The problem is that all the hub's data is reset on container recreation?

**Comment by Marrrrrrrrry** [comment]
You're absolutely right. I took a closer look at the documentation for both Hub and Agents, and indeed the paths are different. When I submitted the issue, it was midnight local time, so I was a bit careless.  

Additionally, I've opened a new suggestion in #1236. However, implementing that refactor would involve a significant amount of work. I hope you could review the solution I've already prepared for the current issue (posted under #1236) and consider including this fix in the development roadmap.

**Comment by henrygd** [comment]
Thanks, no worries! I appreciate the contribution :+1:

**Comment by parkr** [comment]
Hey, it looks like this modified both agent and hub Docker images, even though the docker-compose instructions mount this path as a volume for only the hub. I prefer to mount volumes from my host to prevent docker volumes from (eventually) filling my disk, so I'm wondering if there's a plan to migrate the agent to this same path in the future? Could I follow an issue or something? Apologies, but I couldn't find one.

This is [the documentation](https://beszel.dev/guide/hub-installation) I was referring to:

```
services:
  beszel:
    image: henrygd/beszel:latest
    container_name: beszel
    restart: unless-stopped
    ports:
      - 8090:8090
    volumes:
      - ./beszel_data:/beszel_data # hub uses /beszel_data so no extra VOLUME created by the image
      - ./beszel_socket:/beszel_socket

  beszel-agent:
    image: henrygd/beszel-agent:latest
    container_name: beszel-agent
    restart: unless-stopped
    network_mode: host
    volumes:
      - ./beszel_agent_data:/var/lib/beszel-agent # agent does NOT use /beszel_data so now there's an extraneous VOLUME
      - ./beszel_socket:/beszel_socket
    environment:
      LISTEN: /beszel_socket/beszel.sock
      HUB_URL: http://localhost:8090
      TOKEN: <token>
      KEY: "<key>"
```

Marrrrrrrrry:
I made a mistake in the submission, but this issue was fixed before merging.

parkr:
Thanks! Can you point me to that fix? The diff in this PR still shows `VOLUME ["/beszel_data"]` for all images. Perhaps it's a GitHub problem, but it's showing the same path for all images. Apologies if I'm missing something obvious!

henrygd:
I fixed here: https://github.com/henrygd/beszel/commit/898dbf73c8a75c9d15c01856d9470cc40c87ef97

parkr:
Thank you!

PR #1245, opened 2025-10-06
OpenWRT - graceful service stop, restart and respawn if crashes

evrial:
### 🔧 Fixed
- This message:
```
Restarting beszel-agent via procd…
Terminated
```
When you use USE_PROCD=1, procd automatically tracks and stops the process it started.
So normally, you don’t need to manually kill the process — procd does that cleanly on stop_service().
If you override stop_service() with killall, you’re:
Bypassing procd’s PID tracking
Potentially killing unrelated instances
Losing logging and restart behavior consistency

- Added respawn if crashes (default 5s, 5 attempts in 5min)


henrygd:
Perfect, thanks again for your contributions :)

PR #1250, opened 2025-10-07
Retry agent's attempt to get the Docker version

amanda-wee:
## 📃 Description

While working on a new feature for wollomatic/socket-proxy, I discovered that there are situations where the socket proxy might not be ready to accept Docker API requests from the Beszel agent, yet the Beszel agent makes a request for the Docker version. Unfortunately, since there is no retry logic, the failed request sticks, and consequently the Beszel agent fails to retrieve Docker stats even though the socket proxy becomes ready soon after.

This PR implements retry logic for the Docker version request with a 5 second wait. I have implemented it as a loop in anticipation that the maximum number of tries or the wait between tries might change or even be configurable in the future.

### ✏️ Changed

- add retry logic to Docker API request for Docker version

### DIFF
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -479,7 +479,21 @@ func newDockerManager(a *Agent) *dockerManager {
 	var versionInfo struct {
 		Version string `json:"Version"`
 	}
-	resp, err := manager.client.Get("http://localhost/version")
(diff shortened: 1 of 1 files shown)


henrygd:
Makes sense, thanks :+1:

PR #1280, opened 2025-10-15
Add a total line to the tooltip of charts with multiple values 

titouv:
## 📃 Description

Add a total line to charts displaying multiple values.

SCR-20251015-rtbr

Active for charts ✅
  - docker CPU usage
  - memory usage
  - docker memory usage
  - disk I/O
  - bandwith
  - docker network I/O

Disabled for charts ❌ (because represent only one value or aggregation does not make sense)
  - cpu usage
  - disk usage
  - load average
  
## Explanation

I noticed that I sometimes want to quickly see the total RAM usage at a specific point in time on my Beszel dashboard. However, my chart looks like the one below. Therefore, having this value readily available would be useful.

image

## 📖 Documentation

Not yet done

## 📷 Screenshots

SCR-20251015-rtbr


henrygd:
Thanks, makes sense and looks good. I will try to merge before the next release.

henrygd:
Merged, thank you :+1:

Issue #1254, opened 2025-10-07
[Bug]: no more battery graphs since V0.13


C2gl:
### Component

Hub & Agent

### Description

there used to be battery graphs in beszel for laptop and battery enabled devices on the beszel hub site 

but (i think) ever since V0.13.0 or 12.9 these graphs have vanished 

### Expected Behavior

battery graph for battery enabled devices in beszel hub 

### Steps to Reproduce

update beszel agent/hub to V0.13.0 and see the lack of battery graphs 

reinstall beszel agent and hub to v0.12.8 and see there is a battery graph

### Category

Metrics

### Affected Metrics

Sensors

### OS / Architecture

windows

### Beszel version

v0.13.1

### Installation method

Docker

### Configuration

```yaml
installed on windows as binary using NSSM
```

### Hub Logs

```json
found no errors in logs
```

### Agent Logs

```shell

```

nsheaps:
Not sure if this helps, but I just set up Beszel agent on ubuntu 24.04 (v0.13.2). For the initial period where the agent was operating it was reporting battery, both while charging/charged and while depleting. Upon coming back to my computer this morning (don't recall but likely post restart), the agent is running, but the binary is no longer in the path and not reporting battery %. Other generic system metrics are reporting fine.

When viewing a 24 hourperiod where there was battery data prior, the hub _does_ show the data from that time. **However** when there is data within 12 hours, viewing it at the 12 hr period does not include the chart. New data does not appear after restarting the agent with `systemctl restart beszel-agent`. It seems more like it's not being reported anymore, but also potentially an issue in the Hub UI/API as well.

EDIT: As I was collecting this data, it now no longer shows battery data for 24hr and now only for 1wk

EDIT2: I'm guessing unrelated, but this error also appears when checking for agent health:

```bash
# > /opt/beszel-agent/beszel-agent health
2025/10/16 19:15:48 stat /tmp/beszel_health: no such file or directory
```

EDIT3: This may be related to this other issue: https://github.com/henrygd/beszel/issues/1170

the `battery` package says it should be similar to `acpi -b`, here's the output of both on my machine:

```bash
# > battery
BAT0: Discharging, 98.55%, 8h58m26.382s remaining [Voltage: 17.27V (design: 15.48V)]
[...]

Issue #1170, opened 2025-09-15
[Bug]: Battery stats shown on iMac system without battery


blackketter:
### Component

Agent

### Description

Monitoring an M1 iMac (which does not have a battery), but a battery statistic is shown with "Current state: Empty".  

### Expected Behavior

No battery field should be shown

### Steps to Reproduce

Run agent on an iMac

### Category

Charts & Visualization

### Affected Metrics

Other

### OS / Architecture

macOS

### Beszel version

0.12.6

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

henrygd:
Thanks for reporting, I changed battery detection in the next release.

Upstream issue: https://github.com/distatus/battery/issues/34

henrygd:
Let me know if you still see this after upgrading the agent to 0.12.8.

No rush 👍

blackketter:
I've updated the agent and restarted the hub process on the other machine.  

In the 1hr view the battery chart is missing, switching to the default 24hr view shows the ghost battery.  

Not sure if this is the expected behavior.

henrygd:
Yes, thanks for confirming. The historical data has already been created but will clear out eventually.

PR #1287, opened 2025-10-17
fix: send battery stats even if some batteries are missing stats

nsheaps:
Added error reporting for battery information retrieval.

Handles cases like bluetooth devices where `battery` finds it but it has incomplete information about that battery, which would have otherwise caused a null pointer error.

This works by stating that the device does indeed have a battery as long as the len of the list of batteries is not 0, ignoring errors. Error checking is deferred to actual metric calculation.

**NOTE** Also changes the percentage calculation to utilize the "Full" attribute, instead of the Designed capacity. For some reason, this number was flapping and didn't align with the actual percent. Using the `battery` cli tool (uses [the same underling lib](https://github.com/distatus/battery/blob/master/cmd/battery/main.go)), it utilizes only Full, and doing so eliminates the flapping.

Related: #1254

### 🔧 Fixed

- Corrected battery metrics calculations to handle devices that don't report full metrics (like some bluetooth mice)

### Tested by

1. Stop the service on my machine `sudo systemctl stop beszel-agent`
2. Build the code using `go fmt && make`
3. move the built agent binary from ./build to /opt/beszel-agent/beszel-agent
4. restart the agent
5. note battery % on hub appears, matches `battery` cli and system tray %, works with bluetooth mouse attached.

### DIFF
--- a/agent/battery/battery.go
+++ b/agent/battery/battery.go
@@ -68,7 +68,7 @@ func GetBatteryStats() (batteryPercent uint8, batteryState uint8, err error) {
 	}
 
 	if totalCapacity == 0 {
-		// for mac's there's sometimes a ghost battery with 0 capacity
+		// for macs there's sometimes a ghost battery with 0 capacity
 		// https://github.com/distatus/battery/issues/34
 		// Instead of skipping over those batteries, we'll check for total 0 capacity
 		// and return an error. This also prevents a divide by zero.


henrygd:
Thanks! 

Not sure if you've seen this, but there's an issue in the battery library where it returns a ghost battery on macOS: https://github.com/distatus/battery/issues/34

This was brought up in an issue here (#1170) and is why we don't just trust the list length. I don't have a macOS device so I can't debug it further. From the changes it seems like this would reintroduce the problem, but maybe I'm wrong. Any chance you have a device where you could replicate this with the CLI so we can add a proper workaround?

Also I think line 25 should be `err != nil`, right?

nsheaps:
Previously in the implementation, I saw that you were checking for zero capacity and then saying it has no batteries at all. In the way it is currently calculated, where the percentage is of total charge of all batteries detected by the library, if the values were zero then the numbers would still be at zero. The issue that occurred at that point was it then tried to divide by zero to get the percent, which raises an error (and was thrown away). If it did manage to get set, it would still say "your computer has a battery" with zero capacity, and then show(? still curious how `0/0` is computeable) a battery when the imac had none. I had made an assumption that those batteries had a "partial data" error that would be skipped by the error handling logic I [added](https://github.com/henrygd/beszel/pull/1287/files#diff-5f55a7cfec6adc25e6a344524352738ab9ea460417abe43178f1ad2bd75ed01fR57).

In reviewing the past code, linked issues, and what I implemented a second time, I came to the same conclusion that you did, that it'd likely reintroduce the issue. I unfortunately also don't have a mac at the moment, but I updated the logic with 3 changes:
1. Now the stats are only set if there's no error.
2. If we compute them and the full capacity of any battery is 0, that battery is skipped (which fixes the issue, since that ghost battery has a full size of 0.)
3. If we finish adding up capacity of all batteries and the total size is still 0, then we'll return the stats with an error (and thus won't get set).
	- I am wondering if the battery data lowers the "full" value as a battery ages, and if a 0 value is possible.
 
That should prevent the issue from reoccurring.... _however_

In implementing these fixes, I discovered a new issue and the fix could be a potential feature improvement:

Whether or not a system has a battery is cached, which can prevent sending of battery data for new batteries if they're added after start.

To resolve that, we could change the schema of the system stats to just support sending all battery data individually, and the cached boolean value is no longer needed since the list size is treated as dynamic anyway. Doing this also allows the web UI to show each battery individually, both by % and by charge mAh. This additional data can allow users to do things like alerting if their wireless mouse battery is low, and keeping the overall charge of the system accurate (by not including other batteries). I do still think that it's worth it to do some massaging locally so that we're not sending battery data for devices where there is no numeric metrics (like the ghost battery, or my mouse, which it errors on for incomplete data, but still can get the name). It might be worth inspecting the list of batteries anyway, since the alternative is making sure we're getting the system battery or nothing (my bluetooth earphones report battery % to ubuntu but `battery` can't find it).


In any case, @henrygd this should be ready to review, just tagging you since I can't do that via the PR mechanism

henrygd:
This is merged, thanks again.

I didn't realize the library could pick up battery level from peripherals. I thought it was just system batteries, which is why we just sum everything.

Listing the batteries separately in that case would be better (or putting them in one line chart). I'll add it to my list and check if I have any peripherals that work with it.

PR #1307, opened 2025-10-21
[Feature] Let y axis for temperature not start at 0

### LINKED ISSUE #1306: [Feature]: Decrease range temperature graph
### Component

Hub

Decrease the displayed range on the temperature graph, to only focus on the operating area. Being able to see from 0 degrees and up isn't super useful, in my opinion, and I would personally rather have it focus on the operating area.

### Motivation / Use Case

The change would make the temperatures a little easier to inspect and keep an eye on. 

The display range could be +2 degrees above highest temperature and below lowest temperature.
This way it's easier to see a more detailed view of the temperatures.

There could possibly be a toggle between the current range and the new range, if any users should prefer that.

### Screenshots

A lof of wasted space, and not a very clear view of the temperatures
Image

### Category

Metrics

### Affected Metrics

Sensors

svenvg93:
## 📃 Description

This PR adjust the temperature chart to not start at 0. Now the changes in temperature are better visible.  

Closes #1306 

Scherm­afbeelding 2025-10-21 om 19 37 23

## 📷 Screenshots

If this PR has any UI/UX changes it's strongly suggested you add screenshots here.

### DIFF
--- a/internal/site/src/components/charts/temperature-chart.tsx
+++ b/internal/site/src/components/charts/temperature-chart.tsx
@@ -49,31 +49,6 @@ export default memo(function TemperatureChart({ chartData }: { chartData: ChartD
 
 	const colors = Object.keys(newChartData.colors)
 
-	/** Calculate temperature domain to focus on operating range */
-	const tempDomain = useMemo((): [number, number] | [number, "auto"] => {
-		if (newChartData.data.length === 0) {
-			return [0, "auto"] as [number, "auto"]
-		}
-		let minTemp = Infinity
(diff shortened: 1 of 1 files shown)


henrygd:
Thanks! Looking at the docs, I'm wondering if using `domain={['auto', 'auto']}` or `domain={['dataMin - 5', 'dataMax + 5']}` would work for this instead of calculating ourselves?

https://recharts.github.io/en-US/api/YAxis#domain

svenvg93:
You are absolutely right `domain={['auto', 'auto']}` did the trick as well.

Issue #1262, opened 2025-10-09
[Feature]: Cache measurements on agent


ysmilda:
### Component

Agent

### Describe the feature you would like to see

The ability to let the agent cache its measurements while unable to connect to the Hub. Once the connection is restored the cache would be send to the Hub to backfill the missing data.

### Motivation / Use Case

The use case here is that we have several nodes that are not always connected to a network, but we would like to have centrally collected historical data from.

### Describe how you would like to see this feature implemented

An opt-in environment flag that stores the measurements in a buffer while the connection to the Hub is severed. Once the connection is back up the buffer is emptied towards the Hub.

### Category

Metrics

### Affected Metrics

Other

ysmilda:
@henrygd I see you've put this on the backlog. If that means that this is something you'd like to support, I would be able to do a proposal for what this change would entail as well as implement it.

henrygd:
This is something I've wanted to add for a while but is behind a handful of other things right now.

My thought was to give agents their own SQLite or BadgerDB database to store core metrics every minute.

Right now an agent doesn't do anything unless it gets a request from the hub. We'd need to change this so the agent independently updates the core stats every minute and returns hub requests for the 1m stats from the db.

This also opens the door to another big feature. Right now granularity decreases as the chart time frame increases. With agents storing 1m records, we can eventually allow users to select a custom time window and stream in 1m records on fly from the agent.

Let me know if this lines up with what you're thinking.

ysmilda:
This lines up pretty well with what I wanted to propose. Where my thoughts differ is that I originally envisioned a simple in memory cache that gets drained once the connection to the hub is restored. This way the resource footprint of the agent stays as small as possible.

A couple things need to change as you mentioned;
 - The agent retrieves its metrics on a set interval (settable via env?).
 - The agent pushes the metrics to the hub while the connection is up. If a timestamp is added to the `CombinedData` struct we can just push the whole cache on a reconnect and have the hub store it.
 - The agent needs to keep track of the connection state to the hub. If it is down we start caching when enabled (env).
 - The hub needs to be able to temporarily override the retrieval interval of the agent.

What we need to decide on is if the caching needs to survive the agent going down, and if it keeps its cache once it has been uploaded to the hub.

henrygd:
You're right, I think in-memory caching fine for this purpose.

The agent does keep track of it's connected state, so we can start a timer on a disconnection event.

[...]

PR #1153, opened 2025-09-08
[Feature] Basic systemd service monitoring

smtucker:
This introduces a new feature to monitor systemd services on Linux hosts. The agent now collects service statuses and sends them to the hub, where they are displayed on the system detail page and summarized on the main dashboard.

Key Changes:

- Agent:
  - The agent now collects systemd service names and their statuses (active, inactive, failed, etc.).
  - It gracefully handles permissions by attempting to connect to the system-wide systemd instance first, then falling back to the user-level instance if necessary. This ensures functionality whether the agent is run as root or a standard user.

- System Detail Page:
  - A new "Systemd Services" card displays a detailed list of all services.
  - The view is collapsed by default, showing only failed services for immediate attention. An expander reveals the full list.
  - Services are sorted to prioritize failed ones, and each status is color-coded with an indicator dot for quick visual assessment.
  - A filter allows users to search for specific services by name.

- Dashboard:
  - The main systems table now includes a "Services" column.
  - This column displays the number of failed services in red or a green checkmark if all services are running correctly, providing an at-a-glance health check.
  - The column header now uses a distinct ListChecks icon for better clarity.

20250907_23h48m26s_grim

20250907_23h44m19s_grim

PR #1317, opened 2025-10-24
Add offline caching to the agent

ysmilda:
## 📃 Description

This pull request adds offline caching capabilities to the agent as discussed in https://github.com/henrygd/beszel/issues/1262. The goal is to keep gathering statistics while the agent cannot reach the hub. Once the connection is restored the backlog is committed to the hub.

Currently there is a basic cache implemented for the agent. What still needs to be done before this can be merged:

- [ ] Add support to the hub for processing multiple statistics at once without breaking compatibility with older clients.
- [ ] Check if alerts need to be changed to timestamp as well.
- [ ] Add graph to the hub for showing connection status over time. To indicate which measurements were live and which were cached.
- [ ] Add tests.

Closes https://github.com/henrygd/beszel/issues/1262

### ➕ Added

- Offline cache to the agent
- `Timestamp` field to the `CombinedData` struct
- `Timestamp` field to the database instead of using the `created` field. This allows the data to be backfilled.

### ✏️ Changed

- `AgentResponse` `SystemData` field is now an array of `[]*system.CombinedData`. This is a breaking change for older agents.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -40,6 +40,7 @@ type Agent struct {
 	systemInfo                system.Info                                           // Host system info
 	gpuManager                *GPUManager                                           // Manages GPU data
 	cache                     *systemDataCache                                      // Cache for system stats based on cache time
+	offlineCache              *systemOfflineCache                                   // Cache for offline data storage
 	connectionManager         *ConnectionManager                                    // Channel to signal connection events
 	handlerRegistry           *HandlerRegistry                                      // Registry for routing incoming messages
 	server                    *ssh.Server                                           // SSH server
@@ -53,8 +54,9 @@ type Agent struct {
 // If the data directory is not set, it will attempt to find the optimal directory.
 func NewAgent(dataDir ...string) (agent *Agent, err error) {
 	agent = &Agent{
-		fsStats: make(map[string]*system.FsStats),
-		cache:   NewSystemDataCache(),
+		fsStats:      make(map[string]*system.FsStats),
+		cache:        NewSystemDataCache(),
+		offlineCache: NewSystemOfflineCache(),
 	}
 
 	// Initialize disk I/O previous counters storage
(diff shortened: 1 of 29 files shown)


henrygd:
Awesome, thanks very much!

It may be a week or two before I can get to this as I'm trying to finish up with S.M.A.R.T. support and have a couple other long-waiting PRs at the top of my queue.

Don't worry about any merge conflicts that may be introduced in the meantime :+1:

ysmilda:
The PR is now functional. It does introduce a breaking change within the `AgentResponse` which would require the update of the hub and agent simultaneously. There are some workarounds possible, but none particularly elegant.

While I made the argument for just an in memory store, I would like to revisit that. It is not guaranteed the agent will get a connection to the hub before a shutdown. So adding optional persistency seems to be a nice feature. Introducing an option to cache to sqlite wouldn't be too hard to do.

henrygd:
Thanks! Should be able to get to this during this upcoming week.

henrygd:
Apologies, I'm still wrapped up with #1153 and a couple small things. I will switch this to in-progress and try to get to it as soon as I can.

ysmilda:
No worries, if you'd have the time to do a review or give some pointers I would be more than happy to continue the development and carry some of the load.

ysmilda:
@henrygd Any chance we can get this moving again?

henrygd:
Sorry for the wait and lack of updates!

I have a good idea of what I want to do with this, and made some precursor architectural tweaks. Then had to pause work on the whole project for a bit, and I'm still working through a pile of smaller things that accumulated.

My plan is to get https://github.com/henrygd/beszel/pull/539 added for 0.19 (it's been waiting for nearly a year, and I finally have a Proxmox machine). Then this is the top priority for 0.20.

ysmilda:
Great to hear it's on the roadmap. If at that time you could use some help I will be available!

henrygd:
Thanks, I'll let you know!

Issue #770, opened 2025-04-21
[Feature]: Exclusion list for docker containers


Jisagi:
### Describe the feature you would like to see

Hi, I was wondering if it is somehow possible to exclude specific containers (e.g. by name/prefix) from being monitored. I did see https://github.com/henrygd/beszel/issues/25 where a ui filter was proposed, but by use case is somewhat different. I have gitlab runners being created and deleted every now and then. They only live for a few minutes, sometimes only a few seconds and then get removed again. The beszel agent picks them up and often throw errors, because they are removed too fast:
`Error getting container stats err="runner-smubnyrhe-project-9-concurrent-0-cf39553b288f203e-predefined - no memory stats - see https://github.com/henrygd/beszel/issues/144"`

The error itself is fine and should certainly stay, but the agent picking them up all the time is especially "annoying" when they show up in the ui with their super long name and making the hover popup completely unreadable when you pan over it. They show up, have a name wider than the ui element and the percentages on the right side start to disappear.

![example](https://i.imgur.com/TaYQE2E.png)

### Describe how you would like to see this feature implemented

Having some kind of exclusion/filter list for docker containers to not get stats from would therefore be really nice to have, be it as env vars, configurable list in the settings or anywhere else. The list could not only contain exact matches but maybe also regex or any other matcher type logic.

karanravindra:
+1

henrygd:
I ran into a similar issue with Coolify because it appends a UUID to every container. In the next release I'll truncate the name if it's over a certain width.

A whitelist / blacklist is a good idea so I added it to the roadmap.

Jisagi:
The truncation works perfectly, thanks for the (temporary) fix

![img](https://i.imgur.com/ZrSApb8.png)

svenvg93:
@henrygd We can update the agent to filter monitored containers based on labels. If a container has the label `BESZEL_METRICS_INCLUDE=false`, it will be excluded from monitoring.

Jisagi:
Not a bad idea in most cases, but in my case it won't work. The containers I want to filter are automatically started and destroyed without any possible way to give them a label. Not sure if it would be too much for both solutions to be implemented, because the label one would indeed be easier for a self managed container/stack.

edit: the containers are gitlab runner containers that are spawned for each pipeline job. The most I have are the monthly deployments that start around 35-40 individual jobs in total and each one gets its own container

svenvg93:
[...]

PR #1352, opened 2025-10-30
[Feature] Add env var to exclude containers from being monitored

svenvg93:
## 📃 Description

The agent now supports excluding containers based on name patterns using the `CONTAINER_EXCLUDE `environment variable. For example: `CONTAINER_EXCLUDE="runner-,temp-,test-,jenkins"`

## Documentation 

https://github.com/henrygd/beszel-docs/pull/29

### ➕ Added

- closes https://github.com/henrygd/beszel/issues/770

### DIFF
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


henrygd:
Thanks a lot, these smaller PRs will be a lot easier to manage :)

For this one it might be best to allow the use of pattern matching with `path.Match` like we do with `SENSORS` and `NICS`. We don't need the whole whitelist / blacklist thing, but allowing wildcards will make it a little more flexible and less prone to error than `strings.Contains`.

I can add this if you don't have time.

svenvg93:
Should have started with smaller PRs :) 

You are right `path.Match` is better, I updated the code. Will make the PR for the docs as well.

henrygd:
Thanks a lot! I did change the name to `EXCLUDE_CONTAINERS` just because I like the sound of it a little better :)

svenvg93:
Thanks for implementing  this in 0.15.3 @henrygd. Would you mind to close the linked issue so the original requester knows it's implemented as well.

Issue #1257, opened 2025-10-08
[Feature]: Per core monitoring / CPU hog detection


iconoclasthero:
### Component

Hub & Agent

### Describe the feature you would like to see

Per discussion on Matrix ca. 2025-09-28/29:
Would like to be able to monitor usage of individual cores and alert on high usage (per core monitoring).  Additionally, it would be nice to be able to see processes that are running at moderate but concerning levels for long periods of time (CPU hog detection).

I've worked on some scripting using some linux utilities and system files (ps, mpstatus, etc.) but just to have something for alerts, particularly on the per/core monitoring.

### Motivation / Use Case

I found a process that had been running for weeks using the entirety of a single core which would not have triggered a CPU alert because it was only on one core:
`
```
$ ps -fp 614623
UID          PID    PPID  C STIME TTY          TIME CMD
$USER     614623 99.7 0.0 270432 30776 ? R Aug19 54552:15 mplayer -vo dummy -ao dummy -identify ./.var/app/com.usebottles.bottles/data/bottles/bottles/eac2/drive_c/users/steamuser/Temp/eac test audio data.wav.flac...
```
and would like to be alerted for something like this.

Further, I was poking around and found that a docker service was restarting itself and consuming moderate but significant (10-15% CPU) for no good reason.  I also didn't realize how much a script I ran on screen (i.e., not being watched) was consuming because it piped display through lolcat.  

Having some better awareness of these issues would be helpful in reducing CPU usage.

### Describe how you would like to see this feature implemented

WRT to the CPU hog detection:  checking pid usage would be helpful, but with the lolcat example, it would probably need to be tracked by the `comm` name or a hash of the binary to be useful.

I'm not entirely sure that this pid/command monitoring is something that is inside the design scope of beszel, but it is something I'd like as an alert from something.

WRT to /core usage:  I suppose a simple checkbox on the CPU Usage graph that says `[ ] Show per core usage` and charts that with a line instead of an area graph would be helpful.... then be able to create a /core alert for over threshold over time.

### Category

Metrics

### Affected Metrics

CPU

svenvg93:
@henrygd Think this one can be closed, this is already implemented.

henrygd:
Thanks Sven. I'll go ahead and close this one.

@iconoclasthero We do have charts now that show usage of cores. I know the scope of the requests is wider, but it would be helpful if scope could be narrowed with separate requests instead of one issue with multiple requests. If you want to make new issues that's perfectly fine 👍

[...]

Issue #1122, opened 2025-08-31
[Feature]: CPU multi-core


FasteryXD:
### Component

Hub

### Describe the feature you would like to see

As in the title, it would be nice to have a multicore view of the CPU.

### Motivation / Use Case

Improved personalization and clarity.

### Describe how you would like to see this feature implemented

In the CPU diagram there is a small control on the right to turn the multicore on and off.

### Category

Charts & Visualization

### Affected Metrics

CPU

svenvg93:
This was recently implemented. By clicking the three dots on the CPU chart, you can view charts per CPU core, as well as an overview of the total time spent in each state.

PR #1356, opened 2025-10-30
[Feature] Add detailed CPU metrics (User, System, IOWait, Steal) with per-core monitoring

svenvg93:
## 📃 Description

This PR adds IOWAIT, Steal, User, System for the CPU monitoring as well as monitoring per core.

## 📖 Documentation

### ➕ Added

- closes https://github.com/henrygd/beszel/issues/278
- closes https://github.com/henrygd/beszel/issues/1257
- closes https://github.com/henrygd/beszel/issues/1122

## 📷 Screenshots

Scherm­afbeelding 2025-10-30 om 21 57 20
Scherm­afbeelding 2025-10-30 om 21 57 16

### DIFF
--- a/agent/battery/battery.go
+++ b/agent/battery/battery.go
@@ -10,8 +10,10 @@ import (
 	"github.com/distatus/battery"
 )
 
-var systemHasBattery = false
-var haveCheckedBattery = false
+var (
+	systemHasBattery   = false
+	haveCheckedBattery = false
+)
 
 // HasReadableBattery checks if the system has a battery and returns true if it does.
 func HasReadableBattery() bool {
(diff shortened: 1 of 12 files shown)


henrygd:
Thanks Sven. I love adding steal, etc., but I'm not sure if including all of that data for every core is necessary. It adds up on systems with lots of cores and isn't a metric I look at often.

I'll pick at this for a bit. I think either we should leave out per-core or collect only percent use per core.

svenvg93:
We can easily change it to only collect the total per core. Or just remove the per core completely. It was requested so I added it 😅

Another idea maybe is to make it an environment variable CPU_CORE_DETAILS so the user can enable it if they want it.

henrygd:
Yep, no worries! I'm working on it now and should have it in the next release.

svenvg93:
Thanks for implementing  this in 0.15.3 @henrygd . Would you mind to close the linked issue so the original requester knows it's implemented as well.

PR #1405, opened 2025-11-13
fix: detect and handle immutable filesystems like Fedora Silverblue

### LINKED ISSUE #1402: [Bug]: Agent reporting 100% disk usage on Fedora Silverblue
### Component

Agent

### Description

Due to a change in Fedora Atomic Desktops, the real root filesystem is now mounted on /sysroot, instead of /

From the [Fedora Wiki](https://fedoraproject.org/wiki/Changes/ComposefsAtomicDesktops#User_Experience):

> The main visible change will be that the root filesystem (/) is now small and full (a few MB, 100% used). The real root is mounted in /sysroot and most of the data is stored in /var.

This results in the Hub showing 100% disk usage, when in reality in reality it's not correct.

Image
Image

### Expected Behavior

The Beszel install script should be able to detect whether the system is using an immutable filesystem, like Fedora Silverblue, and change the root path accordingly to /sysroot, thus displaying correct disk usage.

### Steps to Reproduce

Add any machine running an immutable filesystem like Fedora Silverblue to the Beszel Hub.

### Category

Metrics

### Affected Metrics

Storage

### OS / Architecture

Fedora Silverblue 43

### Beszel version

0.15.4

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

CommitToday:
**Overview**
This PR addresses issue #1402 by implementing detection and proper handling of immutable filesystems such as Fedora Silverblue, where the actual root filesystem is located at /sysroot instead of the standard / mountpoint. The changes ensure proper disk usage reporting on these systems.

**Checklist**
- [x] Code changes implemented
- [x] Tests added/updated and passing (verified: ✓ CodeRabbit review passed\n)
- [x] Code compiles/builds successfully (verified)
- [x] Linting/formatting checks pass (verified)
- [x] Documentation reviewed (CONTRIBUTING.md and other .md files read and followed)

**Proof**
This fix was tested by implementing a new `isImmutableSystem()` function that checks for common immutable system indicators (like Fedora Silverblue, CoreOS, Flatcar) and verifies the presence of /sysroot. The disk initialization logic was updated to look for the root filesystem at /sysroot on these systems. The changes compile successfully and maintain existing functionality while adding support for immutable systems.

**Closes #1402**

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -223,9 +223,9 @@ func (a *Agent) getFingerprint() string {
 	return fingerprint
 }
 
-// isImmutableSystem detects if the system is likely an immutable system like Fedora Silverblue
-// where the real root filesystem is mounted at /sysroot instead of /
-func (a *Agent) isImmutableSystem() bool {
+// getRootMountPoint returns the appropriate root mount point for the system
+// For immutable systems like Fedora Silverblue, it returns /sysroot instead of /
+func (a *Agent) getRootMountPoint() string {
 	// Check for typical immutable system indicators
 	// 1. Check if /etc/os-release contains indicators of an immutable system
 	if osReleaseContent, err := os.ReadFile("/etc/os-release"); err == nil {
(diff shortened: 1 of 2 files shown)


henrygd:
Thanks! I'm wondering if this can be simplified. Can we just use a variable that is `/` by default, set it to `/sysroot` if it exists, then replace all the hardcoded `/` instances with the variable?

CommitToday:
@henrygd Yeah it is possible, I just implemented and commited.

henrygd:
Thank you!

PR #1434, opened 2025-11-16
Font ligatures create unwanted artifacts in random ids

m90:
## 📃 Description

When monitoring resources (i.e. containers) that are named using random id schemes, the current font setup will sometimes turn certain sequences into ligatures (e.g. `x` is interpreted as multiplication when in between two numbers). This creates visual noise that is slightly confusing.

Let me know if adding this globally is unwanted and the rule should be applied selectively only (if at all).

## 📖 Documentation

n/a

## 🪵 Changelog

Disabled font-ligatures for all text content.

## 📷 Screenshots

Before:

image

After:

image

### DIFF
--- a/internal/site/src/index.css
+++ b/internal/site/src/index.css
@@ -142,6 +142,7 @@
 
 	body {
 		@apply bg-background text-foreground;
+		font-variant-ligatures: no-contextual;
 	}
 
 	button {


henrygd:
Thank you very much!

Looking at the MDN docs, I'm wondering if `font-variant-ligatures: no-contextual` would work. Should cover the screenshot example and still allow users to use ligatures if they choose.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-variant-ligatures

Also, can we move the rule to the `body` selector instead of using `*`? If `body` doesn't work for whatever reason then no worries.

Cool profile btw :+1:

m90:
> Looking at the MDN docs, I'm wondering if font-variant-ligatures: no-contextual would work.

Yeah, I've been wondering the same, but forgot about it again, so thanks for the reminder. It's updated now.

> Also, can we move the rule to the body selector instead of using *

This is also a good idea and also updated.

henrygd:
Perfect, thank you!

PR #1442, opened 2025-11-18
fix: Fix the issue where the Add System button is visible to read-only users.

zjkal:
When the login user's permission is "ReadOnly," hide the "AddSystemButton" button in the upper right corner of the page.


henrygd:
Thanks!

PR #1493, opened 2025-12-03
feat: add network connection stats

svenvg93:
## 📃 Description

This PR added network connections stats for IPv4 and IPv6. 
Total amount is shown on the main page, with a sheet to split per type.
Scherm­afbeelding 2025-12-03 om 19 26 55
IPv6 chart is hidden when there is no session data.
Scherm­afbeelding 2025-12-03 om 19 26 47


svenvg93:
Converting to draft till #1513 is merged

PR #1509, opened 2025-12-08
feat: new config and temprature selector

marmar76:


marmar76:
image

add capability to select which sensor to show on the home
image
image

With catch
1. it need to add migration to the system table, i dont know how to do that
2. it only work when the page is being opened, i know i need to deep dive to the agent, go is not my expertise
3. my frontend sucks

henrygd:
I appreciate your work, but we already have a way to do this with the `PRIMARY_SENSOR` environment variable. You would just set `PRIMARY_SENSOR=coretemp_core_0` on the agent.

Alternatively, you could exclude the nct sensors with `SENSORS=-nct*`. Let me know if this doesn't work for you.

marmar76:
Alright, sorry—I couldn’t find the documentation at https://beszel.dev/guide/environment-variables#primary-sensor. Maybe my example could be added to the guide so future users can benefit from it.
from this
image
to this
image
by adding `SENSORS=-nct*` on agent env

PR #1511, opened 2025-12-09
fix: When there is no client, LoaderCircle will always transfer

Zero2A11:
## 📃 Description

Bug：When there are no Systems (or the PB items list is empty), the All Containers page would stay in a permanent loading state (the LoaderCircle spinner never goes away).

image
image

## 📖 Reason

**Root cause**: the code only called setData(...) when items.length > 0 (using items.length && setData(...)), so when items was empty the data state remained undefined and the UI treated that as “still loading”.

### 🔧 Fixed

When pb.collection(...).getList(...) returns an empty items, call setData([]) so data reflects “loaded but no items”.

## 📷 Screenshots

After fixed：

image


henrygd:
Thanks :+1:

PR #1516, opened 2025-12-10
fix: handle invalid memory stats in cgroup2 container environments

dudgns0507:
## 📃 Description

Fix invalid memory stats in container environments with cgroup2 where memory.max is set to "max" (unlimited).
In certain environments (e.g., Oracle Cloud VMs), gopsutil's mem.VirtualMemory() returns invalid values, causing memory to be displayed as ~16777216 TB instead of the actual value (~5.5GB).
This PR adds validation for memory values and falls back to parsing /proc/meminfo directly when invalid values are detected.

### ➕ Added

- Fallback memory parsing from /proc/meminfo when gopsutil returns invalid values
- Debug logging when fallback is triggered

### 🔧 Fixed

- Memory displayed as ~16777216 TB in cgroup2 container environments with unlimited memory

## 📷 Screenshots

Before
MemPct: 3.1169216869205e+11
MemUsed: 1.717986917667e+10
Memory displayed as 16777216 TB
image

After
Memory displayed correctly as 5.51 GB
image

### DIFF
--- a/agent/system.go
+++ b/agent/system.go
@@ -115,6 +115,35 @@ func (a *Agent) getSystemStats(cacheTimeMs uint16) system.Stats {
 
 	// memory
 	if v, err := mem.VirtualMemory(); err == nil {
+		// Validate memory values - in some container environments (e.g., Oracle Cloud with cgroup2),
+		// gopsutil may return invalid values when memory.max is set to "max" (unlimited).
+		if v.Used > v.Total || v.UsedPercent > 100 {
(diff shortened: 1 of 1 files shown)


henrygd:
Thank you!

I have no objections to this, but maybe it would make more sense to add the fix upstream in gopsutil? That would resolve the problem for everyone else using the library as well.

https://github.com/shirou/gopsutil/blob/master/mem/mem_linux.go

Let me know what you think :+1:

dudgns0507:
That sounds like a good idea. I’ll try applying the fix upstream in gopsutil.

PR #1530, opened 2025-12-12
Upgrade React to address CVE-2025-55182

daviddavis:
## 📃 Description

This addresses [a critical vulnerability in react](https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components).


henrygd:
Thanks! Technically I don't think we're vulnerable because we don't run any server-side React / JS, but doesn't hurt to upgrade anyway.

PR #1587, opened 2025-12-30
nvml: add rtd3 memory workaround, fix slog imports

crimist:
Only call MemoryInfo NVML APIs if the GPU is actually being used by checking `> 0% usage`. Fixes RTD3 since the MemoryInfo NVML APIs seem to refresh the 21 second suspend window.

Also fixed slog imports since the difference packages have different global log instances therefore breaking logging.

### DIFF
--- a/agent/gpu.go
+++ b/agent/gpu.go
@@ -15,7 +15,7 @@ import (
 
 	"github.com/henrygd/beszel/internal/entities/system"
 
-	"golang.org/x/exp/slog"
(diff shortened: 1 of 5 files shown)


henrygd:
Thank you! Confirmed this keeps me in D3cold unless using the GPU. I think this is is a good solution for now.

Should have a new release out in the next couple of days. Appreciate your contributions :+1:

PR #1599, opened 2026-01-06
site: only hide GPU power graph if entire timescale is 0W

crimist:
### Description

Current logic hides the GPU power usage graph when a sample reports 0W. This means that useful history is hidden when a GPU enters RTD3 D3cold suspend and drops to 0W. This change _should_ leave the graph up unless the entire history of the current timescale on it is 0W.

Related to RTD3 NVML changes (#1522)

### Testing

I am crashing trying to npm run / build this on main and this branch so if you could test that would be much appreciated. 

```
$ npm --version
11.7.0
$ node --version
v25.2.1
$ npm run dev --host 0.0.0.0
npm warn "0.0.0.0" is being parsed as a normal command line argument.
npm warn Unknown cli config "--host". This will stop working in the next major version of npm.

> beszel@0.18.0-beta.2 dev
> vite --host 0.0.0.0

[1]    481847 bus error (core dumped)  npm run dev --host 0.0.0.0
```

### Screencast of old behavior

https://github.com/user-attachments/assets/dca17edc-3bb4-41b2-86ca-838ac3e9bb33

Issue #1522, opened 2025-12-11
[Feature]: Allow NVIDIA GPUs with RTD3 enabled to enter D3cold


crimist:
### Component

Agent

### Describe the feature you would like to see

Allow NVIDIA GPUs to enter power saving mode.

### Motivation / Use Case

Improves laptop battery life.

### Describe how you would like to see this feature implemented

Will follow up with quotes from previous discussion.

### Category

Performance

### Affected Metrics

GPU

crimist:
Discussion from https://github.com/henrygd/beszel-docs/pull/35

https://github.com/henrygd/beszel-docs/pull/35#issue-3714045772
> Spent many hours trying to figure out why my new laptop GPU refused to enter D3cold. Hopefully this will save others.
> 
> For additional context the RTX 5060 being stuck in D0 causes it to consume 5-7W of additional power on my Lenovo 16IAH10. This leads to a massive reduction in idle battery life:
> 
> (84 watt hours) / (16 watts) = 5.25 hours (84 watt hours) / (10 watts) = 8.4 hours


https://github.com/henrygd/beszel-docs/pull/35#issuecomment-3639329331
> Thanks, I hadn't considered this before. I wonder if we could check the content of `/sys/class/drm/cardX/device/power_state` and skip collection if the card is sleeping?
> 
> ```
> watch -n 0.5 cat /sys/class/drm/card1/device/power_state
> ```
> 
> I don't have an Nvidia card on hand at the moment. Would you be able to verify if that populates with a set value when the card is in D3cold vs not in it?
> 
> Edit: Also important is how quickly it goes back to D3cold, assuming it does change when running `nvidia-smi`.


https://github.com/henrygd/beszel-docs/pull/35#issuecomment-3639377461
> Yeah, it's very easy to determine the RTD3 state of a GPU and check whether RTD3 is enabled for NVIDIA GPUs using /sys/ files.
> 
> The problem is that running nvidia-smi at all (even just as a single call) causes the GPU to enter D0 for a period of time. I want to say somewhere in the realm of 10-120 seconds, I can run some experiments when I get home.
> 
[...]

PR #1624, opened 2026-01-12
site: only hide GPU engine graph if entire usage is 0%

crimist:
# Explanation

Same issue as #1599. Logic was only checking last sample instead of the entity of the graphs range leading to "flickering".

# Screencast of pre-patch behavior

https://github.com/user-attachments/assets/87ca378c-cc6f-4589-94fe-fe08c263b13e


crimist:
This patch isn't necessary since this UI situation is due to a bug in the collector (#1522). Regardless, it may be more sound logic for handling when to hide the graph.

henrygd:
Makes sense, thanks.

PR #1636, opened 2026-01-13
feat: Added tooltips for navbar buttons to clear meaning of each one

### LINKED ISSUE #1623: [Feature]: Add tooltips to any clickable item that doesn't have text next to it
### Component

Hub

Add tooltips to any clickable item that doesn't have text next to it.

The top buttons for example aren't 100% clear on what they do, containers and smart info specifically, but tooltips should be on all buttons where there is no explaining text. :)

### Motivation / Use Case

Accessibility/discoverability

Tooltips :D

### Screenshots

_No response_

### Category

Metrics

### Affected Metrics

CPU

Fahleen1:
## 📃 Description
Add tooltips to any clickable item that doesn't have text next to it.

Closes #1623

### ➕ Added
Added tooltips for each navbar items to clear each purpose.

## 📷 Screenshots
Screenshot 2026-01-13 at 2 38 09 PM

### DIFF
--- a/internal/site/src/components/lang-toggle.tsx
+++ b/internal/site/src/components/lang-toggle.tsx
@@ -10,32 +10,32 @@ import { Tooltip, TooltipContent, TooltipTrigger } from "./ui/tooltip"
 export function LangToggle() {
 	const { i18n } = useLingui()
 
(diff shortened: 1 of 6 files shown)


henrygd (approved):
Thank you! 👍

Issue #1622, opened 2026-01-12
[Feature]: Show hostname as a tooltip if elided


Justinzobel:
### Component

Hub

### Describe the feature you would like to see

Currently, a hostname is elided (... at the end if it is too long) but we cannot see the hostname easily).

Add the full hostname as a tooltip on that item.

### Motivation / Use Case

Seeing a full hostname

### Describe how you would like to see this feature implemented

Add the full hostname as a tooltip on that item.

### Category

UI / UX

### Affected Metrics

Other

Fahleen1:
@Justinzobel Can you please tell where is this issue exist I mean attach the screenshot too if possible

Justinzobel:
<img width="651" height="290" alt="Image" src="https://github.com/user-attachments/assets/f88de687-eada-4d34-84e3-f96403c45bbe" />

svenvg93:
This was implemented in #1640. It will be in the next release.

PR #1640, opened 2026-01-14
feat: add tooltip to system name in systems table

Fahleen1:
## 📃 Description

#1622 Add tooltip to system name in systems table for better accessiility

### ➕ Added

- Added tooltip on system hostname for better user experience

## 📷 Screenshots
Screenshot 2026-01-14 at 11 00 42 AM


henrygd:
Good idea, thanks!

This will be in the next release. I did change it to use a title attribute instead of a tooltip, and only shown if the name is truncated.

Issue #1653, opened 2026-01-19
[Bug] SMART_DEVICES doesn't support devices where the type name has a comma


jules2689:
This line https://github.com/henrygd/beszel/blob/031abbfcb3b4f00660fca9080cc8bfd870a938d4/agent/smart.go#L205 simply splits on a comma, but a comma is load bearing on a number of device names for requried arguments (like `jmb39x-q,0`).

The easiest fix is probably to use a separate non-load bearing character as the split (like a `|` character), but I realize that might be hard given that it wouldn't be backwards compatible.

Perhaps an easier approach may be a `SMART_DEVICES_CHAR` to override the split char to be something like a `|`, but defaulting to a comma so it's backwards compatible?

Ultimately I'd like to be able to specify the following:
```
cat /etc/systemd/system/beszel-agent.service
[Unit]
Description=Beszel Agent Service
After=network.target

[Service]
StandardOutput=file:/var/log/beszel.log
StandardError=file:/var/log/beszel.log
Environment="PORT=45877"
Environment="GPU=true"
Environment="SMART_DEVICES=/dev/sdb:jmb39x-q,0" # This line is impacted
Environment="KEY=ssh..."
Environment="EXTRA_FILESYSTEMS=..."
CapabilityBoundingSet=CAP_SYS_RAWIO CAP_SYS_ADMIN
SupplementaryGroups=disk
AmbientCapabilities=CAP_SYS_RAWIO CAP_SYS_ADMIN
ExecStart=/opt/beszel-agent/beszel-agent
User=beszel
Restart=always
RestartSec=5

[Install]
WantedBy=multi-user.target
```

jules2689:
I have this PR running locally https://github.com/henrygd/beszel/pull/1655 and it seems to work well for me!

henrygd:
Thanks so much for reporting and for your PR!

Your solution makes sense, and I'll get back to you soon when I have a bit more time to review. Will try to make sure this gets into the next release.

henrygd:
The 0.18.3 release is out with the `SMART_DEVICES_SEPARATOR` env var. Let me know if it's not working properly for anyone.

PR #1655, opened 2026-01-19
Add SMART_DEVICES_SEPARATOR + allow drives with the same name to be added with different types (e.g. raid controllers)

jules2689:
## 📃 Description

- Fixes an issue where types with commas cannot be used in `SMART_DEVICES`
- Fixes an issue where multiple drives with the same name (e.g. in a raid controller) could not be added by making a composite key between name + type

## 📖 Documentation

https://github.com/henrygd/beszel-docs/pull/39

## 🪵 Changelog

- `SMART_DEVICES_SEPARATOR` added to override `,` separator in `SMART_DEVICES`

### ➕ Added

- Allows for overriding the `,` separator for SMART_DEVICES so you can use types with commas in it like `jmb39x-q,1`. Uses `SMART_DEVICES_SEPARATOR`. This was chosen to remain backwards compatible

### ✏️ Changed

- Uses a composite key of type + name so we can re-use names for SMART devices
  - Allows for some better raid controller support by allowing all drives to be added

image

### 🔧 Fixed

Closes https://github.com/henrygd/beszel/issues/1653


jules2689:
There's a few failed tests I'd like your input on, if this PR is otherwise a good path:

The failing tests relate to copying metadata between overridden and existing/scanned. Because it relied on name before, it would copy over other metadata. This seemed like a salient approach previously, but I'm not so sure that it is now with this PR considering a name+type composite key. Is this behaviour we should maintain, or is this something that may not make sense now that the assumptions have changed?

henrygd:
Thanks so much! I added a small update which should preserve the new functionality you added, as well as the previous behavior for devices with unique names. The tests pass and this has been merged into main.

Let me know if my changes broke anything on your end :+1:

PR #1661, opened 2026-01-20
chore: update workflows and templates

svenvg93:
## 📃 Description

This PR updates the templates to be simpler and provides clearer guidance on how to fill them in.
The templates now include additional checkboxes, which should encourage users to search for existing issues or requests before submitting a new one.
Based on the templates I made for another project; https://github.com/alexjustesen/speedtest-tracker/issues/new/choose

The workflow that automatically added labels based on the template has been removed. While the idea was nice, it added unnecessary clutter to the issue overview.

Also added a Code Owners to make sure all the changes are reviewed by you. 

### ✏️ Changed

- Issue template to be simpler with more guidelines on how to fill it in
- Feature reqeust template to be simpler with more guidelines on how to fill it in

### 🗑️ Removed

- Workflow to add labels based on the template.

### DIFF
--- a/.github/DISCUSSION_TEMPLATE/support.yml
+++ b/.github/DISCUSSION_TEMPLATE/support.yml
@@ -36,7 +36,7 @@ body:
   - type: textarea
     id: description
     attributes:
-      label: What did you do?
+      label: Problem Description
       description: |
         How to write a good bug report?
 
@@ -57,13 +57,13 @@ body:
     validations:
       required: true
   
-  - type: input
-    id: version
-    attributes:
-      label: Beszel version
-      placeholder: 0.9.1
-    validations:
-      required: true
+#  - type: input
+#    id: version
+#    attributes:
+#      label: Beszel version
+#      placeholder: 0.9.1
+#    validations:
+#      required: true
 
   - type: dropdown
     id: install-method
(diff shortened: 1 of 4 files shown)


henrygd (commented):
Thanks Sven! I agree with your reasoning, and everything looks great besides a couple things with the support discussion template.

1. Let's remove the "Beszel version" input field. I think the checkbox you added is better and we don't need both.

2. The label "What did you do?", at least to me, can read a bit like we're blaming the user for something. I know that's not the intent, but maybe something like "What is your question or problem?" could be better? Let me know what you think.

svenvg93:
> 2. The label "What did you do?", at least to me, can read a bit like we're blaming the user for something. I know that's not the intent, but maybe something like "What is your question or problem?" could be better? Let me know what you think.

Agreed , I think “What did you do?” was intended for the “Steps to Reproduce.” I’ve changed it to “Problem Description” in the bug report and support discussion templates.

> 1. Let's remove the "Beszel version" input field. I think the checkbox you added is better and we don't need both.

Unfortunately, I’ve seen many cases where people simply check the required boxes without actually reading or following up on them. For now, I’ve commented out the “Beszel version.” Let’s see how this goes. we can always bring it back if needed.

Edit: Also updated the workflow for stale issue. Issues with the `awaiting-requester` label will be marked as stale after 14 days of inactivity, then closed 7 days later if still inactive. However, issues with the `enhancement` label are exempt from this process. This allows you to use the label to request more information from the requester without automatically closing them, while still being able to track and monitor them easily. Unless you want them to be closed as well if the requester does not reply within 3 weeks.

henrygd:
Perfect, thank you!

PR #1683, opened 2026-01-25
Bug: Agent Healthcheck failed for non root docker

### LINKED ISSUE #1455: [Bug]: beszel-agent /agent health: stat /tmp/beszel_health: no such file or directory
### Component

Agent

### Description

The health command of the beszel-agent reports always unhealthy since the version 0.16.1.

compose.yaml:
`healthcheck:
      test: ['CMD', '/agent', 'health']`

Checking the health state via docker inspect:

`docker inspect --format "{{json .State.Health }}" beszel-agent | jq`

reports:

`{
  "Status": "unhealthy",
  "FailingStreak": 3,
  "Log": [
    {
      "Start": "2025-11-20T21:51:15.994206409+01:00",
      "End": "2025-11-20T21:51:16.099188829+01:00",
      "ExitCode": 1,
      "Output": "2025/11/20 20:51:16 stat /tmp/beszel_health: no such file or directory\n"
    }`

Manually executing the command also gives the same error.

`docker compose exec beszel-agent /agent health`
`2025/11/20 20:52:32 stat /tmp/beszel_health: no such file or directory`

### Expected Behavior

healthy status for the beszel-agent container

### Steps to Reproduce

Use the healthcheck example for the agent on https://beszel.dev/guide/healthchecks#agent-example

### Category

Metrics

### Affected Metrics

Containers

### OS / Architecture

linux/amd64

### Beszel version

0.16.1

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

The agent health check was failing with stat `/tmp/beszel_health: no such file or directory` when running as a non-root user in Docker. This was caused by Docker's `COPY` command not preserving directory permissions - the /tmp directory in scratch images had 755 permissions instead of 1777, preventing non-root users from creating files.
Changed the health file location from `/tmp/beszel_health` to the agent's data directory.

Fixes #1455

### ✏️ Changed

- Defaults to `/tmp/beszel_health` as a fallback
- Uses `SetDataDir()` to set the health file path when the agent finds a valid data directory

### DIFF
--- a/agent/update.go
+++ b/agent/update.go
@@ -1,12 +1,10 @@
 package agent
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"runtime"
-	"strings"
 
 	"github.com/henrygd/beszel/internal/ghupdate"
 )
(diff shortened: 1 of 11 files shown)


henrygd:
I looked a bit further into it because I wanted to avoid mixing this into `agent.go` if possible, and figured out that we can just write to `/dev/shm` instead of `/tmp`. I was able to fix by doing this: https://github.com/henrygd/beszel/commit/afc19ebd3b857aec7988562402c5c3e7114d04ea.

I'll close the PR, but I appreciate your work on this anyway!

PR #1736, opened 2026-02-09
SMART: add eMMC health via sysfs

VACInc:
Adds Linux eMMC wear/EOL health reporting via sysfs (/sys/class/block/mmcblk*/device: pre_eol_info + life_time / device_life_time_est_typ_a/b), surfaced in the existing SMART device list as Type=EMMC.

Includes mocked sysfs tests and a small UI tweak to treat 0C temp as N/A.

image

### DIFF
--- a/agent/smart.go
+++ b/agent/smart.go
@@ -28,8 +28,7 @@ type SmartManager struct {
 	SmartDevices    []*DeviceInfo
 	refreshMutex    sync.Mutex
 	lastScanTime    time.Time
-	binPath         string
-	hasSmartctl     bool
+	smartctlPath    string
 	excludedDevices map[string]struct{}
 }
 
@@ -177,11 +176,11 @@ func (sm *SmartManager) ScanDevices(force bool) error {
 		hasValidScan   bool
 	)
 
-	if sm.hasSmartctl {
+	if sm.smartctlPath != "" {
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer cancel()
 
-		cmd := exec.CommandContext(ctx, sm.binPath, "--scan", "-j")
+		cmd := exec.CommandContext(ctx, sm.smartctlPath, "--scan", "-j")
 		output, err := cmd.Output()
 		if err != nil {
 			scanErr = err
@@ -459,7 +458,7 @@ func (sm *SmartManager) CollectSmart(deviceInfo *DeviceInfo) error {
 		}
 	}
 
-	if !sm.hasSmartctl {
+	if sm.smartctlPath == "" {
 		return errNoValidSmartData
 	}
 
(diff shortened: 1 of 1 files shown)


VACInc:
@henrygd Works on my end; made one additional minor change due to the merge conflict.  Also; there are three potential outcomes of eMMC health checking:
  - 0x01 => PASSED
  - 0x02 => WARNING
  - 0x03 => FAILED

As of right now only failed (beyond life expectancy) triggers the alert.  I would think we want warning too but was thinking that would result in extra custom plumbing for eMMC only in internal/alerts/alerts_smart.go and that doesn't even make me happy as much as I want it.

Any thoughts?

henrygd:
Good point. I updated `alerts_smart.go` to handle the warning state. Let me know if that looks good to you: https://github.com/henrygd/beszel/pull/1736/commits/d45682bfdc7a6244b23eee389d9ef151f6aaeeef

VACInc:
Awesome!  That's great!  One last thought.  Do we want to trigger warnings if a disk goes from UNKNOWN to WARNING or FAILED as well?

Something like this?

```
  func shouldSendSmartDeviceAlert(oldState, newState string) bool {
      oldSeverity := smartStateSeverity(oldState)
      newSeverity := smartStateSeverity(newState)

  -   // Ignore unknown states and recoveries; only alert on worsening transitions
  -   // from known-good/degraded states into WARNING/FAILED.
  -   return oldSeverity >= 1 && newSeverity > oldSeverity
  +   // Also alert when UNKNOWN escalates to WARNING/FAILED.
  +   if oldSeverity == 0 && (newSeverity == 2 || newSeverity == 3) {
  +       return true
  +   }
  +
  +   // Existing behavior: alert on worsening known transitions.
  +   return oldSeverity >= 1 && newSeverity > oldSeverity
  }
```

If yes I can make the change (and update tests to reflect both the change you just made and this one)

henrygd:
My reasoning for that is to reduce potential noise. If the device is reporting UNKNOWN, it may be a temporary issue or quirk. But we don't know what it was reporting before it became UNKNOWN.

If it goes UNKNOWN to FAILED, it was likely in FAILED state previously before going to UNKNOWN. So this could result in repeated FAILED notifications for the rest of time.

If the device goes from PASSED to WARNING/FAILED or WARNING to FAILED, we know for sure that it's a new escalation, and less likely to flap back and forth like UNKNOWN to anything else.

To me it makes more sense to err on the side of caution, but I'm open to considering other takes.

VACInc:
I kind of figured that was what the thought pattern behind it was and I agree on that.

The only other thought (which would certainly be scope creep at this point) would be a `last_known_state` or something of the like.  But then that'd have to be stored in the DB which again makes it an even bigger change.

henrygd:
Possibly, but let's hold off on that for now. I have limited time at the moment and want to get this in asap.

If a device shows UNKNOWN, it's generally a configuration issue anyway that needs to be fixed before any other state can be reached (at least with `smartctl`).

I added another tiny refactor that shouldn't change anything, and I think it's ready to merge at this point. Does that sound good to you or is there anything else you want to tweak first?

VACInc:
Understood and agreed; LGTM.  Thanks @henrygd !

henrygd:
Perfect, thanks again! Really appreciate your work :+1:

Issue #1746, opened 2026-02-14
[Feature]: Mac GPU stats


raccettura:
### Welcome!

- [x] I have searched open and closed feature requests to make sure this or similar feature request does not already exist.
- [x] This is a feature request, not a bug report or support question.

### Component

Agent

### Description

Would be nice to get GPU stats for MacOS M series GPU's.

It seems it would be possible to use the `powermetrics` command and scrape the output for GPU data.  For example:

```
$ sudo powermetrics --samplers gpu_power -i 500 -n 1
Machine model: Mac14,10
OS version: 25D125
Boot arguments: 
Boot time: Thu Feb 12 16:20:49 2026



*** Sampled system activity (Sat Feb 14 00:42:06 2026 -0500) (503.05ms elapsed) ***


**** GPU usage ****

GPU HW active frequency: 444 MHz
GPU HW active residency:   0.97% (444 MHz: .97% 612 MHz:   0% 808 MHz:   0% 968 MHz:   0% 1110 MHz:   0% 1236 MHz:   0% 1338 MHz:   0% 1398 MHz:   0%)
GPU SW requested state: (P1 : 100% P2 :   0% P3 :   0% P4 :   0% P5 :   0% P6 :   0% P7 :   0% P8 :   0%)
GPU idle residency:  99.03%
GPU Power: 4 mW
```

### Motivation / Use Case

Want stats

raccettura:
<img width="1133" height="689" alt="Image" src="https://github.com/user-attachments/assets/fc0d8c07-7296-4135-bef2-be88021bd036" />

lnlyssg:
I've installed macmon (and tested it works) and have added `GPU_COLLECTOR=macmon` to the Beszel config and then restarted the service but in the logs I'm seeing `2026/02/21 10:13:24 DEBUG GPU err="no configured GPU collectors are available"`

wmorgue:
Doesn't work for me. What I've missed?

```bash
❯ cat .config/beszel/beszel-agent.env
KEY="ssh-ed25519 ..."
LISTEN=45876
GPU_COLLECTOR=macmon
LOG_LEVEL=debug
SENSORS=""

❯ which macmon
/opt/homebrew/bin/macmon

❯ macmon pipe -i 2500

[...]

PR #1747, opened 2026-02-14
Mac GPU Stats

raccettura:
## 📃 Description

This is basically vibe coded, but it works, so figured I'd setup a PR for anyone who wants to pick this up.  I'm not much of a go developer.

I'm not sure how permissions are best dealt with re bezel on Mac OS, I tested via using sudo.

## 📖 Documentation

### ➕ Added

- issue #1746 1746

## 📷 Screenshots

If this PR has any UI/UX changes it's strongly suggested you add screenshots here.

image

### DIFF
--- a/agent/gpu.go
+++ b/agent/gpu.go
@@ -20,12 +20,13 @@ import (
 
 const (
 	// Commands
-	nvidiaSmiCmd  string = "nvidia-smi"
-	rocmSmiCmd    string = "rocm-smi"
-	tegraStatsCmd string = "tegrastats"
-	nvtopCmd         string = "nvtop"
-	powermetricsCmd  string = "powermetrics"
-	noGPUFoundMsg    string = "no GPU found - see https://beszel.dev/guide/gpu"
+	nvidiaSmiCmd    string = "nvidia-smi"
+	rocmSmiCmd      string = "rocm-smi"
+	tegraStatsCmd   string = "tegrastats"
+	nvtopCmd        string = "nvtop"
+	powermetricsCmd string = "powermetrics"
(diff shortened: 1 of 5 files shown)


henrygd:
Thanks for your work. Unfortunately I don't have a Mac, so I can't test or help with this. I suppose we could mark it as experimental, make it opt-in, and ask others to test.

The `sudo` requirement is a unfortunate, though. Generally I want to avoid that unless there's no other choice.

I found a tool called [macmon](https://github.com/vladkens/macmon) which seems to provide the same data without `sudo`, and should be very efficient. Maybe we could use this instead? The code would be very similar, but we'd need to update the parser to work with `macmon`'s JSON output. I think this command would work:

```bash
macmon pipe -i 2500
```

Let me know what you think.

raccettura:
I think that's a good improvement.  

Installed with `brew install macmon`.  Updated the PR to prioritize macmon then fallback to the built in powermetrics.

This seems to work for me.  But again, i'm not a go developer so my sanity checking of this code is limited to "this doesn't look completely backwards".

henrygd:
Merged, thanks very much!

Since I can't easily test myself, I changed to be opt-in only for now. So you need to set either `GPU_COLLECTOR=macmon` or `GPU_COLLECTOR=powermetrics`.

I should have a new release out this week, and I'll add a note in the docs linking to #1746 and asking for feedback. We'll enable it by default if others can confirm it's working properly. :+1:

raccettura:
Sounds good.  Thanks. Next up is getting disk IO working.  Seems Mac's use `iostat`.

PR #1750, opened 2026-02-16
Add Linux mdraid health monitoring

VACInc:
## Summary
- add Linux mdraid array health via sysfs (no full-device scan)
- keep existing eMMC sysfs health support
- allow Linux agents with mdraid/eMMC to run without smartctl

## Testing
- go test ./agent


henrygd:
Merged, thanks again for your work!

Nice thing about this feature is that it works out of the box in Docker since it's sysfs-based :+1:

VACInc:
Absolutely; thank you for starting and keeping the project going!  Looking forward to contributing more

PR #1754, opened 2026-02-17
fix(agent): Retry Docker check on non-200 HTTP response

ElioDiNino:
## 📃 Description

The current retry behaviour only catches some errors, including inaccessible hosts, but not others, such as failed authentication (401) or service unavailability (503). This largely applies when using a socket proxy, and this change mitigates some erroneous behaviour and increases reliability.

A specific issue I am solving for my use case is https://github.com/wollomatic/socket-proxy/issues/83, where the socket proxy reads Docker labels to assign socket permissions. Since the Beszel agent boots up extremely quickly, the socket proxy blocks the version check request with a 403 before receiving the Docker labels and correctly updating permissions for future requests.

Note that AI was used to help generate the tests, but the cases are my own, and I have also tested locally using the make commands. Additionally, since the retry logic is being tested in three of the test cases, 15 seconds of delay is added to the total test run time. If this is unacceptable, please let me know how you would like the delay to be shortened/skipped just for the tests (a simple change would be an additional dockerManager value, but that solution is not my favourite).

Builds on #1250

### 🔧 Fixed

- Agent now retries Docker connection initialization when a non-200 HTTP response is received

### DIFF
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -594,18 +594,22 @@ func (dm *dockerManager) checkDockerVersion() {
 	const versionMaxTries = 2
 	for i := 1; i <= versionMaxTries; i++ {
 		resp, err = dm.client.Get("http://localhost/version")
-		if err == nil {
+		if err == nil && resp.StatusCode == http.StatusOK {
 			break
 		}
 		if resp != nil {
 			resp.Body.Close()
 		}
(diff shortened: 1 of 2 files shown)


henrygd:
Thanks very much!

PR #1764, opened 2026-02-22
Fix volume path in Docker run command

guidNull:
## 📃 Description

Fix invalid relative path `./beszel_agent_data` in the Docker run command generated by the UI. Docker does not accept relative paths as bind mount sources, causing the container to fail to start.

```bash
/opt/beszel-agent# docker run -d --name beszel-agent --network host --restart unless-stopped -v /var/run/docker.sock:/var/run/docker.sock:ro -v ./beszel_agent_data:/var/lib/beszel-agent -e KEY="KEY" -e LISTEN=45876 -e TOKEN="TOKEN" -e HUB_URL="https://localhost:8090" henrygd/beszel-agent
docker: Error response from daemon: create ./beszel_agent_data: "./beszel_agent_data" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
```

### 🔧 Fixed

- Replace `./beszel_agent_data` with a named volume `beszel_agent_data` in the generated `docker run` command to fix Docker error: *includes invalid characters for a local volume name*

### DIFF
--- a/internal/site/src/components/install-dropdowns.tsx
+++ b/internal/site/src/components/install-dropdowns.tsx
@@ -43,7 +43,7 @@ export function copyDockerCompose(port = "45876", publicKey: string, token: stri
 
 export function copyDockerRun(port = "45876", publicKey: string, token: string) {
 	copyToClipboard(
-		`docker run -d --name beszel-agent --network host --restart unless-stopped -v /var/run/docker.sock:/var/run/docker.sock:ro -v ./beszel_agent_data:/var/lib/beszel-agent -e KEY="${publicKey}" -e LISTEN=${port} -e TOKEN="${token}" -e HUB_URL="${getHubURL()}" henrygd/beszel-agent`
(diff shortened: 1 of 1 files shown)


henrygd:
Are you on a really old version of Docker by any chance? Looks like support for relative source paths in `docker run` was added in [23.0.0](https://docs.docker.com/engine/release-notes/23.0/#2300), which has been out for a few years now.

```bash
docker version
```

guidNull:
I’m running Debian 12 (Bookworm) where the packaged [docker.io version is 20.10.24+dfsg1 from the Debian repositories.](https://packages.debian.org/bookworm/docker.io) Debian Stable does not update packages to newer feature releases unless there are critical security issues, which is why Docker there sticks to 20.10.x. As a result, relative bind mounts (-v ./…) behave differently and this leads to the error described.

Upgrading to a newer Docker CE (≥ 23.0.0) from the official Docker repository does fix the relative path behavior, but the vanilla docker.io in Debian doesn’t get updated to that version.

Because many Debian users will still have 20.10.x, this fix makes the instructions work on older Docker versions too.

henrygd:
Merged, thanks. I agree with your reasoning and it shouldn't hurt anything.

PR #1766, opened 2026-02-22
feat: add dismiss alert function

### LINKED ISSUE #1483: [Feature ]: Add the option to dismiss alerts when the host goes offline
### Component

Hub

### Description

An alert is correctly triggered when the monitored PC reaches a high memory usage state. However, if I shut down the PC while the alert is active, the 'resolved' notification is not sent until the machine comes back online. It is only after the host is reachable again and the integration can verify that memory usage is normal that the alert is marked as resolved.

This behaviour seems incorrect. In most monitoring systems (e.g. Zabbix, Prometheus and Nagios), if a host becomes unavailable, the alert is paused or set to an 'unknown' state. Resolution is not delayed until the host returns because this creates misleading or late notifications.

### Expected Behavior

When the monitored host goes offline, the alert should not remain open, waiting for the host to return to a normal state. Instead, the alert should switch to a 'paused' or 'unknown' state. At the very least, it should avoid sending a delayed 'resolved' notification once the host comes back online.

**Actual behavior:**
The alert is only resolved when the host becomes reachable again and the integration detects normal memory usage.

### Steps to Reproduce

1. Trigger a high memory usage alert.
2. Shut down or disconnect the monitored PC.
3. Turn the PC back on later.
4. Observe that the “resolved” notification appears only after the host becomes reachable again.

### Category

Notifications & Alerts

### Affected Metrics

Memory

### OS / Architecture

Windows 10 (agent), Debian13 / Docker (server)

### Beszel version

0.16

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

sternma:
## 📃 Description
Closes #1483 based on the disposition in the comments of the issue.

Adds the ability to dismiss alerts.

Note that this does not include any "snooze" functionality; this is particularly visible when developing with alert time set to 1 minute, as if you dismiss at 57 seconds into the profiling period, the alert will return again 3 seconds later.  A later solution could be to add a "snooze" period, either automatic or configured, or somehow reset the timer for the profiling period for that specific alert type.

### ➕ Added

- feat: add dismiss alert function

## 📷 Screenshots

image
image
image

### DIFF
--- a/internal/alerts/alerts_status.go
+++ b/internal/alerts/alerts_status.go
@@ -134,6 +134,10 @@ func (am *AlertManager) handleSystemUp(systemName string, alertRecords []*core.R
 			}
 			continue
 		}
+		// Skip alerts that were already resolved/dismissed while the system was down.
+		if !alertRecord.GetBool("triggered") {
+			continue
+		}
 		// No alert scheduled for this record, send "up" alert
 		if err := am.sendStatusAlert("up", systemName, alertRecord); err != nil {
 			am.hub.Logger().Error("Failed to send alert", "err", err)
--- a/internal/alerts/alerts_test.go
+++ b/internal/alerts/alerts_test.go
@@ -450,6 +450,60 @@ func TestStatusAlerts(t *testing.T) {
 	})
 }
 
+func TestStatusAlertDismissedWhileDownDoesNotSendUpNotification(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		hub, user := beszelTests.GetHubWithUser(t)
+		defer hub.Cleanup()
+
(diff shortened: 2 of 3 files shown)


henrygd:
Thanks for your work on this.

Like you said, the downside to this approach is the alert will likely reappear soon after, since the condition that created the alert is still active. It can also create misleading end times in the alert history table.

Considering alternatives, maybe we could add a `dismissed` column in the `alerts` collection, then set that to `true` if the alert is dismissed. Then add some logic to not display in UI / not send notification if `dismissed` is true?

Let me know what you think.

If you want to edit the collection, just toggle the switch in PocketBase at `/_/#/settings` and edit them in the PocketBase UI -- any changes should automatically create a migration file.

PR #1777, opened 2026-02-27
Container aliasing

### LINKED ISSUE #1385: [Feature]: Container identifier alternatives
### Component

Hub

An option to either set a custom name/alias for a container or set which key should be used by the Hub to identify the container. These would replace the container name used in the graphs and table.

### Motivation / Use Case

As containers accumulate it gets harder to identify them, and we don't always get to name them as we please. this leads to situation where a graph can have tens of containers and having no idea which is which.

As an illustration I attached a stolen screenshot from a recent PR by @titouv , his case is even worse than mine but basically we seem to be in the same situation : a Coolify with limited renaming options and a messy list.

I see 2 options to implement this : 

**1. Give an option to give an alias to containers:** 
In the name column of the "All Containers" section we could imagine an icon allowing to give an alias to the container. That alias would then be used by said section and graphs in the rest of the hub.
One issue might be that containers don't always redeploy using exactly the same name I think.

**2. Give the choice to the user of what key should be used to set the container's identifier:**
Since Beszel has access to the container's details (as seen in the details tab when we click on a container) we could imagine a setting on the settings page where the user could set which key he would want the containers to be identified by. In my case I think I would use the key  "Labels" => "coolify.serviceName". 
If the key doesn't exist the name would fallback to the container's name.

1. might be more straightforward for the users but more challenging technically so I think 2. might be better
### Screenshots

Image

### Category

UI / UX

### Affected Metrics

Containers

ErnestoMuniz:
## 📃 Description

This pull request introduces container aliasing with support for both manual and automatic alias management. 

closes #1385 

### ➕ Added

- Show container alias as first option and fallback to container name, both in the all containers table and in the charts.
- Manual alias editing in the "all containers" table.
- Auto container aliasing option in the system editing dialog. 
- You can still see the original container name if you hover over the alias in the all container table. 

## 📷 Screenshots

image

image

image


henrygd (commented):
Thank you, I appreciate your work on this.

However, I'm not sure that sending across all labels and controlling an alias in the UI is the best way to solve this problem.

It adds a potentially huge amount of extra data to transfer and store (containers can have thousands of labels), and the UI option may confuse people who aren't already familiar with this issue.

I'm a bit swamped at the moment, but I'll spin this up when I have time and spend a bit more time thinking about the best way to handle this.

svenvg93:
Just my two cents: wouldn’t it be easier to read a label like `beszel.name=<container_id>`? If that label is set on the container, it could be used as the container name.

henrygd:
We were talking about adding a `CONTAINER_LABEL` env var in #1385 which would work similarly but can be set to use existing labels like `coolify.serviceName`. So you wouldn't need to define a custom label for every container.

But the problem is that multiple containers could have the same value for a label, and the charts don't handle this properly because we rely on the container name being unique to link different records together.

We would probably need to store the container ID and link them with that. But that's not ideal because 1) it adds more data and 2) more importantly, IDs change when containers are recreated, which would break the continuity in the charts. So I'm not sure what the best solution is.

Maybe we could just do something in the UI where we detect if that system has a container starting with `coolify` and, if so, trim off the UUID suffix of all containers for that system wherever we display the container names (chart tooltip, All Containers table).

svenvg93:
> Maybe we could just do something in the UI where we detect if that system has a container starting with `coolify` and, if so, trim off the UUID suffix of all containers for that system wherever we display the container names (chart tooltip, All Containers table).

This would only work for Coolify. There are other platforms that could benefit from a similar solution, but that would require adding specific support for each of them.

That’s why I suggested using a custom label instead. It allows users to decide whether they want to use it or not and covers more use cases in my opinion. The agent could simply replace the current name value with the one defined in the label.

Redbirdma:
Can we please please have this implemented?  It's a pain in the neck to track containers without custom labels

PR #1792, opened 2026-03-04
Edit windows sensors management to not depends on Admin

KriKrixs:
## 📃 Description

a WIP version to get CPU temps on Windows without asking for admin privileges. It works by separating the agent services and LHM as 2 different services.

## 📖 Documentation

This is a WIP version. I'm not familiar enough with GO to complete the full change

## 📷 Screenshots

Capture d&#39;écran 2026-03-04 112914


henrygd:
Thanks for your work on this. I'm swamped right now but I'll get back to you as soon as I can.

henrygd:
To clarify, this uncouples the `beszel_lhm.exe` program from the agent, right? So the user needs to create a separate service for `beszel_lhm` which works independently and still needs admin privileges?

KriKrixs:
Yes it uncouples it, so i guess that the agent installer can create another nssm service with the same local system account that the agent is currently using to run beszel_LHM. From what i've tested, this fixed the admin privileges and UAC requirements to get all the temps for windows.

henrygd:
Thanks, does this method somehow provide access to more sensors compared to running the whole agent as admin like it needs to currently? Or is the goal only to limit how much of the program needs privileges?

KriKrixs:
No it's the same as if you tweak the nssm agent to run as admin. The only goal is to no give an admin Access to the agent and to avoid prompting the UAC panel so that (from my point of view) is more user friendly and easy to install

henrygd:
Okay, thank you. Sorry for asking so many questions, but can you briefly tell me a bit more about the difference with UAC prompting?

My assumption is you would get the same experience with both methods, since one program still needs to run as admin. I'm not the most experienced Windows user so I may be missing something obvious.

KriKrixs:
Dont be sorry 😉

From what i've experienced, separating the agent and beszel_lhm completely remove the requirements to run as admin one or the other(so it also remove the UAC prompt at every reboot). It seems like, right now, when the agent create the sub program for beszel_lhm, the sub process will lack some permissions that the local system account (the default account used by the agent) have already to get temps.

What i've tried is to run the agent as local system account from powershell and i didn't get the temps from beszel_lhm. But when i've separated them and run both as local system account from powershell. The magic happens and i got the temps.

This will means that the installation of beszel_agent will just need to change to create both services via nssm with the same exact rights as right now and no further configuration for the user after. It's just working fine

Like i said on the discussion, that's where i need someone to end the work here since i dont know a single thing about go or nssm or the installation script you made. I let the ai edit the agent quickly so you can understand what the thing is about.

Hope you understand what i've achieved, thanks for you amazing work on beszel and have a great weekend 😉

Issue #1537, opened 2025-12-15
[Feature]: Beszel Agent for Router Mikrotik


denispollini:
### Component

Hub

### Describe the feature you would like to see

Hi 
I want to say if it's possible to deploy the beszel agent on a Mikrotik Router?
Thanks

### Motivation / Use Case

Beszel Agent Mikrotik

### Describe how you would like to see this feature implemented

Beszel-Agent Mikrotik

### Category

Installation

### Affected Metrics

CPU

henrygd:
The binary is generally the best method for routers, but from a quick google I can't find instructions to run static binaries on RouterOS.

Seems like they might force you to use the container functionality: https://help.mikrotik.com/docs/spaces/ROS/pages/84901929/Container

If we don't have a prebuilt image available for your arch you should be able to do something like this:

```sh
# Clone the repository
git clone --branch v0.17.0 --depth 1 https://github.com/henrygd/beszel.git

# Navigate to the repository
cd beszel

# build the image with specific architecture
docker buildx build --platform linux/arm/v6 -t henrygd/beszel-agent:latest -f ./internal/dockerfile_agent .

# save the image as a file
docker save -o beszel-agent.tar henrygd/beszel-agent:latest

# transfer to router
rsync -av beszel-agent.tar admin@192.168.88.1:/data/disk1/
```

denispollini:
Hi
But the router has to have decent hardware to run container, now for monitoring mikrotik router i use Zabbix with SNMP.
But if there was a way to monitor via beszel it would be great that can i have all centralized in Beszel.

svenvg93:
If the router is powerfull enough you can follow the steps given above to build the image needed and run it on the  Mikrotik. 
Keep in mind that even the Mikrotiks RouterOS is based on linux, possible modifications might make it that not everything will work.

denispollini:
For the moment i use zabbix with snmp for mikrotik stuff related, thanks.

Eirikr70:
For what it's worth, it works for me, on a MikroTik hAP ax³, with the standard Beszel-agent image. What I don't have yet is the network traffic. Probably related to the fact I couldn't yet succeed in giving the container the additional capacities and the host network mode. I'll try digging some further.

Eirikr70:
I figured out how to set the container in host network mode, but I still don't have any network traffic. I'll go on investigating. Anyway I can confirm that the standard agent image works flawlessly.

marmar76:
[...]

PR #1795, opened 2026-03-05
Agent for SNMP Devices

marmar76:
## 📃 Description

This PR adds support for monitoring **MikroTik devices via SNMP** from the internal agent.

If the environment variable `BESZEL_AGENT_SNMP_TARGET` is set, the agent will query SNMP statistics from the specified MikroTik device instead of relying solely on local network metrics.
I tested it on hAP AX 2

The goal is to allow Beszel agents running inside a MikroTik container to collect accurate network interface statistics from the host router.

## 📖 How To Use it

On my case, I'm installing the container on the mikrotik itself, then its on the Container menu, i'm declaring the ENV Here
1. Declare ENV
Go to Container -> Envs, add one by one
image

2. Create VETH Interface
Go to Interfaces -> New VETH
image

3. Connect the VETH to the Bridge, so it can get IP from DHCP Go to Bridge -> Ports -> Add
image

4. Create the Container -> Container -> New
image
image

Then Click OK

### ➕ Added

* SNMP package for querying MikroTik interface statistics
* Support for `BESZEL_AGENT_SNMP_TARGET` environment variable

## 📷 Screenshots

image

image

cummulative download is accurate !!
image

## 💬 Context

This PR follows the discussion in issue #1537.

## 🙋 Notes

This is my first time working with Go, so feedback is very welcome.
The current implementation works for my setup, but I'm sure there are better or more idiomatic approaches.
Also, the agent size is getting bigger around 500kb
image

Any suggestions or improvements are appreciated!


marmar76:
Im also opening my beszel instance at https://beszel.kusuri.my.id/ 
email: guest@kusuri.my.id 
pass: NothingToSee
you can see the implementation yourself

henrygd:
Thanks so much for your work on this. I'm swamped right now but I'll get back to you as soon as I can.

bassiebal (commented):
Hi @marmar76,

As I was searching for this exact functionality I thought I'd read up on your PR and to speed things along here are some comments you can already have a look at pending the review of @henrygd:

Looking at it, I would argue this is a generic SNMP provider, not a MikroTik specific integration.

All OIDs used here are standard IF-MIB ([RFC 2863)](https://www.rfc-editor.org/rfc/rfc2863.html). They work on any SNMP-capable device: Cisco, UniFi, Synology, etc. The file should be renamed from mikrotik_helpers.go to something like snmp_network.go, and all Mikrotik prefixed functions and env vars should follow. `MIKROTIK_IP` becomes `SNMP_TARGET`, etc.

Like I mention in my comment as well, the existing codebase uses manager structs (dockerManager, gpuManager etc) for optional services. SNMP arguably fits that same pattern.

A NetworkIOProvider interface with a local `/proc` implementation as default and an SNMP implementation selected at startup when `SNMP_TARGET` is set would keep network.go clean. The SNMP client gets initialized once, stored on the provider struct, and reused for every poll. This also avoids the current issue of reinitializing `gosnmp.Default ` globals in multiple places.

Finally the SNMP version is hardcoded to v2c and the community string to "public". At minimum these should be env vars (`SNMP_TARGET`, `SNMP_COMMUNITY`, `SNMP_VERSION`). v3 support can come later but locking to v2c rules out anyone using v3.

Apologies for the wall of comments on someone else's repo. I just want to see this feature land and hopefully this feedback helps get it there faster.

Great start overall, especially for a first Go project. The collection logic works and your demo instance looks solid!

PS: an example of the interface approach I mentioned: 

```go
 // NetworkIOProvider abstracts network interface stats collection.
  type NetworkIOProvider interface {
      IOCounters() ([]psutilNet.IOCountersStat, error)
  }

  // localNetworkIO reads from /proc via gopsutil (default).
  type localNetworkIO struct{}

  func (l *localNetworkIO) IOCounters() ([]psutilNet.IOCountersStat, error) {
      return psutilNet.IOCounters(true)
  }

  // snmpNetworkIO reads interface stats from a remote SNMP target.
  type snmpNetworkIO struct {
      client *gosnmp.GoSNMP
  }

  func newSNMPNetworkIO(target, community string) (*snmpNetworkIO, error) {
      client := &gosnmp.GoSNMP{
          Target:    target,
          Community: community,
          Version:   gosnmp.Version2c,
          Timeout:   2 * time.Second,
      }
      if err := client.Connect(); err != nil {
          return nil, fmt.Errorf("snmp connect: %w", err)
      }
      return &snmpNetworkIO{client: client}, nil
  }

  func (s *snmpNetworkIO) IOCounters() ([]psutilNet.IOCountersStat, error) {
      // walk IF-MIB, return []psutilNet.IOCountersStat
  }
```

Then in `NewAgent()`:

```go
  if target, exists := utils.GetEnv("SNMP_TARGET"); exists {
      community, _ := utils.GetEnv("SNMP_COMMUNITY")
      if community == "" {
          community = "public"
      }
      agent.networkIO, err = newSNMPNetworkIO(target, community)
  } else {
      agent.networkIO = &localNetworkIO{}
  }
```

marmar76:
Hi @bassiebal, thank you for your feedback — I really appreciate it. I completely missed that my IP was hard‑coded. You're right: an SNMP implementation should behave more like a database connection — initialized once and kept alive. My current approach doesn’t reflect that well, so I’ll work on improving it based on your suggestions.

I’ve only been using SNMP on a MikroTik router, so I focused specifically on that. I didn’t realize other devices also commonly expose SNMP, so thanks for pointing that out.

I’d also love your feedback on this discussion: https://github.com/henrygd/beszel/issues/1537.
That one is about a first‑party approach where MikroTik itself hosts the agent. I’m curious what you think about using a third‑party agent instead, so everything is fully SNMP‑based when collecting data from the target.
Thanks again for taking the time to review my work!

bassiebal:
No problem! And I gave some feedback as well on the discussion. In my opinion a remote agent polling SNMP seems like the best approach but that entirely depends on the scope of this project. I can also imagine it a good solution to skip networking gear entirely to keep the project lean and simple.

marmar76:
Hi @bassiebal,
I’ve updated the code with your suggestion, but the agent size still hasn’t changed.
image

I also just learned that Go has this kind of abstraction—where you can call the same method but get different results depending on how the struct is initialized. It really taught me OOP on a deeper level.
I’ve installed it on my MikroTik, and the demo above is already using the newest image.
Thanks again for your guidance!

Eirikr70:
Hello, is the VulnCheck conflict a barrier to this pull request ?

russssl:
Is this being merged any time soon ?

marmar76:
> Is this being merged any time soon ?

I’ve updated the PR description.
It looks like the software owner might still be busy, so you can go ahead and build the image.
I’m planning to publish the image on Docker Hub, and I’ve been using this setup for about 4 months now—so it’s been stable and working well.
You can also check out my demo here: https://beszel.kusuri.my.id/
email: [guest@kusuri.my.id](mailto:guest@kusuri.my.id)
pass: NothingToSee

PR #1806, opened 2026-03-09
fix(alerts): skip "up" notification if "down" was never sent

### LINKED ISSUE #1519: [Bug]: Recovery notification sent but down is never sent
### Component

Hub & Agent

### Description

I randomly get both emails and Pushover notifications letting me know my device is "back up", but I never get the notification that it goes into down state. Seems to be the general device connection ("Status") sensor only so far. Changing the alert delay time to max. (60 minutes from default 10 minutes) seems to lower the rate of notifications.

Also, it's not limited to one device, I just don't have any demo notifications for the other in the image below.

See image for clarification: Image

### Expected Behavior

Device goes down => notify user of downtime => device goes back up => notify user of recovery.

### Steps to Reproduce

Set up an Ubuntu 24.04 machine (Hetzner in my case) with Beszel Hub 0.17.0 and another one with Beszel Agent 0.17.0, enable notifications for device "Status", set up Pushover and email notifications and wait for the "back up" notification.

### Category

Notifications & Alerts

### Affected Metrics

Sensors

### OS / Architecture

Ubuntu 24.04/x86 AMD (Hub) & Ubuntu 24.04/x86 Intel/AMD (Agent)

### Beszel version

0.17.0

### Installation method

Binary

### Configuration

```yaml
On Hetzner, both devices are on a shared internal network so all communication is automatically fully open (ports and other network requirements), standard Beszel config. Everything else works with no issues, just the notifications.
```

### Hub Logs

```json
{
  "id": "xyrvnkyh7gdw3mt",
  "created": "2025-12-10 18:15:19.684Z",
  "data": {
    "err": "dial tcp 10.0.0.4:45876: connect: connection refused",
    "system": "AZX-APP001"
  },
  "message": "System down",
  "level": 8
}
```

### Agent Logs

```shell
Dec 10 18:15:19 AZX-APP001 beszel-agent[3391911]: 2025/12/10 18:15:19 WARN Connection closed err=EOF
Dec 10 18:15:19 AZX-APP001 beszel-agent[3391911]: 2025/12/10 18:15:19 WARN Disconnected from hub
Dec 10 18:15:19 AZX-APP001 beszel-agent[3391911]: 2025/12/10 18:15:19 INFO WebSocket connected host=BESZEL_DOMAIN
Dec 10 18:21:56 AZX-APP001 beszel-agent[3391911]: 2025/12/10 18:21:56 WARN Connection closed err=EOF
Dec 10 18:21:56 AZX-APP001 beszel-agent[3391911]: 2025/12/10 18:21:56 WARN Disconnected from hub
Dec 10 18:21:56 AZX-APP001 beszel-agent[3391911]: 2025/12/10 18:21:56 INFO WebSocket connected host=BESZEL_DOMAIN
```

svenvg93:
## 📃 Description

Fixes a race condition that caused spurious "system back up" notifications without a preceding "system down" notification.

Before sending an "up" notification, check whether the alert's triggered field is true. The triggered flag is only set to true when `sendStatusAlert("down", ...)` actually runs, so if it's false, the "down" notification was never sent and the "up" should be suppressed.

Closes #1519

### DIFF
--- a/agent/gpu_amd_linux.go
+++ b/agent/gpu_amd_linux.go
@@ -33,8 +33,8 @@ func (gm *GPUManager) hasAmdSysfs() bool {
 		return false
 	}
 	for _, vendorPath := range cards {
-		vendor, err := os.ReadFile(vendorPath)
-		if err == nil && strings.TrimSpace(string(vendor)) == "0x1002" {
+		vendor, err := utils.ReadStringFileLimited(vendorPath, 64)
+		if err == nil && vendor == "0x1002" {
 			return true
 		}
 	}
@@ -88,12 +88,11 @@ func (gm *GPUManager) collectAmdStats() error {
 
 // isAmdGpu checks whether a DRM card path belongs to AMD vendor ID 0x1002.
 func isAmdGpu(cardPath string) bool {
-	vendorPath := filepath.Join(cardPath, "device/vendor")
-	vendor, err := os.ReadFile(vendorPath)
+	vendor, err := utils.ReadStringFileLimited(filepath.Join(cardPath, "device/vendor"), 64)
 	if err != nil {
 		return false
 	}
-	return strings.TrimSpace(string(vendor)) == "0x1002"
+	return vendor == "0x1002"
 }
 
 // updateAmdGpuData reads GPU metrics from sysfs and updates the GPU data map.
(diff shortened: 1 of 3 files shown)


henrygd:
Thanks Sven. I did some refactoring in addition to this check, so I'm fairly confident the problem is fixed. Strangely I was never able to reproduce it, but I think it was a race when the system switched almost immediately from up -> down -> up.

PR #1812, opened 2026-03-11
Charts: add max period up to 1 year

VACInc:
## Summary
- add 90 day and 1 year chart ranges
- add a Max Period setting that defaults to 30 days and caps the chart range selector
- retain 480m rollups for 1 year and add backend coverage for the new user setting default

## Testing
- go test -tags testing ./internal/users ./internal/records

### DIFF
--- a/internal/records/records.go
+++ b/internal/records/records.go
@@ -545,11 +545,11 @@ func deleteOldSystemStats(app core.App) error {
 		retention  time.Duration
 	}
 	recordData := []RecordDeletionData{
-		{recordType: "1m", retention: time.Hour},             // 1 hour
-		{recordType: "10m", retention: 12 * time.Hour},       // 12 hours
-		{recordType: "20m", retention: 24 * time.Hour},       // 1 day
-		{recordType: "120m", retention: 7 * 24 * time.Hour},  // 7 days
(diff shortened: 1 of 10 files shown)


VACInc:
@henrygd opened this as a draft to get your feedback/thoughts first -- added it as a enable/disable/set default option for those that don't have the storage/space/want it

any thoughts?

henrygd:
Sorry for the wait, I've been swamped and a couple things have taken longer than anticipated.

A configurable max retention period is a good idea. Maybe could be an env var rather than a user specific setting. Let me get the next release out and I'll try to look into this for 0.19.0 or 0.20.0.

VACInc:
I like it. Keeps things less chaotic in the UI

Gabriel-Graf:
@henrygd Hi, would you accept a PR that adds an env var for long-term retention, as you suggested?

  - DATA_RETENTION (also BESZEL_HUB_DATA_RETENTION), e.g. 180d. It only affects 480m records, and the default stays 30d.
  - Adds 90d / 180d / 1y chart ranges, shown only if the retention covers them.
  - No user setting, no migration.

  @VACInc happy to leave it to you if you'd like to rework this PR. Otherwise I'd open a new one and credit yours.

PR #1827, opened 2026-03-21
fix: correct DST offset handling in daily quiet hours

### LINKED ISSUE #1825: [Bug]: Daily quiet hours windows show "Inactive" and don't suppress notifications for DST-observing timezones
### Welcome!

### Component

Hub

### Problem Description

Daily quiet hours windows fail to suppress notifications and always display as **Inactive** for users in DST-observing timezones (e.g. US Eastern) during the summer DST period. One-time windows work correctly.

When saving a daily quiet hours window, the frontend converts the user's local time to UTC using a hardcoded historical date:

```js
new Date(`2000-01-01T${startTime}:00`).toISOString()
```

January 1, 2000 falls in **winter time** (e.g. EST = UTC−5), so the UTC conversion always uses the winter offset — even when the user is currently in **summer time** (e.g. EDT = UTC−4). This causes a 1-hour error in the stored UTC value for any user whose current DST offset differs from their winter offset.

For example, a user in EDT (UTC−4) entering **7:00 PM – 11:00 PM** gets stored as `00:00–04:00 UTC` instead of the correct `23:00–03:00 UTC`.

A second bug in `getWindowState` compounds this: it uses `now.getTimezoneOffset()` (the *current* date's offset) to convert the stored UTC time back to local time for display and state calculation, but the stored time was encoded with the *historical* winter offset for Jan 2000. This makes the window display as **Inactive** even when the user is within the intended hours.

One-time windows are unaffected because they use a direct `time.After(start) && time.Before(end)` UTC comparison with no hour extraction.

### Expected Behavior

The quiet hours windows should work as expected, even for DST-observing timezones.

### Steps to Reproduce

  1. Be in a DST-observing timezone during summer (e.g. US Eastern, EDT)                                                         
  2. Create a daily quiet hours window (e.g. 7:00 PM – 11:00 PM)
  3. Observe the window shows "Inactive" during the configured hours                                                             
  4. Confirm in browser console: stored UTC start hour is off by 1   

### OS / Architecture

linux/amd64(hub)

### Beszel version

latest

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Daily quiet hours windows were showing "Inactive" and failing to suppress notifications for users in DST-observing timezones (e.g. US Eastern).

There were two related bugs:

Encoding bug (handleSubmit): Times were encoded using the hardcoded date 2000-01-01, which always falls in winter. For US Eastern users currently in EDT (UTC−4), creating new Date('2000-01-01T19:00:00') applied the winter EST offset (UTC−5) instead — storing 00:00 UTC instead of the correct 23:00 UTC. This caused a 1-hour discrepancy for anyone in a DST period.

Decoding bug (getWindowState): The function converted stored UTC times back to local time using now.getTimezoneOffset() (the current DST offset). But since the times were encoded with a different (winter) offset, the decoded local time was off by 1 hour — causing the window to appear Inactive even when the user was within the configured timeframe.

Fixes #1825

### ✏️ Changed

- Replace 2000-01-01 with today's date when creating the UTC ISO string for daily windows, so the correct current DST offset is used.
- Use startDate.getTimezoneOffset() / endDate.getTimezoneOffset() (the stored date's offset) instead of now.getTimezoneOffset(), ensuring the UTC→local conversion is consistent with how the time was originally encoded.

### DIFF
--- a/internal/site/src/components/routes/settings/quiet-hours.tsx
+++ b/internal/site/src/components/routes/settings/quiet-hours.tsx
@@ -134,10 +134,10 @@ export function QuietHours() {
 			const startMinutes = startDate.getUTCHours() * 60 + startDate.getUTCMinutes()
 			const endMinutes = endDate.getUTCHours() * 60 + endDate.getUTCMinutes()
 
-			// Convert UTC to local time offset
-			const offset = now.getTimezoneOffset()
-			const localStartMinutes = (startMinutes - offset + 1440) % 1440
-			const localEndMinutes = (endMinutes - offset + 1440) % 1440
+			// Convert UTC to local time using the stored date's offset, not the current date's offset
+			// This avoids DST mismatch when records were saved in a different DST period
+			const localStartMinutes = (startMinutes - startDate.getTimezoneOffset() + 1440) % 1440
+			const localEndMinutes = (endMinutes - endDate.getTimezoneOffset() + 1440) % 1440
 
 			// Handle cases where window spans midnight
 			if (localStartMinutes <= localEndMinutes) {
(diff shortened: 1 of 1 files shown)


henrygd:
Perfect, much appreciated!

PR #1830, opened 2026-03-22
feat(hub): show "update available" notification in hub web UI

svenvg93:
## 📃 Description

This PR adds a "update available"  in the UI when a new version is available.

### ➕ Added

- Adds a `GET /api/beszel/newversion` endpoint that checks GitHub for a newer release and returns the tag and release URL if one is available. Results are cached server-side for 12 hours.

### ✏️ Changed

- Extracts a reusable `CheckForUpdate` function from the existing `ghupdate` package so the version check logic is not duplicated.

## 📷 Screenshots

Scherm­afbeelding 2026-03-22 om 16 18 39


henrygd:
Cool, thanks! For now I made it opt-in with `CHECK_UPDATES=true` just because I think it's nice to have the default behavior be to avoid hitting external services unless the user opts into doing so.

PR #1831, opened 2026-03-22
feat: persist view preferences and language to user settings

### LINKED ISSUE #1858: [Feature]: Default filters or bookmarkable filters
### Welcome!

### Component

Hub

### Description

I have 9 systems in my dashboard, but being a home network there are typical only 3 or so powered on at any given time.  I'd love to be able to default to the result of `View -> Status -> Up` but I could see this being useful for other folks in different situations.

I think feeding the options set into GET params would be ideal so users could bookmark any number of selections, but a default view would work too.

### Motivation / Use Case

Remove noise in the dashboard, get to what you need faster.

### LINKED ISSUE #2117: [Feature]: Make dashboard customizable
### Welcome!

### Component

Hub

### Description

It would be nice if the Columns shown on the Dashboard Page could be customized. For example, my dashboard shows a Battery column, although none of my Hosts have a Battery to monitor.

### Motivation / Use Case

With #2111 on the horizon, customizability would be even more beneficial, since the dashboard would get even wider.

### LINKED ISSUE #2142: [Feature]: Persistence preference settings
### Welcome!

### Component

Hub

### Description

Every time I login into Beszel one a new system or new browser I have to change the theme and Visible Fields hence wondering if could be possible to set them per user and always remaining the same.

### Motivation / Use Case

Having not to change the theme and Visible Fields every time I login specially during login with the toggle theme button being so far on the upper right corner of the screen.

svenvg93:
## 📃 Description

Language preference and column visibility are now saved to the user's server-side settings, so they persist across browsers and devices instead of being limited to localStorage.

closes #1858 
closes #2117
closes #2142

### ➕ Added

- `lang` and `cols` fields to UserSettings to store preferred language and column visibility server-side
- `statusFilter`, `viewMode`, `sortMode`, `grid`, and `displayMode` fields to `UserSettings`, persisting systems table and system detail view preferences across browsers and devices
- Brief green checkmark on the View button (systems table) and Settings button (system detail) after any view preference is saved

### ✏️ Changed

- Language preference is now included when saving General Settings and applied on login via `updateUserSettings`
- Column visibility is now synced to user settings with a 1-second debounce on every toggle (silently, no toast)
- Column visibility initializes from user settings on mount, with a `localStorage` fallback for backwards compatibility
- Status filter, view mode (table/grid), sort order, chart width (grid/full), and display mode (default/tabs) all follow the same pattern — server-authoritative with localStorage fallback, debounced silent save on change


milanobrtlik:
Hi Sven, I pulled your branch and went through it. The `saveUserSettings` split and the storage write-back both look right: the function is identical to mine, comment included, the `besz-` prefixed keys match what `useBrowserStorage` wrote, and the silent callers no longer drag the settings route into the main bundle. The description now covers the sortMode/statusFilter behaviour change, so my earlier point 2 is settled, and point 3 is moot — `saved` is load-bearing for the check icon now.

Three things I think are still off.

**1. `updateUserSettings` and `getLocale` race for the locale.**

`updateUserSettings` now calls `dynamicActivate(settings.lang)`, and `I18nApp` still calls `dynamicActivate(getLocale())` in its own effect. Neither awaits the other, so the winner is whichever *resolves* last, not whichever is called last.

The asymmetry is what makes it bite: `dynamicActivate("en")` takes the synchronous branch, every other locale awaits a dynamic import of its chunk. So with account lang `en` and a browser reporting `de`, on a device with no `lang` in localStorage, the settings request resolves and applies `en`, then the `de` chunk finishes downloading and overwrites it. A small JSON API call beating a translation bundle over the wire isn't an unusual outcome on a cold load — and "no `lang` in localStorage" is exactly the case this PR exists for.

Skipping the `getLocale()` activation when there's a valid auth store, or sequencing the two, would settle it.

**2. The six debouncers can drop each other's writes.**

Each key has its own `debounce(…, 1000)`, and `saveUserSettings` is read-modify-write: GET the current settings, PUT the merge. Two *different* keys changed within roughly one round trip of each other give you GET-GET-PUT-PUT, and the second PUT is built on a copy that predates the first.

The View dropdown makes this reachable — it stays open and holds view mode, status filter and the column toggles side by side, so clicking two of them half a second apart is ordinary use. Same for grid/display mode in the info bar.

The failure mode is the awkward part: `saveUserSettings` ends with `$userSettings.set(response)`, so the store reverts to the losing value, but the component keeps its local state and `applied` blocks the resync — the UI still shows the change that was just discarded, and the user only finds out on the next reload.

Repeated toggles of the *same* key are fine, they share one debouncer. A single shared debounced flush sending the keys accumulated in `$userSettings` would close the window, and the handlers already write there via `setKey`.

**3. `statusFilter` is the only one of the six without a browser-storage fallback.**

`cols`, `viewMode`, `sortMode`, `grid` and `displayMode` all read from storage before the server answers. `statusFilter` starts at `"all"` and jumps once `updateUserSettings` resolves — the same flash the write-back was restored to avoid, in the one place that never had a key to fall back on.

Nit: `info-bar.tsx` and `systems-table.tsx` have formatting the Biome config would rewrite. Nothing enforces it in CI, so it's easy to miss — `bun run check:fix` sorts it.

svenvg93:
Really appreciate the deep dive here — this is exactly the kind of review that's easy to miss on your own. All three points are legit, and I've pushed fixes for all of them:

- **Locale race** — the settings load and the browser-detected locale were fighting over which one "wins," and you're right that it could go either way. Now the settings-based one always takes priority when you're logged in.
- **Debounce issue** — you nailed it, saving two different settings close together could silently drop one of them. They now all get bundled into a single save instead of racing each other.
- **Missing fallback** — added the missing local fallback for the status filter so it doesn't flash back to "all" before your saved preference loads in.

Also cleaned up the formatting nit, thanks for flagging it.

Thanks again for taking the time to actually pull the branch and dig in — this was a great catch all around 🙏

ChangkeunJ:
@svenvg93 I've sent a small startup fix to your branch in https://github.com/svenvg93/beszel/pull/2. A reload with a saved login gets stuck before the settings request because `I18nProvider` is waiting for its first locale.

I checked the German-browser/English-account case with a delayed translation download, plus login/logout and reload, in both development and production builds using sample API responses. The account language still wins.

svenvg93:
> @svenvg93 I've sent a small startup fix to your branch in [svenvg93#2](https://github.com/svenvg93/beszel/pull/2). A reload with a saved login gets stuck before the settings request because `I18nProvider` is waiting for its first locale.
> 
> I checked the German-browser/English-account case with a delayed translation download, plus login/logout and reload, in both development and production builds using sample API responses. The account language still wins.

Thanks, much appreciated!

henrygd (changes requested):
Thanks all, looks good but I think we should probably remove the green check. It’s not really clear what it means to the user, and since it appears before the debounced save actually completes, it could be misleading. These preferences are low-stakes enough that silently saving in the background seems cleaner in my opinion.

svenvg93:
> Thanks all, looks good but I think we should probably remove the green check. It’s not really clear what it means to the user, and since it appears before the debounced save actually completes, it could be misleading. These preferences are low-stakes enough that silently saving in the background seems cleaner in my opinion.

Yeah make sense, I removed it.

henrygd (approved):
Thank you!

PR #1839, opened 2026-03-26
Prevent temperature collection from blocking agent stats

Jimbojones1:
## 📃 Description

  While testing the agent on a Jetson Orin Nano, I found that temperature collection could hang inside the gopsutil sensor path when the agent was running
  in Docker. When that happened, the entire GetData response stalled, which prevented the hub from receiving stats and system details.

  This PR adds a timeout around temperature collection so the agent fails open instead of blocking the whole stats response.

  ## 📖 Documentation

  No documentation changes.

  ## 🪵 Changelog

  ### ✏️ Changed

  - Wrap temperature collection in a timeout before reading sensors.
  - Continue returning agent stats even if temperature collection does not complete in time.

  ### 🔧 Fixed

  - Prevent GetData from hanging indefinitely when sensor temperature collection blocks on a Jetson Orin Nano Docker setup.
  - Skip stale temperature updates cleanly when temperature collection times out.

  ## 📷 Screenshots

  No UI changes.

### DIFF
--- a/agent/sensors.go
+++ b/agent/sensors.go
@@ -2,12 +2,14 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"path"
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 	"unicode/utf8"
 
 	"github.com/henrygd/beszel/agent/utils"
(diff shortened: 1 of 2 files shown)


henrygd:
Great, thanks!

PR #1845, opened 2026-03-28
Auto-update system host if the agent's IP has changed

DX-Kevin:
Auto-update system host if the agent's IP has changed
updateSystemHost checks if the current connection's IP matches the system's recorded host

### DIFF
--- a/internal/hub/agent_connect.go
+++ b/internal/hub/agent_connect.go
@@ -3,6 +3,7 @@ package hub
 import (
 	"context"
 	"errors"
+    "log/slog"
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


svenvg93:
Quick question: what will happen when I configured a dns name when adding the system and not a ip address?

btw please remove the committed agent binary from the commit 🙂

Kinsue:
The necessity of this modification isn't entirely clear. Furthermore, as @svenvg93  pointed out, the current implementation relies on a simple string comparison, which would forcibly overwrite the domain name configured by the user.

Regarding the use case you mentioned, here are a few suggestions:
- **For frequently changing IPs**: The best practice is to use a Domain Name combined with DDNS. Handling dynamic IP resolution directly within the codebase exceeds the intended scope of this project, which aims to remain a lightweight monitoring solution.
- **For restricted environments (No Domain/DDNS)**: the current version (v1.18.6) actually supports communication between a private network Agent and a public network Hub, as illustrated below:
image
You can simply enter any placeholder IP in the frontend. 

Actually, older versions did indeed have an issue where Agents would go offline when the IP changed. But I haven't encountered it at all since upgrading to the newer releases.

I'd recommend trying out the latest version to see if it resolves your issue. 😄

PR #1880, opened 2026-04-02
fix(install): use daemon user on OPNsense to survive reboots

### LINKED ISSUE #1221: [Bug]: OPNsense deletes beszel user and group upon reboot
### Component

Agent

### Description

As someone new to OPNsense this caught me by surprise. After rebooting OPNsense the `beszel` entries in `/etc/passwd` and `/etc/group` were gone, so the agent would not start. I'm not familiar with the "proper" way to add a persistent user to OPNsense (other than using the GUI).

As a workaround for now I ran:

```
service -E beszel_agent_user=daemon beszel-agent start
```

### Expected Behavior

Agent starts after reboot.

### Steps to Reproduce

Install using Hub-generated command line. Reboot.

I should mention that I changed the OPNsense system name before rebooting, but I don't think that had anything to do with the issue.

### Category

Installation

### Affected Metrics

Other

### OS / Architecture

freebsd/amd64 (agent)

### Beszel version

0.12.12

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Auto-detect OPNsense in the agent install script and use the `daemon` system user instead of creating a `beszel` user. OPNsense regenerates `/etc/passwd` and `/etc/group` on every reboot, which wipes any custom users added via `pw user add`, preventing the agent service from starting after a reboot.

Closes #1221

### ✏️ Changed

- FreeBSD agent install script now detects OPNsense and skips `beszel` user creation, using the `daemon` user instead
- Agent directory and env file ownership use the resolved service user rather than a hardcoded `beszel`
- `beszel_agent_user` is now explicitly written to `rc.conf` via `sysrc` on all FreeBSD installs

### DIFF
--- a/supplemental/scripts/install-agent.sh
+++ b/supplemental/scripts/install-agent.sh
@@ -12,6 +12,10 @@ is_freebsd() {
   [ "$(uname -s)" = "FreeBSD" ]
 }
 
+is_opnsense() {
+  [ -f /usr/local/etc/opnsense-version ] || [ -f /etc/opnsense-release ]
+}
+
 is_glibc() {
   # Prefer glibc-enabled agent (NVML via purego) on linux/amd64 glibc systems.
   # Check common dynamic loader paths first (fast + reliable).
@@ -549,6 +553,7 @@ else
 fi
 
 # Create a dedicated user for the service if it doesn't exist
+AGENT_USER="beszel"
 echo "Configuring the dedicated user for the Beszel Agent service..."
 if is_alpine; then
   if ! id -u beszel >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


henrygd:
Thank you!

supplemental/scripts/install-agent.sh
```diff
@@ -12,6 +12,10 @@ is_freebsd() {
   [ "$(uname -s)" = "FreeBSD" ]
 }
 
+is_opnsense() {
+  [ -f /usr/local/etc/opnsense-version ] || [ -f /etc/opnsense-release ]
```
davidemyers: I have two OPNsense systems, running 25.7.11_9 and 26.1.5, and neither has either of these files. They both have `/usr/local/sbin/opnsense-version`.

henrygd: Thank you, this has been updated: https://github.com/henrygd/beszel/commit/eb9a8e1ef93901451ae2617432cfd134d834c0ea

PR #1886, opened 2026-04-03
feat(agent): collect and display stopped/exited containers

### LINKED ISSUE #1786: [Feature]: Collect and display stopped/exited/crashed containers
### Welcome!

### Component

Hub & Agent

### Description

Currently the agent queries the Docker API at `/containers/json` without the `all=1` flag, which means only running containers are ever collected, stored, or displayed.

It would be useful to also collect stopped, exited, restarting, and paused containers so users can:

- See at a glance how many containers are down or crashed
- Get a full inventory of all containers on a system, not just healthy/running ones
- Potentially trigger alerts when a container unexpectedly stops

### Motivation / Use Case

Users running many containers need to know when something has crashed without having to open a terminal. A "stopped" badge or count would surface this immediately.

svenvg93:
## 📃 Description

Collects and displays stopped, exited, and paused Docker containers in addition to running ones. Stopped containers appear in the containers table with their Docker status string (e.g. `Exited (0) 2 hours ago`) and zero CPU/memory/network stats. A colored status dot is shown next to the container name to indicate state at a glance.

closes #1786

### ➕ Added

- Collect stopped/exited/paused containers from the Docker API by default
- Status indicator dot next to container name (green = running, yellow = restarting, muted = paused/created, red = exited/dead)
- `DOCKER_RUNNING_ONLY` environment variable to opt out and collect only running containers (previous behavior)

### ✏️ Changed

- Docker container list query now uses `?all=1` by default

## 📷 Screenshots

Scherm­afbeelding 2026-04-03 om 20 19 09


boxwork-ctrl:
Thanks for working on this. I ran into this exact limitation while building a read-only container status view from Beszel’s API. Since only running containers are currently collected, a stopped container becomes indistinguishable from a stale/disappeared record for an external dashboard.

The all=1 behavior in this PR would make the containers API much more useful for that kind of monitoring. Happy to test this on a small multi-host Docker setup if testing would be useful.

PR #1900, opened 2026-04-07
fix: redirect-based OAuth fallback for browsers blocking popups

### LINKED ISSUE #1848: [Bug]: Ios Brave Browser OAuth Popup blocked
### Welcome!

### Component

Hub

### Problem Description

When trying to log in with Pocket-ID, I get “Please enable popups for this site,” even though I disabled Brave Shields and turned off the “block pop-ups” setting. This could be a Brave issue because when I request the desktop site on the mobile Brave browser app, it works and I get the popup. But I don’t quite understand why the OAuth login is even done with a popup. The only other service that does these popup logins is PayPal, and it’s very annoying. Is this a limitation of PocketBase?

### Expected Behavior

Direct redirect to the OAuth provider with no need for a popup.

### Steps to Reproduce

Use Brave Browser on iOS/iPadOS and try to log in with an OAuth provider.

### OS / Architecture

ignored

### Beszel version

ignored

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Implements a redirect-based OAuth fallback for browsers that block popups (e.g., Brave on iOS). When `window.open()` returns `null` (popup blocked), instead of showing an error, the app now automatically falls back to a full-page redirect OAuth flow using PocketBase's `authWithOAuth2Code`.

**Note:** The redirect flow requires users to register their app's base URL (e.g., `https://beszel.example.com`) as an additional OAuth redirect URI in their provider settings, alongside the existing `{url}/api/oauth2-redirect` entry.

Closes #1848

### ✏️ Changed

- OAuth login on iOS/mobile: when `window.open()` is blocked, automatically fall back to a redirect-based OAuth flow instead of showing a "Please enable pop-ups" error
- On app load, detect OAuth callback parameters (`?code=...&state=...`) and complete the authentication via `authWithOAuth2Code`, then clean up the URL


henrygd:
Thank you! I removed the fallback and instead made it the default behavior if `OAUTH_DISABLE_POPUP=true`.

PR #1905, opened 2026-04-09
feat: add last-updated timestamp column to dashboard

### LINKED ISSUE #1876: [Feature]: Show last-updated timestamp on the dashboard
### Welcome!

### Component

Hub

### Description

It would be helpful to display a "last updated" timestamp on the main dashboard — either per-system or as a global indicator. Right now there's no way to tell at a glance whether the data you're looking at is current or stale.

### Motivation / Use Case

Sometimes the dashboard silently stops updating (connection drop, agent hiccup, browser tab backgrounded, etc.) and everything looks normal because the metrics are still displayed — they're just frozen. A visible timestamp would make it immediately obvious when data has gone stale, so you're not making decisions based on outdated information.

vincent067:
This PR implements the feature requested in #1876 - showing a "last updated" timestamp on the main dashboard to help users identify stale data.

## Changes
- Added new "Last Seen" column to the systems table
- Added formatRelativeTime() utility function for human-readable timestamps
- The column shows relative time (e.g., "2 minutes ago") with tooltip for exact date
- Column is hidden by default to avoid cluttering the default view

## Why this is useful
The relative time format helps users quickly identify when a system's data was last updated, making it obvious if the data has gone stale due to connection issues or agent problems.

Closes #1876

### DIFF
--- a/internal/site/src/components/systems-table/systems-table-columns.tsx
+++ b/internal/site/src/components/systems-table/systems-table-columns.tsx
@@ -32,6 +32,8 @@ import {
 	copyToClipboard,
 	decimalString,
 	formatBytes,
+	formatRelativeTime,
+	formatShortDate,
 	formatTemperature,
 	parseSemVer,
 	secondsToUptimeString,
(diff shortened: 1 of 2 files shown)


henrygd:
Let me get back to you on this. Right now it won't work properly because the record's `updated` field is updated when anything on the record changes. So even if you pause / unpause the system.

We want a time for when the agent last provided data, so we would either need to add that to the `info` json blob or change the `updated` column to not auto-update, then set it manually when the agent data updates.

vincent067:
Hey @henrygd! 👋 Just wanted to check in on this PR - no rush at all, I know you're probably busy with all the recent releases!

The "Last Seen" column has been working great in my local testing. It really helps spot when an agent has gone quiet. Happy to make any changes if you'd like a different approach to the relative time formatting or column defaults.

Thanks again for all the work you put into Beszel - the new chart refactoring in the latest release looks amazing! 🙏

PR #1909, opened 2026-04-10
fix(cron): log unhandled records cleanup errors

alrs:
This picks up five errors that happen inside a cron job to no effect.

Previously, the cleanup tasks would return any encountered error and fail before the other cleanup tasks ran. The error was left unhandled.

Because the Pocketbase cron doesn't do much in the way of error handling, this approach turns the failures into log messages and continues trying subsequent tasks even if one fails.

### DIFF
--- a/internal/records/records.go
+++ b/internal/records/records.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"log/slog"
 	"math"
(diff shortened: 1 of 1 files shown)


henrygd:
Makes sense, thanks. I assumed the error would be logged if returned, but even so this change still allows the other deletion operations to continue.

PR #1917, opened 2026-04-13
feat: sync display name with hostname

### LINKED ISSUE #1910: [Feature]: Update display name when hostname changes
### Welcome!

### Component

Hub

### Description

I updated the hostname on one of my servers and the hostname field in the device pages update, but not the name on the main page.

### Motivation / Use Case

Clarity

svenvg93:
## 📃 Description

Adds an optional per-system setting to automatically keep the display name in sync with the agent's hostname. When enabled, the hub updates the system's display name whenever the hostname reported by the agent changes.

closes #1910

### ➕ Added

- `sync_name` boolean field on `systems` collection (migration included)
- "Sync name with hostname" checkbox in the add/edit system dialog

### ✏️ Changed

- Hub requests system details on every poll cycle when `sync_name` is enabled (instead of only on first fetch), so hostname changes are picked up without a hub restart

## 📷 Screenshots

Scherm­afbeelding 2026-04-13 om 19 10 40


svenvg93:
@henrygd I made the sync optional, would it not be better to just always sync it when the agent connects?

henrygd:
Sorry, I missed this one.

I don't think we should automatically sync agent names with hostnames. Some may want that but others won't.

What about an env var on the hub to enable this rather than controlling per system in the UI? If someone does want it, I assume they'll want it for all systems. Env var seems like the simplest solution to me.

svenvg93:
> Sorry, I missed this one.
> 
> I don't think we should automatically sync agent names with hostnames. Some may want that but others won't.
> 
> What about an env var on the hub to enable this rather than controlling per system in the UI? If someone does want it, I assume they'll want it for all systems. Env var seems like the simplest solution to me.

No worries at all, I know you're working your way through all the open PRs.

You're right, I've changed it to an env var set on the hub. Setting SYNC_SYSTEM_NAMES=true makes the hub update each system's name to the hostname the agent reports when it connects. I removed the per-system checkbox and the migration, so the change is now just a few lines in system.go plus a test.

henrygd (approved):
Gracias amigo. I'll update the docs.

PR #1919, opened 2026-04-14
agent: Allow agent to connect to hub via SOCKS5 proxy

marmelo:
## 📃 Description

Some machines may not have a public IP and need to connect to the internet via a proxy.
Ideally both HTTP and SOCKS proxies should be supported, but SOCKS would be enough.

The Agent uses https://github.com/lxzan/gws to establish a WebSocket connection to the Hub.

GWS already supports basic proxy support via `golang.org/x/net/proxy`, as seen in their example:
https://github.com/lxzan/gws?tab=readme-ov-file#proxy

This change will allow to define the `ALL_PROXY` or `all_proxy` env var, which will create the Dialer that proxies the WebSocket request.
If both `ALL_PROXY` and `all_proxy` are missing or they do not have a valid scheme (only `socks5` and `socks5h` are allowed at the moment), a direct connection will be used.

Example use:
```bash
ALL_PROXY="socks5://127.0.0.1:1080"
ALL_PROXY="socks5://username:password@127.0.0.1:1080"
```

## 📖 Documentation

https://github.com/henrygd/beszel-docs/pull/57

### ➕ Added

- Agent `ALL_PROXY` environment variable

### DIFF
--- a/agent/client.go
+++ b/agent/client.go
@@ -20,7 +20,7 @@ import (
 	"github.com/fxamacker/cbor/v2"
 	"github.com/lxzan/gws"
 	"golang.org/x/crypto/ssh"
- 	"golang.org/x/net/proxy"
+	"golang.org/x/net/proxy"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


henrygd (approved):
Looks great, thanks so much!

PR #1926, opened 2026-04-15
fix(cron): log CreateLongerRecords errors

alrs:
This is a follow-up PR that does for `RecordManager.CreateLongerRecords()`what was done for `RecordManager.DeleteOldRecords()` in https://github.com/henrygd/beszel/pull/1909. A comment is added in both locations.


henrygd (approved):
Thanks!

PR #1889, opened 2026-04-04
fix(hub): case-insensitive email login

svenvg93:
## 📃 Description

Normalize email addresses to lowercase during login and registration to prevent case-sensitive authentication failures. Users who registered with mixed-case emails (e.g., `User@Example.com`) could not log in using the lowercase variant (`user@example.com`).

### 🔧 Fixed

- Email login is no longer case-sensitive — addresses are normalized to lowercase on both the frontend (login/register form) and backend (user creation endpoint)

PR #1946, opened 2026-04-23
fix(hub): case-insensitive email login (#1887)

### LINKED ISSUE #1887: [Bug]: Login E-Mail is case Sensitive
### Welcome!

### Component

Hub

### Problem Description

So as a Login you have to use an E-Mail adress.
But E-Mails are never Case Sensitive.
The Login IS case sensitive.

### Expected Behavior

not case sensitive Email/username field for login

### Steps to Reproduce

1. Register an EMail adress with Uppercase letters.
2. try to login with lowercase letters.

### OS / Architecture

Debian 13

### Beszel version

0.18.6

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json
{
  "id": "uavj4wtsromuagc",
  "created": "2026-04-03 19:33:16.249Z",
  "data": {
    "auth": "",
    "details": "invalid login credentials",
    "error": "Failed to authenticate.",
    "execTime": 0.255281,
    "method": "POST",
    "referer": "http://192.168.1.2:30333/_/",
    "remoteIP": "192.168.1.200",
    "status": 400,
    "type": "request",
    "url": "/api/collections/_superusers/auth-with-password",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:149.0) Gecko/20100101 Firefox/149.0",
    "userIP": "192.168.1.200"
  },
  "message": "POST /api/collections/_superusers/auth-with-password",
  "level": 8
}

{
  "id": "kf7g2odylajguk5",
  "created": "2026-04-03 19:33:07.057Z",
  "data": {
    "error": "failed to locate a sendmail executable path"
  },
  "message": "[recordAuthResponse] Failed to send login alert",
  "level": 4
}
```

### Agent Logs

```shell

```

BootstrapperSBL:
## Description

Fixes #1887. Users who register with a mixed-case email (e.g. `Foo@bar.com`) could not log in as `foo@bar.com` because PocketBase's password auth performs an exact-match lookup on the email field.

### Before
```
POST /api/collections/_superusers/auth-with-password
{"identity":"foo@bar.com","password":"..."}

400 Bad Request: Failed to authenticate. (invalid login credentials)
```

### After
Same request succeeds regardless of the case used at login.

## Approach

Two small hooks in the `users` package:

1. `NormalizeEmail` is bound to `OnRecordCreate` for both `users` and `_superusers`. It lowercases the email before save so stored values are canonical going forward. The existing PocketBase unique email index then naturally rejects case-variant duplicates.
2. `ResolveAuthIdentity` is bound to `OnRecordAuthWithPasswordRequest`. When PocketBase's exact-match lookup does not find a record, it falls back to `[[email]] = {:email} COLLATE NOCASE`. This covers accounts created on versions without the normalize hook.

No DB schema change, no migration, nothing to break on existing installs. The fix is scoped to the auth path and stays within PocketBase's hook API.

### Fixed

- Login is now case-insensitive for the email identity on both the `users` and `_superusers` collections. Fixes #1887.

### Changed

- New user/superuser records are saved with the email lowercased, preventing future case-variant duplicates.

## Tests

`internal/hub/case_insensitive_email_test.go` adds two test functions:

- `TestEmailIsNormalizedOnCreate` — verifies that freshly created users/superusers have their email persisted lowercase, and that a case-variant second registration is rejected by the unique index.
- `TestCaseInsensitiveEmailLogin` — simulates a legacy account whose email was stored mixed-case before this fix, then exercises the `/auth-with-password` endpoint for `users` and `_superusers` with lowercase, uppercase and original-case identities. Also covers the negative paths (wrong password, unknown email).

All hub package tests pass. The pre-existing `agent` and `internal/alerts` test flakes on `upstream/main` are unchanged by this PR.

### DIFF
--- a/internal/hub/case_insensitive_email_test.go
+++ b/internal/hub/case_insensitive_email_test.go
@@ -1,0 +1,98 @@
+//go:build testing
+
+package hub_test
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	beszelTests "github.com/henrygd/beszel/internal/tests"
+	pbTests "github.com/pocketbase/pocketbase/tests"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// authWithPasswordScenario builds an API scenario for the auth-with-password
+// endpoint. Keeps the test tables below compact.
+func authWithPasswordScenario(name, collection, identity, password string, status int, content []string, factory func(t testing.TB) *pbTests.TestApp) beszelTests.ApiScenario {
+	return beszelTests.ApiScenario{
+		Name:            name,
+		Method:          http.MethodPost,
+		URL:             fmt.Sprintf("/api/collections/%s/auth-with-password", collection),
+		Body:            strings.NewReader(fmt.Sprintf(`{"identity":%q,"password":%q}`, identity, password)),
+		ExpectedStatus:  status,
+		ExpectedContent: content,
+		TestAppFactory:  factory,
+	}
+}
+
+// TestEmailIsNormalizedOnCreate verifies that new user/superuser records have
+// their email lowercased on save. This prevents future case-variant duplicates
+// and keeps the stored data canonical.
+func TestEmailIsNormalizedOnCreate(t *testing.T) {
+	hub, _ := beszelTests.NewTestHub(t.TempDir())
+	defer hub.Cleanup()
+
+	// StartHub binds the event hooks; the returned error is about the TestApp
+	// not being a full *pocketbase.PocketBase and is expected here
+	_ = hub.StartHub()
+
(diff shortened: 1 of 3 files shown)


BootstrapperSBL:
Heads up: I just noticed @svenvg93 opened #1889 back on April 4 for the same issue. Missed it when I went scouting — sorry for the duplicate. The approach there is simpler (lowercase-on-write) and @henrygd raised the concern about legacy mixed-case accounts that this PR's COLLATE NOCASE fallback is meant to cover. Happy to close this one if you'd rather iterate on #1889.

henrygd:
Apologies, I went on a months long break right around when you posted this. I'll look further into it when I can. Ideally we'd be able to fix just with a migration file that adds `COLLATE NOCASE` to `users` and `_superusers` collections.

PR #1932, opened 2026-04-17
fix(agent): keep every EXTRA_FILESYSTEMS mount when device keys collide

BootstrapperSBL:
## What

`EXTRA_FILESYSTEMS` now correctly registers every listed mount when multiple entries sit behind systemd-automount (or any other pseudo-device that reuses a single device identifier across mounts).

Fixes #1931.

## Why

`agent/disk.go`'s `registerFilesystemStats` dedupes by device-derived key. For most real devices that's correct — `/dev/sda1` can only be mounted in one place at a time. But `x-systemd.automount` reports every mount with the literal device `systemd-1`, so the fsStats key collides on the first entry and every subsequent mount is silently dropped (no log line, no error).

Reporter's config, which should register two disks, only surfaces one:

```
EXTRA_FILESYSTEMS=/media/video__Video,/media/backup__Backup
```

```
INFO Detected disk name=Video device=systemd-1 mount=/media/video io=systemd-1 root=false
DEBUG Extra FS data=map[Video:0x...]
```

Only `Video` ever makes it into `fsStats`; `Backup` is silently discarded because `fsStats["systemd-1"]` was already set.

## How

`registerFilesystemStats` now distinguishes two "existing key" cases:

1. **Same mountpoint already registered** — real duplicate. Still rejected with `ok=false` (the old behaviour).
2. **Different mountpoint under the same key** — pseudo-device collision. Fall back to a unique key derived from the mountpoint basename (`/media/backup` → `backup`), with a `-N` suffix if that basename also collides.

A small `uniqueFsStatsKey` helper encapsulates the basename/suffix logic.

## Before / After

For `EXTRA_FILESYSTEMS=/media/video__Video,/media/backup__Backup`:

| | v0.18.7 | After |
|---|---|---|
| fsStats keys | `{"systemd-1": Video}` | `{"systemd-1": Video, "backup": Backup}` |
| Mounts visible in UI | Video only | Video and Backup |

No change for any setup where the device key uniquely identifies the mount (i.e. every real block device). The new path only triggers on key collisions, so existing user-visible names stay stable.

## Test plan

- [x] `go build ./agent/...`
- [x] `gofmt -l agent/disk.go agent/disk_test.go` — clean
- [x] `go vet ./agent/...`
- [x] `go test -race ./agent/...` — all pass, including three tests covering the new paths:
  - updated "skips re-registration of the same mountpoint" (true duplicate case)
  - new "keeps both mounts when a pseudo-device key collides (issue #1931)"
[...]

PR #1951, opened 2026-04-24
fix(hub): stop launching system updaters after manager shutdown

### LINKED ISSUE #1950: [Bug]: Panic / Nil pointer dereference in `internal/hub/systems/system.go` during tests
### Welcome!

### Component

Hub

### Problem Description

During the Nix build process on **Apple Silicon (Darwin/ARM64)**, the test suite for `github.com/henrygd/beszel/internal/alerts` consistently fails with a nil pointer dereference. 

The panic occurs in a background goroutine started by `StartUpdater`. The stack trace shows it failing inside the PocketBase `RecordQuery` core. Interestingly, **this issue does not seem to reproduce on x86_64 Linux/Darwin builds**, suggesting a potential race condition that is more susceptible to the ARM memory model or specific to how Nix handles ARM sandboxing.

I have found that wrapping the logic in a `recover()` allows the build to finish, but the underlying nil dereference in `getRecord` needs an architectural fix.

```nix
  postPatch = ''
    substituteInPlace internal/hub/systems/system.go \
      --replace-fail "func (sys *System) StartUpdater() {" "func (sys *System) StartUpdater() { defer func() { recover() }();"
  '';
```

### Expected Behavior

The test suite should complete without a runtime panic. The `StartUpdater` function should safely handle cases where the underlying PocketBase app or record is not available.

### Steps to Reproduce

Run `nix build -L nixpkgs#beszel` on an Apple-Silicon device.
It will fail with the following error message:
```log
beszel> Running phase: checkPhase
beszel> ok      github.com/henrygd/beszel/agent 1.782s
beszel> ok      github.com/henrygd/beszel/agent/deltatracker    0.328s
beszel> ok      github.com/henrygd/beszel/agent/health  0.357s
beszel> ok      github.com/henrygd/beszel/agent/utils   0.331s
beszel> panic: runtime error: invalid memory address or nil pointer dereference
beszel> [signal SIGSEGV: segmentation violation code=0x2 addr=0xb0 pc=0x101467140]
beszel>
beszel> goroutine 38652 [running]:
beszel> github.com/pocketbase/pocketbase/core.(*BaseApp).RecordQuery(0x6d7eee754008, {0x102a41b40?, 0x6d7eee446008?})
beszel>         /nix/var/nix/builds/nix-71066-3105078627/source/vendor/github.com/pocketbase/pocketbase/core/record_query.go:38 +0x70
beszel> github.com/pocketbase/pocketbase/core.(*BaseApp).FindRecordById(0x6d7eee754008, {0x1028b7480?, 0x102a76260?}, {0x6d7eedefeee0, 0xf}, {0x0, 0x0, 0x1015f1d58?})
beszel>         /nix/var/nix/builds/nix-71066-3105078627/source/vendor/github.com/pocketbase/pocketbase/core/record_query.go:223 +0x78
beszel> github.com/henrygd/beszel/internal/hub/systems.(*System).getRecord(0x6d7eee63f0a0)
beszel>         /nix/var/nix/builds/nix-71066-3105078627/source/internal/hub/systems/system.go:347 +0x50
beszel> github.com/henrygd/beszel/internal/hub/systems.(*System).setDown(0x6d7eee63f0a0, {0x102a7d940, 0x6d7eee692550})
beszel>         /nix/var/nix/builds/nix-71066-3105078627/source/internal/hub/systems/system.go:362 +0x94
beszel> github.com/henrygd/beszel/internal/hub/systems.(*System).StartUpdater(0x6d7eee63f0a0)
beszel>         /nix/var/nix/builds/nix-71066-3105078627/source/internal/hub/systems/system.go:88 +0xc4
beszel> created by github.com/henrygd/beszel/internal/hub/systems.(*SystemManager).AddSystem in goroutine 37891
beszel>         /nix/var/nix/builds/nix-71066-3105078627/source/internal/hub/systems/system_manager.go:253 +0x17c
beszel> FAIL    github.com/henrygd/beszel/internal/alerts       63.711s
beszel> FAIL
```

### OS / Architecture

MacOS/arm

### Beszel version

0.18.7

### Installation method

Nix

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

BootstrapperSBL:
## Description

Fixes #1950.

The hub test suite intermittently panics on Apple Silicon during `go test ./internal/alerts/...` with a nil pointer dereference inside PocketBase's `RecordQuery`. The trace ends in `StartUpdater -> setDown -> getRecord -> FindRecordById`, with the goroutine created by `SystemManager.AddSystem`.

### Root cause

`SystemManager.Initialize` spawns a background goroutine that walks the systems loaded from the database and calls `AddSystem` for each one with a `time.Sleep` between calls (up to 2s per system). When a test's `defer hub.Cleanup()` runs while that loop is still sleeping, the loop is not aware of the shutdown:

1. `RemoveAllSystems` iterates `sm.systems.GetAll()` and cancels every system already added.
2. `TestApp.Cleanup` calls `ResetBootstrapState`, nilling the underlying DB.
3. The staggered loop wakes from its sleep, calls `AddSystem` for the next system, and starts a `StartUpdater` goroutine whose `sys.manager.hub` already has a nil DB. The first `update()` call inevitably fails (no real agent), `setDown -> getRecord -> FindRecordById` then panics inside `app.ConcurrentDB()`.

The window is much wider on Apple Silicon because of weaker memory ordering and slower TCP teardown, which is why x86_64 builds usually pass.

### Fix

Give `SystemManager` its own `context.Context` that is cancelled on shutdown:

- The staggered Initialize loop replaces `time.Sleep(sleepTime)` with a `select` on `sm.ctx.Done()` and returns early once shutdown is signalled.
- `AddSystem` refuses to add a system (and therefore does not spawn `StartUpdater`) once `sm.ctx` is cancelled.
- The test-only `RemoveAllSystems` helper cancels `sm.ctx` before tearing systems down, so by the time `TestApp.Cleanup` clears the DB the staggered loop is guaranteed not to start any new updaters.

No `recover()` is added; the production paths are unchanged on the happy path.

### Before / After

Before (from the issue, also reproducible locally on darwin/arm64):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0xb0]
goroutine N [running]:
github.com/pocketbase/pocketbase/core.(*BaseApp).RecordQuery(...)
github.com/henrygd/beszel/internal/hub/systems.(*System).getRecord(...)
github.com/henrygd/beszel/internal/hub/systems.(*System).setDown(...)
github.com/henrygd/beszel/internal/hub/systems.(*System).StartUpdater(...)
created by github.com/henrygd/beszel/internal/hub/systems.(*SystemManager).AddSystem
FAIL    github.com/henrygd/beszel/internal/alerts
```

After: `go test ./internal/alerts/... ./internal/hub/systems/...` is green across 5 consecutive runs on darwin/arm64.

### Fixed

- Hub system manager no longer starts `StartUpdater` goroutines against a hub that is being torn down, fixing a flaky panic in `internal/alerts` tests on Apple Silicon (#1950).

### DIFF
--- a/internal/hub/systems/system_manager.go
+++ b/internal/hub/systems/system_manager.go
@@ -1,6 +1,7 @@
 package systems
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"time"
@@ -45,6 +46,8 @@ type SystemManager struct {
 	systems       *store.Store[string, *System]         // Thread-safe store of active systems
 	sshConfig     *ssh.ClientConfig                     // SSH client configuration for system connections
 	smartFetchMap *expirymap.ExpiryMap[smartFetchState] // Stores last SMART fetch time/result; TTL is only for cleanup
+	ctx           context.Context                       // Cancelled when the manager is shutting down
+	cancel        context.CancelFunc                    // Cancels ctx
 }
 
 // hubLike defines the interface requirements for the hub dependency.
@@ -60,10 +63,13 @@ type hubLike interface {
 // NewSystemManager creates a new SystemManager instance with the provided hub.
 // The hub must implement the hubLike interface to provide database and alert functionality.
 func NewSystemManager(hub hubLike) *SystemManager {
+	ctx, cancel := context.WithCancel(context.Background())
 	return &SystemManager{
 		systems:       store.New(map[string]*System{}),
 		hub:           hub,
 		smartFetchMap: expirymap.New[smartFetchState](time.Hour),
+		ctx:           ctx,
+		cancel:        cancel,
 	}
 }
 
@@ -103,7 +109,13 @@ func (sm *SystemManager) Initialize() error {
 		sleepTime := time.Duration(delta) * time.Millisecond
 
 		for _, system := range systems {
-			time.Sleep(sleepTime)
+			select {
+			case <-sm.ctx.Done():
+				// Abort if the manager has been shut down (e.g. test cleanup)
+				// to avoid starting updater goroutines against a torn-down hub. See #1950.
+				return
+			case <-time.After(sleepTime):
+			}
 			_ = sm.AddSystem(system)
 		}
 	}()
@@ -238,6 +250,11 @@ func (sm *SystemManager) onRecordAfterDeleteSuccess(e *core.RecordEvent) error {
 // It validates required fields, initializes the system context, and starts the update goroutine.
 // Returns error if a system with the same ID already exists.
 func (sm *SystemManager) AddSystem(sys *System) error {
+	if sm.ctx.Err() != nil {
+		// Manager is shutting down; do not start new updater goroutines
+		// against a hub that may be torn down. See #1950.
+		return sm.ctx.Err()
+	}
 	if sm.systems.Has(sys.Id) {
 		return errSystemExists
 	}
--- a/internal/hub/systems/system_test.go
+++ b/internal/hub/systems/system_test.go
@@ -3,9 +3,11 @@
 package systems
 
 import (
+	"context"
 	"testing"
 
 	"github.com/henrygd/beszel/internal/entities/system"
+	"github.com/pocketbase/pocketbase/tools/store"
 )
 
 func TestCombinedData_MigrateDeprecatedFields(t *testing.T) {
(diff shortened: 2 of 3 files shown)


BootstrapperSBL:
Heads up: the VulnCheck failure here is pre-existing on `main` — `golang.org/x/image@v0.38.0` has two advisories (GO-2026-4961 / 4962) that are fixed in 0.39.0. Same failure is blocking #1946 and #1932. Once you bump the dep on main this PR should go green.

henrygd (commented):
Thanks, I'm checking this out now and still getting failed tests. Possibly due to something I added in the past few days. I'll play around w/ it a bit and add on to what you've done here.

PR #1957, opened 2026-04-27
Clamp Linux battery capacity percentage

VACInc:
## Summary
- clamp Linux sysfs battery capacity values to the valid 0-100 range before reporting them
- add a regression test for a kernel/sysfs capacity value above 100

## Context
Some Linux systems can briefly report battery `capacity` values above 100 via `/sys/class/power_supply/*/capacity`. Beszel currently forwards that value directly, which can surface impossible percentages like 105% in integrations.

## Tests
- `go test -tags testing ./agent/battery`
- `go test ./agent/...`


henrygd (approved):
Looks good, much appreciated!

Issue #1011, opened 2025-07-29
[Bug]: Unable to log in on Android devices when using password manager autofill


yurividal:
### Component

Hub

### Description

Beszel works fine on my computer. Any time I try to login on my phone, I get the login page, but pressing the login button does nothing. The page just sits there and never logs me in. I tried multiple browsers including Samsung internet browser and Brave.
Even setting to show desktop version doesn't help.
My beszel instance is behind nginx, by the way.

### Expected Behavior

I should be able to login

### Steps to Reproduce

Open https://beszel.mydomain.com on a mobile browser.
Try to login

### Category

Authentication

### Affected Metrics

Other

### OS / Architecture

Ubuntu 22.04

### Beszel version

Latest

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

henrygd:
Is this for password login, oauth login, or both? Stock Android or spin off?

I can't reproduce on my Pixel 6a or BrowserStack.

yurividal:
This is normal login, o Samsung OneUI.
I think the issue is actually on how beszel is being proxied by nginx proxy manager. If I open by IP and port of the container, it works.
Is there a specific nginx example or template for beszel?
I tried enabling web sockets, but that also didn't work.

henrygd:
I'd recommend trying the nginx config in the PocketBase docs: https://pocketbase.io/docs/going-to-production/#using-reverse-proxy

```nginx
server {
    listen 80;
    server_name example.com;
    client_max_body_size 10M;

    location / {
        # check http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive
        proxy_set_header Connection '';
        proxy_http_version 1.1;
        proxy_read_timeout 360s;

        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;

        # enable if you are serving under a subpath location
        # rewrite /yourSubpath/(.*) /$1  break;

        proxy_pass http://127.0.0.1:8090;
    }
}
```

If you open your browser's devtools network tab or console do you see any errors after clicking the button?

Tailscale-VPS:
[...]

PR #1970, opened 2026-05-03
Fixed the password manager login issue

Damilola-GZY:
## 📃 Description

The login form had two problems that combined to make the button appear unresponsive when pasting credentials from a password manager.

## 📖 Documentation

### ➕ Added

1. Moved `setIsLoading(true)` to after validation passes
The button now only enters loading state when the form is valid and an actual API call is about to be made. If validation fails for any reason, the button stays normal and an error is shown immediately.
2. Filter out the honeypot field from errors, show a fallback message
The error loop now skips the website field entirely. If the only thing that failed was the honeypot (meaning no visible field has an error to display), a generic message is shown on the email field — "Please check your details and try again" — so the user always gets feedback.
3. Clean up the error handling code
Removed the `console.log debug` statement
Properly typed the errors object `(Record<string, string>)`
Replaced `issue.path[0].key` with the `null-safe issue.path?.[0].`
Removed the `@ts-expect-error` suppression

### 🔧 Fixed

- Passport Manager login issue

> Completed with assistance of Replit

### DIFF
--- a/internal/site/src/components/charts/line-chart.tsx
+++ b/internal/site/src/components/charts/line-chart.tsx
@@ -104,7 +104,7 @@ export default function LineChartDefault({
 					isAnimationActive={false}
 					// stackId={dataPoint.stackId}
 					order={dataPoint.order || i}
-					// activeDot={dataPoint.activeDot ?? true}
+					activeDot={dataPoint.activeDot ?? true}
 				/>
 			)
 		})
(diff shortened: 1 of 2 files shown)


aqsede:
Does this fix #1011 issue with Bitwarden

Damilola-GZY:
Yes, I've fixed it and tested it.

scr4bble:
Man, you cannot simply commit such a mess. You are not supposed to change formatting of the whole file...

Damilola-GZY:
> Man, you cannot simply commit such a mess. You are not supposed to change formatting of the whole file...

Thanks man
Will edit

PR #1973, opened 2026-05-03
fix(hub): check X-Real-IP header to resolve agent's real source IP

### LINKED ISSUE #1965: [Bug]: `systems.host` stores Docker bridge MASQ IP instead of agent's real source IP
### Component
Hub

### Description

When the hub runs as a Docker container behind the default bridge network and an agent connects over WebSocket from another host, the `host` column of the `systems` collection in `data.db` is populated with the **container's MASQ source IP** (e.g. `172.21.0.1`) rather than the agent's actual source IP (e.g. `10.8.0.4`).

This is functionally harmless — the agent stays connected, metrics flow correctly — but it's **misleading** when reviewing the hub's database in a multi-host setup. The `host` column reads as if it identifies where the agent is, but it actually identifies which Docker bridge the hub itself is sitting behind.

In my setup all three agents (one local, two over WireGuard at `10.8.0.4` and `10.8.0.5`) are recorded with `172.21.0.1` as their `host` value. From the database alone there's no way to tell them apart by network location.

### Steps to Reproduce

1. Run hub as Docker container on host A (default bridge network).
2. Run an agent on host B (real IP `10.8.0.4`).
3. Agent connects via WebSocket to `https://hub.example/`.
4. Inspect:
   ```sh
   docker run --rm -v <hub-data-volume>:/d alpine sh -c \
     "apk add --no-cache sqlite >/dev/null && sqlite3 /d/data.db 'SELECT name, host FROM systems;'"
   ```
5. Observed: `host` column shows `172.21.0.1` (Docker bridge gateway).
   Expected: `10.8.0.4` (real source IP of the agent's WebSocket connection).

### Suggested Fix

When recording / updating the `host` value on a WebSocket connection, prefer in this order:
1. `X-Real-IP` request header (when present — typical behind reverse proxies),
2. `X-Forwarded-For` first hop (also typical behind reverse proxies),
3. The WebSocket upgrade request's remote peer address (correct when no proxy in front),
4. Fall back to the current behaviour only if none of the above is sensible.

For setups behind Traefik / nginx the first two are already populated automatically.

### Environment
- Hub: `henrygd/beszel:0.18.7` (Docker, default bridge)
- Agents: `henrygd/beszel-agent:0.18.7` (`network_mode: host`, WebSocket mode)
- Behind Traefik reverse proxy

svenvg93:
## 📃 Description

Add `X-Real-IP` header support to `getRealIP` so the `systems.host` column stores the agent's real source IP when the hub runs behind nginx or Traefik in a Docker bridge network setup.

closes #1965

### 🔧 Fixed

- `systems.host` storing Docker bridge MASQ IP (`172.x.x.x`) instead of the agent's real source IP when the hub runs as a Docker container behind a reverse proxy that sets `X-Real-IP` (e.g. nginx, Traefik)


henrygd (approved):
Thanks Sven!

PR #1978, opened 2026-05-04
fix(agent): prevent memory/swap uint64 underflow causing unrealistic usage spikes

### LINKED ISSUE #1975: [Bug]: Unexpected Memory Usage Spike Showing Unrealistic Values (Millions of TB)
### Welcome!

### Component

Hub

### Problem Description

Hi,

I’m experiencing a strange spike in memory usage on Beszel.

At one point, the “Used” memory suddenly jumped to an extremely large value (millions of TB), while cache and buffers remained normal (only a few GB). After that, it immediately returned to normal.

### Expected Behavior

I’ve verified the actual server memory using `free -h`, and everything looks normal, so this seems like a monitoring glitch.

### Steps to Reproduce

Has anyone experienced similar issues with Beszel?
Could this be caused by:
- incorrect parsing of /proc/meminfo?
- unit conversion bug?
- or a polling/data collection issue?

Any insights would be appreciated. Thanks!

Image

### OS / Architecture

linux

### Beszel version

0.12.1

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Fix uint64 integer underflow in memory and swap stat collection that could cause anomalous spikes (millions of TB) in reported memory usage. When `/proc/meminfo` values are transiently inconsistent, gopsutil can return a `Used` value greater than `Total`, or `Shared` can exceed `Cached + Buffers`, both of which cause unsigned integer wraparound to near-`MaxUint64`.

fixes #1975 

### 🔧 Fixed

- Skip memory sample entirely when `v.Used > v.Total` to prevent transient `/proc/meminfo` inconsistencies from producing unrealistic memory usage spikes
- Fix uint64 underflow in `cacheBuff` calculation (`Cached + Buffers - Shared`) which could wrap to ~18 exabytes when `Shared > Cached + Buffers`; also removed the dead `<= 0` guard that could never trigger on a uint64
- Fix uint64 underflow in `SwapUsed` calculation by guarding the subtraction before performing it


SOCSolnet:
Hi
 
I’ve already updated to the latest version, but I’m still experiencing the same problem. The memory usage still occasionally shows unrealistic values (extremely large TB numbers), similar to what was described before.

Here are my details:

Environment: LXC
Beszel version: 0.18.7

It seems like the issue might still occur in certain environments.

Please let me know if you need additional logs or configuration details to help investigate further.

Thanks

image

svenvg93:
> Hi
> 
> I’ve already updated to the latest version, but I’m still experiencing the same problem. The memory usage still occasionally shows unrealistic values (extremely large TB numbers), similar to what was described before.

This change has not been merged and released yet :) the latest version still had this bug in it.

henrygd:
Thanks Sven, I added a fix here and put you as co-author: https://github.com/henrygd/beszel/commit/d40372842bf12ea9bf11d2b093c7912ecaf075a8

PR #1999, opened 2026-05-13
feat(hub): add system journal logs tab

### LINKED ISSUE #37: [ feature request / support ] Add syslog monitoring
I would like to suggest adding monitoring of syslog on system and provide a way to explore them within pocketbase. It would make sense for a log monitoring hub to contain the most important logs for the agents in one place. We can have it either for individual system or combine them.

### LINKED ISSUE #1461: [Feature]: Ability to view systemd logs
### Component

Hub

It would be great if you could see logs for systemd services (e.g. journalctl), similar to how you can see docker logs. Beszel is a lovely piece of software, thanks for all of the hard work on it!

### Motivation / Use Case

Makes it easier to debug systemd service issues

Same way that you can view docker logs

### Screenshots

_No response_

### Category

Metrics

### Affected Metrics

Other

svenvg93:
## 📃 Description

Adds on-demand system journal log viewing to the agent detail page, similar to how container logs work. A new **Logs** tab appears on systems running systemd, showing the last 200 journal lines fetched live from the agent via `journalctl`. Logs can be filtered by service name, and the Services tab now has a **View Logs** button that jumps directly to the Logs tab pre-filtered for that service.

Closes #1461, Closes #37.

### ➕ Added

- New `GetSystemLogs` WebSocket action — agent runs `journalctl --no-pager --output=short-iso -n200 [-u <service>]` and returns the output
- `GET /api/beszel/system/logs?system=<id>&service=<name>` hub API endpoint (service filter is optional)
- **Logs** tab on the system detail page (tabbed layout, only shown when systemd is available)
  - Service name filter input — type a unit name and press Enter or click Refresh
  - Search field for searching
  - Syntax-highlighted output via shiki (`github-dark-dimmed` theme, `log` language)
  - Fullscreen dialog for easier reading
- **View Logs** button in the systemd service detail sheet — switches to the Logs tab and pre-fetches logs for that service

## 📷 Screenshots

Scherm­afbeelding 2026-05-13 om 18 49 40


cquike:
This might be out of the scope of this PR, but I would find it very useful if the logs are somehow persistent rather than only on demand. My use case is a remote system that crashed. In that case it would be very useful if beszel shows the last journal entries fetched.
Also, the last 200 lines would have a mix of info from different priorities. Maybe messages with priority 1 could be somehow kept so that they are not erased by the last non-critical 200 lines?

PR #2001, opened 2026-05-13
feat(docker): add healthcheck to hub docker-compose example

### LINKED ISSUE #1963: [Docs]: Add HEALTHCHECK example using built-in `beszel health` to compose
### Component
Documentation / Hub

### Description
The official `docker-compose.yml` example for the hub does not include a `healthcheck` directive. Users who try to add one via the typical `wget`/`curl` pattern fail because the hub image is distroless (no shell, no `wget`/`curl` binary in the image).

The binary already exposes a `health` subcommand that does an HTTP self-check — perfect for a Docker `HEALTHCHECK` — but it's not mentioned in the deployment docs or compose example as far as I can find.

### Suggested Improvement

Add the following to the example compose at `supplemental/docker/hub/docker-compose.yml`:

```yaml
services:
  beszel:
    # ... existing config ...
    healthcheck:
      test: ["CMD", "/beszel", "health", "--url", "http://127.0.0.1:8090"]
      interval: 30s
      timeout: 5s
      retries: 3
      start_period: 10s
```

Verified working on `henrygd/beszel:0.18.7`:

```console
$ docker exec beszel /beszel health --url http://127.0.0.1:8090
ok
$ echo $?
0
```

### Why this matters

Without this hint, self-hosters running the distroless hub image typically end up with one of:

- silently no `healthcheck` at all (no orchestrator-level liveness signal, container shown as "running" forever even when wedged),
- time spent discovering the workaround in source / issues,
- or external probes (Traefik upstream, blackbox_exporter, Uptime Kuma) that don't surface in `docker ps` / `docker compose ps` and don't trigger Docker's restart policy.

A small addition to the example would prevent all three cases for new users.

Happy to send a PR if it's welcome.

### Environment

- Beszel Hub: `henrygd/beszel:0.18.7`
- Docker: 28.x on Ubuntu 22.04
- Behind Traefik reverse proxy (VPN-only)

Iriome-Santana:
## What

Adds a `healthcheck` directive to `supplemental/docker/hub/docker-compose.yml`.

## Why

The hub image is distroless — no shell, no `wget`, no `curl` — so the typical healthcheck patterns fail silently. The binary already exposes a `health` subcommand that performs an HTTP self-check, making it a natural fit for `HEALTHCHECK`.

Without this, containers running the distroless hub image show as "running" indefinitely even when wedged, with no orchestrator-level liveness signal and no automatic restart policy trigger.

## Verification

Tested on `henrygd/beszel:0.18.7`:

```bash
$ docker exec beszel /beszel health --url http://127.0.0.1:8090
ok
$ echo $?
0
```

Closes #1963

### DIFF
--- a/supplemental/docker/hub/docker-compose.yml
+++ b/supplemental/docker/hub/docker-compose.yml
@@ -7,3 +7,9 @@ services:
       - '8090:8090'
     volumes:
       - ./beszel_data:/beszel_data
+    healthcheck:
+      test: ["CMD", "/beszel", "health", "--url", "http://127.0.0.1:8090"]
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
--- a/supplemental/docker/same-system/docker-compose.yml
+++ b/supplemental/docker/same-system/docker-compose.yml
@@ -9,6 +9,12 @@ services:
       - ./beszel_data:/beszel_data
     extra_hosts:
       - 'host.docker.internal:host-gateway'
+    healthcheck:
+      test: ["CMD", "/beszel", "health", "--url", "http://127.0.0.1:8090"]
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
 
   beszel-agent:
     image: 'henrygd/beszel-agent' #Add -nvidia for nvidia gpus


Iriome-Santana:
Also added the healthcheck to supplemental/docker/same-system/docker-compose.yml since it includes the hub service as well and has the same issue

henrygd:
Thanks!

PR #2004, opened 2026-05-14
feat(hub): decouple systems from user for role-based access control

### LINKED ISSUE #1856: [Feature]: Systems as Standalone Entities (Decouple from Users)
### Welcome!

### Component

Hub

### Description

Systems are currently stored with a users relation field, meaning a system cannot exist without being attached to at least one user. This proposes removing that field and introducing a separate `system_access` join collection to manage user↔system relationships independently.

### Motivation / Use Case

In multi-user deployments, systems represent infrastructure — not personal resources. The current model forces every system to be "owned" by a user, which creates friction in several real scenarios:

- An admin wants to pre-provision systems before assigning them to anyone
- A user is removed from the hub, putting any systems they solely owned in a broken state

Decoupling systems from users lets systems exist as first-class infrastructure objects, with access granted separately and explicitly.

svenvg93:
## 📃 Description

Decouples systems from per-user ownership. Previously each system record had a `users` relation field controlling who could see it. This PR removes that field and adopts a role-based model: all authenticated users can view all systems, only admins can create/edit/delete them. A new in-app user management page replaces the need to use the PocketBase admin UI for user administration.

closes #1856

### ➕ Added

- **User management page** in Settings (admin only) — create, edit role, reset password, and delete users without leaving the app
- **Role-based access control** — three roles: `admin` (full control), `user` (read-only), `readonly` (read-only, no tokens/fingerprints page)
- Migration `2_user_email_visibility` — sets `emailVisibility = true` for all existing users so admins can list them

### ✏️ Changed

- All authenticated users can now view all systems and system-scoped data (containers, stats, SMART, etc.)
- Only `admin` role users can create, update, delete, or pause systems
- Universal token endpoint (`GET /api/beszel/universal-token`) now requires admin role
- Tokens & Fingerprints settings page is now admin-only (previously excluded only `readonly`)
- SMART alerts are now sent to all users with notification settings configured (previously only users with explicit system access)
- `config.yml` `users` field per system removed — systems are globally visible, no per-system user assignment needed
- New users are automatically marked as `verified` and `emailVisibility = true` on creation

### 🔧 Fixed

- Mobile settings sidebar now correctly hides the Tokens & Fingerprints item for `readonly` users (desktop already did this)
- `requireAdminRole` middleware now passes superusers through to endpoint handlers so they receive their specific error messages rather than a generic 403
- `USER_CREATION=true` (OAuth2 mode) create rule now correctly takes precedence over the admin create rule on the users collection

### 🗑️ Removed

- `users` relation field from `systems` collection (migration `1_system_access`)
- `system_access` join collection and all associated code (agent connect, config sync, alert targeting, frontend)
- `SHARE_ALL_SYSTEMS` environment variable — all systems are now shared by default
- `config.yml` per-system `users` list — no longer needed

## 📷 Screenshots

Scherm­afbeelding 2026-05-14 om 16 28 20


svenvg93:
@henrygd Before I fix the conflicts, can I ask you for a quick review and thoughts on this?

henrygd:
Thanks Sven, I love the user management page, but I don't know that we can decouple systems from users at this point.

Until now we've had it so users have their own systems separated from other users. If we merge everything together, all of a sudden everyone will see others' personal systems. Right now this can be opted into with `SHARE_ALL_SYSTEMS=true`. If that isn't set, we have to assume that users may have private systems they don't want shared.

To me, a better solution is adding teams/groups. Admins can create a team and add users to the team. Then users can share systems to the team for everyone in the team to share. What do you think of this approach?

IIRC you may have had a PR for something similar at one point that was closed. If so, sorry about that. I have a bad habit when I see big diffs of tabling for later and never getting back around to it, since there's always new stuff coming in.

svenvg93:
Thanks Henry, that makes sense — I don't want to break the privacy expectation for existing multi-user setups.

I think teams/groups is actually the same thing as decoupling, just done properly. Instead of "system belongs to these N users" it becomes "system belongs to a team, users belong to teams." The reason I'd still like to get away from direct user assignment is operational: as soon as you have a handful of users and a lot of systems, onboarding someone means editing every system's user list, and each system's access drifts on its own. Teams fix that.

Rough model I have in mind:

- `teams` collection, `team_members` join (team ↔ user), and a `system` ↔ `team` relation
- **Read access:** you can see a system if you share a team with it, or you're an admin
- Keep the role model from this PR for **writes**: only admins create/edit/delete systems
- `SHARE_ALL_SYSTEMS=true` maps cleanly to "one team, everyone in it"

For the upgrade path, the migration would create a personal team per existing user and assign each system to teams based on its current `users` / `system_access` rows, so nobody's visibility changes on upgrade. Admins can consolidate into real teams afterward.

On the diff size — fair point, and I'd rather not have this stall. Happy to split it: first PR is just the user management page rebased onto the current access model (no behavior change), then a second PR for the teams model. Does that work for you?

svenvg93:
I will revisit this later on, with a better plan.

PR #2005, opened 2026-05-14
feat(hub): decouple from users and add global alerts

### LINKED ISSUE #897: [Feature]: Add system wide configuration for alerts and views

I think it would be nice to be able to set some configuration system wide (by that I mean for all users) especially alerts, views and webhooks.

As a member of the admin group I would like to be able to create, update, delete webhooks create by other admins. Those webhooks are meant to be created once but used by many members (eg sending message to mattermost) but it is nice to be able to create personal webhooks too.

In the continuity of this I would like to be able to have alerts globally defined so every user can see the `active alerts` panel with the threshold I chose as an admin.

In the same idea I would like to be able as an admin to create a default view (hiding GPU/Temp for example) so I don't need to do it for every new user (every user can edit his view after login obviously)

Let me know if more explanation is needed

### LINKED ISSUE #1914: [Feature]: Alerts: replace 'All Systems' functionality with standalone global alerts
### Welcome!

### Component

Hub

### 📌 Description
When configuring alerts using the **"All Systems"** option, the alerts do not trigger as expected.

Alerts configured individually per system work correctly, but when selecting **"All Systems"**, no alerts are fired.

---

### ⚙️ Environment
- Beszel version: v0.18.7
- Installation type: Binary (manual install with systemd)
- OS: Ubuntu (server)
- Architecture: amd64

---

### 🧩 Steps to Reproduce

1. Go to **Alerts configuration**
2. Select **"All Systems"**
3. Enable any alert (CPU, Memory, Disk, etc.)
4. Set a threshold (e.g., CPU > 10%)
5. Apply/save configuration
6. Generate load to exceed the threshold

---

### ❌ Expected Behavior
Alerts should trigger for all monitored systems when thresholds are exceeded.

---

### ✅ Actual Behavior
No alerts are triggered when using **"All Systems"**.

However:
- Alerts work correctly when configured per individual system

---

### 📷 Screenshot
(attach screenshot if needed)

---

### 🔍 Additional Notes

- No visible errors in UI
- No obvious logs indicating failure
- Agents are connected and reporting metrics correctly
- Issue seems specific to global alert scope

---

### 💡 Possible Cause (guess)
It may be related to how global alert rules are applied or propagated to systems.

---

### 🚀 Impact
This affects centralized monitoring, requiring manual alert configuration per system.

---

### 🙏 Request
Please verify if:
- This is a bug
- Or if additional configuration is required for global alerts

---

Thanks for the great project 👍

### Expected Behavior

Alerts work correctly when configured per individual system

### Steps to Reproduce

When configuring alerts using the "All Systems" option, the alerts do not trigger as expected.

Alerts configured individually per system work correctly, but when selecting "All Systems", no alerts are fired.

### OS / Architecture

amd64

### Beszel version

0.18.7

### Installation method

Binary

### Configuration

```yaml
⚙️ Environment
Beszel version: v0.18.7
Installation type: Binary (manual install with systemd)
OS: Ubuntu (server)
Architecture: amd64
```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Decouples alerts from users and adds global alerts. Alerts are now a single shared record per system/type instead of one per user. Admins can configure global alert defaults that automatically sync to all systems, with per-system exclusion support.

Closes #1914
Closes #897

### ➕ Added

- Global Alerts settings page (admin-only) — alert defaults that auto-apply to all systems
- "Global" badge with tooltip on per-system alerts managed by a global alert
- Per-system exclusions per global alert type
- Users can enable notifications if needed and subscribe to specific systems. 

### ✏️ Changed

- Alerts are now shared across users — one record per system/type
- "Copy from" skips globally managed alerts on the target system

### 🔧 Fixed

- `excluded_systems` changed from relation to JSON field to avoid validation errors when an admin is not a member of all systems

### 🗑️ Removed

- `user` field from `alerts` and `alerts_history` collections
- "All Systems" tab from the alerts sheet

## 📷 Screenshots

Scherm­afbeelding 2026-05-14 om 22 11 17


svenvg93:
@henrygd Before I fix the conflicts, can I ask you for a quick review and thoughts on this?

henrygd:
Apologies for the wait. I love the concept of global alerts, but I'm not sure I like decoupling the alerts from users. I think it would be nice if they still have the ability to customize the thresholds that trigger their alerts.

Off the top of my head, I think the best solution would be:

1. Keep the global alerts page in settings with ability to exclude systems. I really like what you did with this. But it would still apply at user level rather than globally for all users.
2. In the systems table alert sheet, replace the 'All Systems' tab with a 'Global' option. Like the global alerts settings page but without the ability to exclude systems (to keep it simple).

Similar to what you described but give more flexibility to individual users to customize their own global alerts. 

But I'm open to what you did and will test it out if the conflicts are fixed. I also have the option to "Fix with Copilot," which I've never used, but I can try it out here if you want. Also, let me know if I misunderstood anything!

svenvg93:
Thanks for taking a look!

I hear you on threshold customization. In most monitoring tools (Grafana, Alertmanager, Datadog, Zabbix, PRTG) the alert rule / threshold is a shared, centrally-managed thing,  one source of truth per system/metric. What's per-user is the *notification routing*: which alerts reach me, on which channel, during which hours. Per-user thresholds on the same host tend to cause "why did you get paged and I didn't" confusion and config drift, and there's no single place to see what a system's alerting actually is.

henrygd:
What would the migration path from the current setup look like? I assume users would lose their current alerts. Would all alerts be gone after upgrade until an admin can set up global alerts?

svenvg93:
> What would the migration path from the current setup look like? I assume users would lose their current alerts. Would all alerts be gone after upgrade until an admin can set up global alerts?

Good questions, here's the migration path:

Alerts themselves aren't lost.The migration `1_alerts_remove_user` dedupes the existing `alerts` rows to one per `(system, name)`, keeping the most recently updated one, then drops the `user` column. Those rows keep firing immediately after upgrade — nothing needs an admin to act. `global_alerts` starts empty and is purely an optional layer on top for pushing new defaults.

The one real loss: if two users had, say, a CPU alert on the same system with different thresholds, only one value survives (last updated wins) and becomes the shared value.

Notifications were the bigger gap, since the alert no longer carries a user, `SendAlert` now fans out over `user_settings` and only notifies users who have `notificationsEnabled` set. Existing users don't have that flag, so without a migration everyone would go silent after upgrade until they re-opt-in.

So I added a backfill to the same migration: before the `user` column is dropped, it reads each user's `(user, system)` alert pairs and, for every user who had alerts, sets `notificationsEnabled = true` and subscribes them to exactly the systems they had alerts on. Their email/webhook channels are untouched. Net effect: same people get notified for the same systems as before the upgrade.

I'll rebase and clear the conflicts so you can test it end to end.

Let me know what you think, happy to make any changes you want! And if you don't feel like this is the right approach  feel free to close the PR, no hard feelings 😄

svenvg93:
Closing this one in favour of simpler version for reviewing.

PR #2008, opened 2026-05-15
feat(agent): add process state count monitoring

svenvg93:
## 📃 Description

Adds process state count monitoring to the agent. Aggregate counts for running, sleeping, idle, stopped, and zombie processes are collected each stats interval using the existing `gopsutil/v4` library and displayed as a line chart on the system detail page.

### ➕ Added

- Agent now collects process state counts (running, sleeping, idle, stopped, zombie) each stats interval
- New **Processes** line chart on the system detail page showing process counts by state over time

### ✏️ Changed

- `SystemStats` struct extended with `Processes [6]uint16` field (`ps` in JSON/CBOR, key 36)

## 📷 Screenshots

Scherm­afbeelding 2026-05-15 om 18 45 48


henrygd:
I made some updates on this, but I'm not going to merge yet because I'm worried about performance / allocations on systems with lots of processes, considering we need to iterate over every process to get its state.

We would probably need to set an upper limit. Like skip the functionality if the system has more than 2,000 processes or something.

svenvg93:
> I made some updates on this, but I'm not going to merge yet because I'm worried about performance / allocations on systems with lots of processes, considering we need to iterate over every process to get its state.
> 
> We would probably need to set an upper limit. Like skip the functionality if the system has more than 2,000 processes or something.

Makes sense, I understand the concern about iterating every process on high-process-count systems. I added a limit of 2,000 processes for now, so the state scan is skipped entirely above that threshold rather than paying the cost every interval.

PR #2011, opened 2026-05-16
fix(agent): validate HUB_URL host to prevent misleading "lookup api" DNS error

### LINKED ISSUE #1958: [Bug]: Agent: hostname-based HUB_URL surfaces literal "api" in DNS-lookup error, not the configured hostname
## Summary

When the agent is configured with a hostname-based `HUB_URL` (e.g. `http://hub.example.lan:8090`), the WebSocket control channel fails with a DNS lookup error referencing the literal hostname `api`, *not* the configured hostname:

```
WARN WebSocket connection failed err="dial tcp: lookup api on 1.1.1.1:53: no such host"
```

Switching `HUB_URL` to the same hub's literal IP address (e.g. `http://192.0.2.10:8090`) makes the connection succeed immediately. Same hub, same agent binary, same network path — only the form of the address differs.

## Why this matters

For deployments where the hub lives on a hostname that only resolves through an internal/split-horizon DNS resolver, this makes hostname-based `HUB_URL` effectively unusable when the agent host hasn't been configured to use that resolver. Operators have to hardcode the hub's IP in every agent's environment, which is fragile and obscures intent.

A clear error message naming the actual hostname the agent tried to resolve would have made this trivial to diagnose. The misleading `lookup api` masked the real cause for a long time.

## Environment

- Beszel agent: **0.18.7**
- Beszel hub: **0.18.7** (`henrygd/beszel:latest` Docker image, `network_mode: host`)
- Connection mode: universal-token self-registration (`TOKEN` set on the agent)
- Install method: official `https://get.beszel.dev` script with `-k <pubkey> -p <port> -t <token> --auto-update false`

## Steps to reproduce

1. Run the hub on host `H` reachable by hostname `myhub.lan` and IP `192.0.2.10:8090`. Make sure `myhub.lan` is **not** resolvable through whichever DNS resolvers the agent host has configured in `/etc/resolv.conf` (i.e. it only resolves via an internal resolver the agent host does not consult).
2. On a clean agent host, install the agent with the official installer:
   ```
   curl -fsSL https://get.beszel.dev | bash -s -- -k '<HUB_PUBKEY>' -p 45876 -t '<TOKEN>' --auto-update false
   ```
3. Configure `HUB_URL` via systemd drop-in:
   ```
   # /etc/systemd/system/beszel-agent.service.d/override.conf
   [Service]
   Environment="HUB_URL=http://myhub.lan:8090"
   Environment="LISTEN=<agent-ip>:45876"
   ```
   Then `systemctl daemon-reload && systemctl restart beszel-agent`.
4. Verify the env is applied: `systemctl show beszel-agent | grep '^Environment='` should include `HUB_URL=http://myhub.lan:8090`.
5. Observe `journalctl -u beszel-agent -f`.

## Expected behavior

The agent attempts to resolve `myhub.lan` and either:
- Connects (if the resolver can reach the relevant DNS), or
- Fails with an error message that includes the actual hostname it tried to resolve (`myhub.lan`), making the cause obvious.

## Actual behavior

The agent logs a continuous stream (every ~10 s) of:

```
2026/04/29 02:42:30 WARN WebSocket connection failed err="dial tcp: lookup api on 8.8.8.8:53: no such host"
2026/04/29 02:42:40 WARN WebSocket connection failed err="dial tcp: lookup api on 1.1.1.1:53: no such host"
```

The lookup target is the literal string `api` — *not* the configured hostname `myhub.lan`. The configured hostname does not appear anywhere in the agent's output. The SSH server starts normally and continues to listen, but no metrics ever reach the hub.

## Workaround

Replace the hostname in `HUB_URL` with the hub's literal IP:

```
Environment="HUB_URL=http://192.0.2.10:8090"
```

After `systemctl restart beszel-agent`, the WebSocket connects on the next attempt:

```
INFO WebSocket connected host=192.0.2.10:8090
```

## Suspected root causes (not verified — reading symptoms only)

Two hypotheses that fit the observed behavior:

1. **URL parsing failure with silent fallback.** If the agent's URL parsing rejects hostname-only URLs and silently falls back to a hardcoded default that ends up as just `api`, this would explain why the configured hostname never appears in the error. A non-fatal parse failure should at minimum log the parsing error.
2. **Hostname extracted from a sub-path or sub-domain default.** The literal `api` is suspicious — it may be a default subdomain or path segment used by the agent's WebSocket code path. Some templating like `wss://api.${HUB_HOST}/...` could degenerate to just `api` if `HUB_HOST` ends up empty.

Happy to provide additional logs or run further reproduction tests if useful.

## Suggested fixes (in priority order)

1. **Surface the actual hostname in the error message.** Whatever the resolution path, the error should include the hostname the agent tried to resolve. `dial tcp: lookup myhub.lan on 8.8.8.8:53: no such host` would have been self-diagnostic.
2. **Validate `HUB_URL` at startup** and either:
   - Fail loudly with a clear error if the URL can't be parsed (rather than silently falling back).
   - Resolve the hostname once at startup and log the outcome; warn on resolution failure.
3. **Document the gotcha** in the agent README/docs: clarify that `HUB_URL` must point at a hostname resolvable by `/etc/resolv.conf` on the agent host (or use an IP).

## Additional notes

- Reproduces on both glibc and musl agent binaries.
- Setting `LISTEN` to a non-default IP works correctly in parallel — the listener side is fine.
- Universal-token self-registration via `TOKEN` works once the WebSocket actually connects (using IP form), so the auth path itself is healthy.

Thanks for the project. Happy to help further if you need additional repro details.

svenvg93:
## 📃 Description

When `HUB_URL` is set to an empty string (e.g. via a systemd drop-in that doesn't take effect), `url.Parse("")` succeeds silently and the agent constructs the WebSocket address `ws://api/beszel/agent-connect`, causing a confusing `dial tcp: lookup api` DNS error with no indication of what went wrong.

Fixes #1958 

### 🔧 Fixed

- Agent now validates that `HUB_URL` contains a host after parsing and fails immediately with a clear error message instead of attempting to connect to `api` as a hostname


henrygd (approved):
Thanks! It causes a test to fail based on error message, but I'll patch that up afterwards.

PR #2014, opened 2026-05-17
feat(hub): single node mode and alerts button on system detail page

### LINKED ISSUE #1379: [Feature]: Single node dashboard mode
### Component

Hub

First of all thanks for the great software, really happy with it! Allowed me to remove a much more work/custom Telegraf/Influx/Grafana stack for my server at home 👍 

Would it be possible to have an option for a single node mode? I only have a single server at home, on which I run both the hub and the agent (same as https://github.com/henrygd/beszel/issues/73).
Because it's just a single server the initial "All Systems" overview doesn't really make sense/is kind of redundant. I just want to go the the dashboard of that one server :)
So it would be nice if the "All Systems" overview could be disabled/if it would be possible to just have the one node as the main page.

### Motivation / Use Case

Single server at home (which I expect will apply to quiet a lot of people) and want to monitor that. Don't need the "overhead" of a system for managing multiple servers.

Probably a setting/feature toggle?
Not entirely sure how one would determine the system to display. I guess if there's just one it's easy, just pick that one ;) But not sure if it's that simple.

One thing I noticed is that it seems like it isn't possible to configure alerts on the page for a specific system, the alert button seems to only be present in the table on the All Systems at the moment.

### Screenshots

_No response_

### Category

Settings & Configuration

### Affected Metrics

CPU

### LINKED ISSUE #1927: [Feature]: Add "edit alerts" button to host detail page
### Welcome!

### Component

Hub

### Description

Feature request is to add the "edit alert" bell button, to the host detail page. 

This is so that you don't have to return to the overview list, to edit the host alerts.

Image

Image

### Motivation / Use Case

Oftentimes I want to review the stats on a host, and tailor the alert at the same time. 

Going back to the overview page means I have to find the host again in the list to edit it. 

Currently I must use two windows or browser tabs to accomplish this. 

_Bonus:_ Perhaps the hamburger/ellipsis menu contents (again from list view) could also eventually be added to the host detail page, under an "Actions" button or something. Just a thought to improve usability.

svenvg93:
## 📃 Description

Addresses two UX improvements for single-server setups and the system detail page.

Closes #1379 — adds an opt-in **single node mode** that automatically redirects to the system dashboard when only one system is configured, skipping the "All Systems" overview.

Closes #1927 — adds the **alerts button** to the system detail page so alerts can be edited without returning to the overview.

### ➕ Added

- **Single node mode** setting in Settings → General: when enabled and exactly one system is configured, navigating to the home route redirects directly to that system's dashboard
- Alert bell button added to the system detail page info bar, matching the behaviour already present in the systems table

### ✏️ Changed

- Single node mode toggle is disabled and greyed out when more than one system exists
- Single node mode is automatically disabled when a second system is added

## 📷 Screenshots

Scherm­afbeelding 2026-05-17 om 09 25 19
Scherm­afbeelding 2026-05-17 om 09 24 33


pascal-cm:
Very nice!

Could you also add the actions button next to the alert button? (The one with the three dots.)

Otherwise, in single node mode, there's no way of editing the instance.

This button:
image

Should be here:
image

svenvg93:
> Very nice!
> 
> Could you also add the actions button next to the alert button? (The one with the three dots.)
> 
> Otherwise, in single node mode, there's no way of editing the instance.

Good call! Missed that one. I added it for now next the Alert button. Not completely happy with how it looks as it feels a bit crowed now. I will play round with it to find a cleaner way to show them all.

PR #2027, opened 2026-05-20
fix(ui): correct Y-axis tick precision on Docker memory chart

### LINKED ISSUE #2026: [Bug]: Strange Y-Axis scaling on "Docker Memory Usage"
### Welcome!

### Component

Hub

### Problem Description

When set to anything other than the "1 minute" time frame the docker memory usage Y-axis is scaled in a weird way.
The red line is just a hair below the 2 GB indicator, while only about 1 GB is actually used.
When looking at the other indices, 300 MB and 600 MB, you would think that the Y-Axis of this chart is scaled non-linearly.

Image

Im not sure if I maybe misunderstand and some kind of logarythmic scaling is actually intended here?

As a side note, thank you very much for your work on this great tool! I really like it, especially the ability to see SMART data in a nice UI. 

### Expected Behavior

When switching to "1 minute" the chart looks as expected.

Image

### Steps to Reproduce

- "Default time period" is set to "1 hour"
- When I open the details page from the main "All Systems" page the above descibed issue is visible in the "Containers" Tab.
- Switching to "1 minute" shows the graph correctly like in "Expected Behavior"
- Switching back to "1 hour" scales it again to look like 2 GB are being used

### OS / Architecture

Agent and Hub running on arm64

### Beszel version

0.18.7

### Installation method

Docker

### Configuration

```yaml
N/A to issue
```

### Hub Logs

```json
N/A to issue
```

### Agent Logs

```shell
N/A to issue
```

svenvg93:
## 📃 Description

Fix incorrect Y-axis tick label precision on the Docker/Podman Memory Usage chart. Because container memory is stored in MB, the raw value (`val`) was always ≥ 10, so `toFixedFloat` was always called with 0 decimal places. This caused values in the 1–10 GB range to be rounded to the nearest whole GB, producing duplicate and misleading tick labels (e.g. both 1.2 GB and 1.5 GB displaying as "1 GB", and 1.8 GB displaying as "2 GB"), making the axis appear non-linear.

Also corrects a wrong JSDoc comment on the container `Stats` type — memory is in MB, not GB.

closes #2026

### 🔧 Fixed

- Docker/Podman Memory Usage chart: Y-axis tick labels for values in the 1–10 GB range now show one decimal place (e.g. "1.8 GB") instead of rounding to the nearest whole GB, preventing duplicate or misleading tick labels


henrygd (approved):
Thank you!

PR #2028, opened 2026-05-20
feat(hub): show public key in universal token settings

### LINKED ISSUE #1244: [Feature]: Make public key more visible when using universal tokens in hub UI
### Component

Hub

When a universal token is enabled in the hub's "Tokens & Fingerprints" settings page, only the token string is prominently displayed. However, the public key (which is required for agent authentication) is hidden inside the deployment command menus (Docker compose, Linux command, etc.).

For first-time users, this creates confusion because:
1. The settings page shows only the universal token
2. The documentation states that `-t` (token) is "optional for backwards compatibility"
3. Users may not realize they need to extract the public key from the deployment commands
4. This leads to failed agent deployments when users try to use only the token

The public key is actually more critical than the token for agent authentication, but the current UI makes it appear secondary or optional.

### Motivation / Use Case

As a first-time Beszel user setting up agents with universal tokens, I attempted to deploy agents using only the universal token (without the public key) because:
- The UI prominently displays only the token
- The documentation suggests the token is optional
- There's no clear indication that the public key is still required

This resulted in agents failing to connect to the hub. Only after discovering that the public key is embedded in the deployment commands (and extracting it manually) did the agents connect successfully.

This UX issue can cause confusion and deployment failures for new users who aren't familiar with Beszel's authentication model.

Possible solutions to improve the UX:

1. **Display both token and public key prominently** in the universal token section
   - Show the public key directly on the settings page alongside the token
   - Add a copy button for the public key similar to the token

2. **Add clear documentation/hints** in the UI
   - Include a note explaining that both the universal token AND public key are required
   - Add a tooltip or help text clarifying the relationship between tokens and keys

3. **Improve the deployment commands menu**
   - Make it more obvious that users should use these commands to get both credentials
   - Add a "Copy public key" option directly in the token menu

4. **Update the settings page layout**
   - Create a dedicated section showing:
     - Universal Token: `[token string]` [Copy]
     - Public Key: `[key string]` [Copy]
     - Usage instructions or link to docs

Any of these approaches would help users understand that both credentials are needed and make them easily accessible without having to dig through deployment commands.

### Screenshots

_No response_

### Category

UI / UX

### Affected Metrics

Other

svenvg93:
## 📃 Description

Improves discoverability of the public key when using universal tokens. Previously, the public key was only accessible inside the install command dropdowns, causing new users to fail agent deployments because they didn't know the public key was required alongside the token.

### ✏️ Changed

- Universal token settings section now shows the public key inline (with a copy button) when the token is enabled

## 🔧 Fixed

- Fixes #1244 — public key not visible in hub UI when using universal tokens

## 📷 Screenshots

Scherm­afbeelding 2026-05-20 om 18 18 49


henrygd (approved):
Good idea, thanks!

Issue #1918, opened 2026-04-13
[Feature]: Report fan RPMs alongside temperatures


DowLucas:
## Summary

Beszel already reads `/sys/class/hwmon/*/temp*_input` for temperatures via gopsutil's `host.SensorsTemperatures()`. It would be useful to report fan speeds from the same hwmon nodes (`fan*_input`, RPM) with the same name/label metadata.

## Motivation

On any modern x86 motherboard, the Super I/O chip (NCT679x, IT87xx, etc.) exposes multiple fan tachometer inputs. Linux's `nct6775`/`nct6683`/etc. kernel drivers surface them at `/sys/class/hwmon/*/fan{1..8}_input` in RPM. Example from an AMD Ryzen 9600X / Gigabyte B650E homelab:

```
nct6799  fan1  0 RPM      # unpopulated header
nct6799  fan2  573 RPM    # CPU fan
nct6799  fan3  493 RPM    # case fan
nct6799  fan4  390 RPM    # case fan
nct6799  fan5  560 RPM    # case fan
nct6799  fan6  399 RPM    # case fan
nct6799  fan7  0 RPM      # unpopulated header
```

This data is already sitting right next to the temperatures Beszel reads — the implementation surface is small and the user value is meaningful:

1. **Silent-fan detection / stuck bearing alerts.** A fan that suddenly drops to 0 RPM while the CPU heats up is the single most common hardware failure that leads to thermal events. Beszel already alerts on CPU temperature, but by the time temperature spikes, the damage window has opened. A "fan RPM dropped below X" alert catches it earlier.
2. **Validating fan curves.** Users tweaking BIOS fan curves or running `fancontrol` want to see the RPM/temperature correlation in a graph, not just temperatures.
3. **Closes a gap vs. other monitoring tools.** Glances, Netdata, node_exporter + Grafana, and HWiNFO all surface fans. Beszel's temperature-only view is the only noticeable omission vs. those tools on physical hosts.

## Proposed implementation

gopsutil doesn't currently expose fans (it only wraps `sensors_temperatures`), so this would need a small custom reader in `beszel-agent`:

```go
// pseudo
for each hwmon in /sys/class/hwmon/*:
    name = read(hwmon/name)
    for each fanN_input:
        rpm = read(fanN_input)
        label = read(hwmon/fanN_label) or "fanN"
        if rpm > 0:  // skip unpopulated headers
            emit(name + "_" + label, rpm)
```

[...]

PR #2032, opened 2026-05-21
feat: fan RPM monitoring + FanSpeed alerts

maxmaxme:
## 📃 Description

Adds fan RPM monitoring as a peer to the existing temperature collection, addressing #1918. Four commits, each layer parallel to Temperature:

1. **agent** — walks `/sys/class/hwmon/*/fan*_input`, skips 0 RPM (unpopulated headers), populates a new `Stats.Fans` map. Linux-only via a build-tagged `hwmonRoot` const; other platforms get an empty string and skip collection entirely. gopsutil has no equivalent for fans, hence the small custom sysfs walker per the issue's maintainer guidance.
2. **hub UI** — new **Fans** chart on the system detail page, RPM-formatted, with `$fanFilter` filter bar. Reads `stats.f` and renders one line per sensor, mirroring `TemperatureChart`.
3. **alerts** — new `FanSpeed` alert type. Fires when the highest fan RPM across all sensors goes above the user's threshold. Same per-sensor `mapSums` aggregation as Temperature so multi-minute alerts name the loudest fan in the notification descriptor (e.g. `"Highest fan pwmfan_fan1"`).
4. **docs** — readme updated to list fan speed under Supported metrics and in the alerts feature line. Companion PR opened against `beszel-docs` for the same change on the website (see below).

Verified end-to-end on a Raspberry Pi 5 (Active Cooler): agent reading `pwmfan_fan1 = 3782 RPM`, hub persisting `"f":{"pwmfan_fan1":3766}` next to `"t":{...}` in `system_stats.stats`, UI rendering the chart live, alert tests passing in both one-minute and multi-minute paths.

### Notes for the reviewer

- The new `FanSpeed` option is added via a sequential migration (`1_add_fanspeed_alert.go`) — applies on existing installs, idempotent, with a working down-migration. Verified on a real DB snapshot pulled from a live Pi install: apply → re-apply → revert, all clean.
- Locale `.po` updates intentionally omitted to match the upstream "update translations" cadence — until then the new msgids (`Fans`, `Fan speeds of system sensors`) fall back to English in every locale, same as any other newly-introduced string.
- `Info.DashboardFan` (max RPM across sensors) exists solely to give the alert path a single number to compare against the threshold. The chart keeps reading the full `Stats.Fans` map so every sensor is drawn separately — no aggregation in the visualisation.

## 📖 Documentation

Companion PR: [henrygd/beszel-docs#61](https://github.com/henrygd/beszel-docs/pull/61) — two-line update to `en/guide/what-is-beszel.md` (Supported metrics + Alerts feature line).

### ➕ Added

- Fan RPM collection from `/sys/class/hwmon/*/fan*_input` on Linux hosts.
- `Stats.Fans` wire field (CBOR `36,keyasint,omitempty`, JSON `f,omitempty`). Omitted entirely when no fan reports a non-zero RPM.
- `Info.DashboardFan` — max RPM across sensors, used by the new alert.
- `FanChart` on the system detail page mirroring `TemperatureChart` (RPM units, `$fanFilter` filter bar, sortable legend).
- `FanSpeed` alert type with multi-minute aggregation and per-sensor descriptors.
- Sequential migration `1_add_fanspeed_alert.go` extending the alerts collection's name-select with `FanSpeed` (idempotent, with down-migration).
- Unit tests for the hwmon walker (happy path, zero-RPM skip, labeled vs. unlabeled, missing root, empty root) and for the one-minute / multi-minute `FanSpeed` alert paths.

### ✏️ Changed

- `agent/system.go` calls `updateFans()` immediately after `updateTemperatures()`.
- `readme.md` — fan speed added to the supported-metrics list and alerts feature line (mirrored in companion docs PR).

## 📷 Screenshots

The new **Fans** card sits under Temperature on the system detail page:
Screenshot 2026-05-22 at 00 56 55


Fail-Safe:
This is useful, but please consider not dropping 0 RPM fan sensors once labels/history exist and adding a “below threshold” fan alert for failed/stalled fans.

Jackie264:
It's a very good function display, and I'm looking forward to being accepted as soon as possible.

Please take the time to review, thank you. @henrygd

matthewr88:
Would this work with fan speeds pulled from LHM for Windows?

ivankozlovcodes:
Tried this branch on a Raspberry Pi 4 (arm64, built from [fork](https://github.com/ivankozlovcodes/beszel)) and it works — RPM shows up in the Fans chart. Thanks for putting it together!
screenshot_20260815_041213

One request while this is still open: could `hwmonRoot` in `agent/fans_linux.go` be overridable rather than a const ? https://beszel.dev/guide/environment-variables#agent `SYS_SENSORS` seems to be fitting here. My use case is a Microchip EMC2101 fan controller. It has no mainline hwmon driver yet. I wrote a small daemon to create hwmon like readings in a sysfs shaped dir tree. Ability to overwrite `hwmonRoot` would be perfect for this.

Happy to send a patch if that's useful.

henrygd (approved):
Thanks so much for your work on this! Apologies for the delay.

I removed the alert code since it wasn't fully hooked up in the UI. It's easy enough to add later if people want it.

I also allowed RPM readings of 0 to be passed through. My laptop fans turn off sometimes, and I think we should retain that data so it's visible and included in the averaged longer time periods.

Thanks again!

henrygd:
@matthewr88 This is Linux only for now. I will look into whether we can pull fan speed from LHM when I have time, but probably not for the next release.

henrygd:
@ivankozlovcodes Would a `HOST_SYS` env var which overrides the location for all sysfs operations work?

Or is it a separate sys only for hwmon? In that case maybe a `HOST_HWMON` env var like `HOST_HWMON=/var/sys-hwmon/class/hwmon`?

PR #2034, opened 2026-05-22
feat: add Uzbek (uz) translation

ErkinjonYusupov:
## Description
Added Uzbek (uz) translation

## Changes
- Added `beszel/site/src/lib/i18n/uz.json`


henrygd (changes requested):
Thanks! Sorry for the delayed response.

Would you mind reverting the non-Uzbek language files to their starting state? We should have only a few changed files that are directly related to the Uzbek addition. I will merge as soon as this is done, thanks!

henrygd (approved):
I took care of it. Thank you again for your work.

PR #2053, opened 2026-06-01
fix: SMART_DEVICES with same path and different -d types only shows one device

Digital39999:
## 📃 Description

When `SMART_DEVICES` contains multiple entries sharing the same device path but with different `-d` types (e.g., RAID bridge slots), only the last entry survives while earlier ones are silently dropped.

Example: `/dev/sda:sat|/dev/sdb:jms56x,0|/dev/sdb:jms56x,1`
Only `/dev/sdb:jms56x,1` appears; `jms56x,0` is lost.

Why?
1. In `mergeDeviceLists`, the name-only fallback matched both `jms56x,0` and `jms56x,1` to the same scanned `/dev/sdb:sat` entry, causing the first to be overwritten by the second.
2. And `SmartData.DiskType` was set from smartctl's JSON-reported type (`sat`) instead of the user-configured type (`jms56x,0`).
3. Later, `updateSmartDevices` pruned entries because the composite key didn't match.

### 🔧 Fixed

- Multi-disk RAID enclosures (USB bridges, MegaRAID, etc.) sharing a device path in `SMART_DEVICES` now correctly track all configured drives separately.
- `SmartData.DiskType` now uses the user-configured type, ensuring device matching in `updateSmartDevices` works for same-path entries.

### DIFF
--- a/agent/smart.go
+++ b/agent/smart.go
@@ -888,7 +888,7 @@ func (sm *SmartManager) parseSmartForSata(output []byte, configuredType string) 
 	smartData.DiskName = data.Device.Name
 	smartData.DiskType = data.Device.Type
 	if configuredType != "" {
-		smartData.DiskType = strings.ToLower(strings.TrimSpace(configuredType))
+		smartData.DiskType = normalizeParserType(configuredType)
 	}
 
 	// get values from ata_device_statistics if necessary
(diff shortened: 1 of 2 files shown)


henrygd:
Thank you!

PR #2064, opened 2026-06-07
fix(ui): prevent usage bars from overflowing card border in grid view

### LINKED ISSUE #2062: [Bug]: Usage bars extend past agent box in a narrow window [Front-end]
### Welcome!

### Component

Hub

### Problem Description

I have Beszel showing a narrow screen. I noticed that if I set the size of the window slightly wider than when a single agent takes up the entire width of the screen, the usage bars on one and only one agent extend beyond the border of the box. See video. 

The only unique thing I can discernin particular about this agent, is that the memory usage is quite high. (It's my VM host, so it's passing most of its memory to a VM. 

https://github.com/user-attachments/assets/bf44ecae-a1c1-4823-a87d-c9c6e45673a6

### Expected Behavior

I would expect the usage bars shrink in the same manner as they do in the other agent boxes. 

### Steps to Reproduce

1. Have an agent with high memory usage (I think? I don't understand it only happens with one agent)
2. Resize the window to be slightly wider than the breakpoint where one agent box takes the entire width of the window. 
3. Usage bars extend past the box. 

### OS / Architecture

linux/amd64 (hub)

### Beszel version

0.18.7

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Fixes usage bars overflowing the card border in the grid view at narrow window widths. The `SystemCard` grid used `1fr` (equivalent to `minmax(auto, 1fr)`), which prevented the value column from shrinking below its content's minimum size. Combined with missing `min-w-0` on the value cell wrapper and `TableCellWithMeter`'s flex container, this caused the card content to visually overflow the card border when cards became narrow (e.g. just past the `sm` breakpoint where the grid switches from 1 to 2 columns).

fixes #2062 

### 🔧 Fixed

- Usage bars extending past the agent card border in grid view at narrow window widths

### DIFF
--- a/internal/site/src/components/systems-table/systems-table-columns.tsx
+++ b/internal/site/src/components/systems-table/systems-table-columns.tsx
@@ -470,7 +470,7 @@ function TableCellWithMeter(info: CellContext<SystemRecord, unknown>) {
 			STATUS_COLORS.down
 	)
 	return (
-		<div className="flex gap-2 items-center tabular-nums tracking-tight w-full">
+		<div className="flex gap-2 items-center tabular-nums tracking-tight w-full min-w-0">
 			<span className="min-w-8 shrink-0">{decimalString(val, val >= 10 ? 1 : 2)}%</span>
 			<span className="flex-1 min-w-8 grid bg-muted h-[1em] rounded-sm overflow-hidden">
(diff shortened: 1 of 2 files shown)


henrygd (approved):
Thanks Sven!

PR #2066, opened 2026-06-07
fix(agent): prevent a malformed Docker response from permanently breaking container stats

Quadrubo:
## 📃 Description

Fixes a bug where an agent's container stats stop showing in the hub and don't recover until the agent restarts. System stats keep working, so the agent looks healthy.

`dockerManager.decode()` reuses a single `json.Decoder` across all responses. If one response is truncated or malformed (e.g. a container removed mid-request, easy to hit during a redeploy), the decoder is left desynced and never recovers. Every later decode then fails, including the container list, so stats silently stop.

This was a regression from d67d638, which switched from a fresh `json.NewDecoder(...).Decode()` per call to a shared decoder.

The fix decodes the buffered bytes with `json.Unmarshal` instead. This keeps the reusable read buffer but holds no state between calls, so one bad response can't break later ones.

### ➕ Added

- `TestDecodeRecoversFromError`, a regression test that fails on the current code and passes with the fix.

### ✏️ Changed

- `dockerManager.decode()` now uses `json.Unmarshal` over the reusable buffer instead of a shared `json.Decoder`.

### 🔧 Fixed

- Container stats no longer stop permanently after a single malformed or truncated Docker/Podman API response.

### 🗑️ Removed

- Unused reusable `json.Decoder` field on `dockerManager`.

### DIFF
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -65,7 +65,6 @@ type dockerManager struct {
 	dockerVersionChecked bool                        // Whether a version probe has completed successfully
 	isWindows            bool                        // Whether the Docker Engine API is running on Windows
 	buf                  *bytes.Buffer               // Buffer to store and read response bodies
-	decoder              *json.Decoder               // Reusable JSON decoder that reads from buf
 	apiStats             *container.ApiStats         // Reusable API stats object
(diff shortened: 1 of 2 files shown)


henrygd:
Perfect, thank you very much!

PR #2078, opened 2026-06-14
Handle WebSocket redirects in `beszel-agent`

say4n:
## 📃 Description

  Handles HTTP redirect responses during the agent WebSocket handshake, including 308 Permanent Redirect
  responses from HTTP-to-HTTPS proxy setups. This prevents agents configured with an HTTP hub URL from failing
  immediately when the hub redirects to HTTPS.

  ## 📖 Documentation

  No documentation changes.

  ## 🪵 Changelog

  ### ✏️ Changed

  - Build the agent WebSocket URL from a copy of the configured hub URL instead of mutating the stored URL.
  - Normalize redirected http/https targets back to ws/wss before retrying the WebSocket connection.

  ### 🔧 Fixed

  - Follow WebSocket handshake redirects, including 308 Permanent Redirect.
  - Resolve absolute and relative Location headers during WebSocket redirects.
  - Cap WebSocket redirect following at five hops to avoid redirect loops.
  - Close redirect response bodies before retrying.
  - Add unit coverage for 308 redirects, relative redirects, preserved wss redirects, non-redirect responses,
    missing Location headers, and malformed redirect URLs.

  ## 📷 Screenshots

  No UI/UX changes.

## Testing

Tested with `go test -tags testing ./agent -run TestResolveWebSocketRedirect`.

### DIFF
--- a/agent/client.go
+++ b/agent/client.go
@@ -24,7 +24,9 @@ import (
 )
 
 const (
-	wsDeadline = 70 * time.Second
+	wsDeadline        = 70 * time.Second
+	maxWSRedirects    = 5
+	agentConnectRoute = "api/beszel/agent-connect"
 )
 
 // WebSocketClient manages the WebSocket connection between the agent and hub.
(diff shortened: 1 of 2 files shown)


back2root:
​I have security concerns with this approach:
​
Proper configuration is better: Why not just configure the client with correct protocol and url in the first place?

​MitM risk: The server requires HTTPS for security, but making unencrypted HTTP request first leaves the agent permanent exposed to a Man-in-the-Middle attack risk.
​
Unnecessary risk: Blindly following Location headers could introduce additional attack vectors

​It feels safer to reject 3xx responses and instruct the user to fix their config instead.

PR #2089, opened 2026-06-18
Implement OPNsense boot hook for Beszel Agent

mhlas7:
### Problem
OPNsense doesn't start enabled third-party `rc.d` services at boot the way base FreeBSD does. The install script sets `beszel_agent_enable="YES"` via `sysrc`, which is honored when you run `service beszel-agent start` manually, but is ignored during OPNsense's boot sequence. The result is a service that works until the first reboot and then silently stays down.

### Solution
OPNsense executes every script in `/usr/local/etc/rc.syshook.d/start/` at boot. When `is_opnsense` is detected, the install script now writes an executable hook there (`99-beszel-agent`) that starts the agent. This runs in addition to the existing `rc.d` setup, which is left unchanged.

The hook checks whether the service is already running before calling `service beszel-agent onestart`. The guard avoids a double start if a future OPNsense release ever begins honoring the `rc.d` enable flag.

## Testing
- Tested on OPNSense 26.1.10-amd64 installing beszel using the modified script and then rebooting OPNSense


mhlas7:
I merged in changes from upstream. Thank you to @Anikesh348 for getting pfsense compatibility in #2214! I refactored my work here to build off of that merge and add a few tweaks that make @Anikesh348 's changes work on OPNsense as well.

henrygd (approved):
Perfect, thanks for your work and sorry for the delay!

mhlas7:
Thanks for this awesome project!

PR #2090, opened 2026-06-18
fix(agent): use nvethernet MAC counters

VACInc:
## Summary
- correct Linux `nvethernet` interface byte totals using MAC octet counters from `ethtool -S`
- keep existing gopsutil behavior for other platforms/drivers
- make total bandwidth delta math reset-safe to avoid uint64 underflow after counter resets

## Why
On NVIDIA Jetson AGX Orin, `/sys/class/net/eth0/statistics/tx_bytes` can report impossible multi-terabyte TX totals while the hardware MAC counter shows only megabytes sent. Beszel reads the inflated sysfs counter and triggers bandwidth alerts even when traffic is idle.

## Evidence
Sanitized live snapshot from an affected Jetson AGX Orin, captured from Beszel's existing agent data plus a local counter probe. Account, URL, IP, and host-list details are intentionally omitted.

### Beszel alert/UI data
At `2026-06-18 08:16:52Z`, the current Beszel agent reported:

| Signal | Observed |
| --- | ---: |
| Bandwidth alert | triggered |
| Outgoing rate | `715,840,593 B/s` (~683 MiB/s) |
| Incoming rate | `15,987 B/s` |
| Per-interface cumulative TX | `20,302,331,773,103 bytes` (~18.47 TiB) |
| Per-interface cumulative RX | `447,818,781 bytes` (~427 MiB) |
| NIC temperature | `73.35 C` |

### Counter source mismatch
On the same interface/boot, Linux sysfs and the hardware MAC counters disagreed by several orders of magnitude:

```text
driver: nvethernet

sysfs tx_bytes: 20371051453685 bytes (~18.53 TiB)
sysfs tx_packets: 24758
sysfs tx_errors: 0
sysfs tx_dropped: 0

ethtool mmc_tx_octetcount_gb: 22371670 bytes (~21.33 MiB)
ethtool mmc_tx_framecount_gb: 34182
```

A 10-second sample showed the same behavior:

```text
sysfs_tx_delta=21474858046 bytes (~20.00 GiB)
mac_tx_delta=22315 bytes (~21.79 KiB)
sysfs_rx_delta=141250 bytes
mac_rx_delta=153958 bytes
```

This is why the fix is scoped to Linux interfaces whose driver symlink is exactly `nvethernet`: replace the bogus sysfs/gopsutil byte totals with the NIC's MAC octet counters before Beszel computes rates.

## Tests
- `go test -tags testing ./agent`
- `go test ./agent ./internal/cmd/agent`
- `go test -tags testing ./...` was attempted but this checkout lacks the embedded web `internal/site/dist` assets, so hub packages fail with `pattern all:dist: no matching files found`.

## Review
- Ran `codex review --uncommitted` before opening the PR; no actionable regressions reported. Codex review could not run tests in its read-only sandbox, so the focused tests above were run locally.

### DIFF
--- a/agent/network.go
+++ b/agent/network.go
@@ -83,6 +83,7 @@ func (a *Agent) updateNetworkStats(cacheTimeMs uint16, systemStats *system.Stats
 	a.ensureNetworkInterfacesMap(systemStats)
 
 	if netIO, err := psutilNet.IOCounters(true); err == nil {
+		netIO = correctNetworkCounterStats(netIO)
 		nis, msElapsed := a.loadAndTickNetBaseline(cacheTimeMs)
 		totalBytesSent, totalBytesRecv := a.sumAndTrackPerNicDeltas(cacheTimeMs, msElapsed, netIO, systemStats)
 		bytesSentPerSecond, bytesRecvPerSecond := a.computeBytesPerSecond(msElapsed, totalBytesSent, totalBytesRecv, nis)
@@ -103,6 +104,7 @@ func (a *Agent) initializeNetIoStats() {
 
 	// get current network I/O stats and record valid interfaces
 	if netIO, err := psutilNet.IOCounters(true); err == nil {
+		netIO = correctNetworkCounterStats(netIO)
 		for _, v := range netIO {
 			if skipNetworkInterface(v, nicCfg) {
 				continue
@@ -198,10 +200,17 @@ func (a *Agent) sumAndTrackPerNicDeltas(cacheTimeMs uint16, msElapsed uint64, ne
 // computeBytesPerSecond calculates per-second totals from elapsed time and totals
 func (a *Agent) computeBytesPerSecond(msElapsed, totalBytesSent, totalBytesRecv uint64, nis system.NetIoStats) (bytesSentPerSecond, bytesRecvPerSecond uint64) {
 	if msElapsed > 0 {
-		bytesSentPerSecond = (totalBytesSent - nis.BytesSent) * 1000 / msElapsed
-		bytesRecvPerSecond = (totalBytesRecv - nis.BytesRecv) * 1000 / msElapsed
+		bytesSentPerSecond = counterDelta(totalBytesSent, nis.BytesSent) * 1000 / msElapsed
+		bytesRecvPerSecond = counterDelta(totalBytesRecv, nis.BytesRecv) * 1000 / msElapsed
 	}
 	return bytesSentPerSecond, bytesRecvPerSecond
+}
+
+func counterDelta(current, previous uint64) uint64 {
+	if current >= previous {
+		return current - previous
+	}
+	return current
 }
 
 // applyNetworkTotals validates and writes computed network stats, or resets on anomaly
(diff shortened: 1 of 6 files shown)


henrygd (changes requested):
Thanks, sorry for the delay. I was mostly off the computer for a while.

Looks good, but if possible I'd like to avoid calling `ethtool`. There's a widely used [ethtool "like" library for Go](https://github.com/safchain/ethtool). Maybe we can use that instead?

```
go get github.com/safchain/ethtool
```

Here's an example reference. Note that I used AI to generate it. I can't test but it seems like a fairly straightforward swap.

**network_counters_linux.go**

```go
//go:build linux

package agent

import (
	"log/slog"
	"math"
	"os"
	"path/filepath"
	"strings"

	"github.com/safchain/ethtool"
	psutilNet "github.com/shirou/gopsutil/v4/net"
)

func correctNetworkCounterStat(v psutilNet.IOCountersStat) psutilNet.IOCountersStat {
	if !isNvidiaEthernet(v.Name) {
		return v
	}

	tx, rx, ok := readEthtoolMACOctets(v.Name)
	if !ok {
		return v
	}

	v.BytesSent = tx
	v.BytesRecv = rx
	return v
}

func isNvidiaEthernet(name string) bool {
	if name == "" || strings.Contains(name, "/") {
		return false
	}

	driverPath := filepath.Join("/sys/class/net", name, "device/driver")
	target, err := os.Readlink(driverPath)
	if err != nil {
		return false
	}
	return filepath.Base(target) == "nvethernet"
}

func readEthtoolMACOctets(name string) (tx, rx uint64, ok bool) {
	stats, err := ethtool.Stats(name)
	if err != nil {
		slog.Debug("Failed to read ethtool network counters", "interface", name, "err", err)
		return 0, 0, false
	}

	tx, okTx := ethtoolCounter(stats, "mmc_tx_octetcount_gb", "mmc_tx_octetcount_gb_h")
	rx, okRx := ethtoolCounter(stats, "mmc_rx_octetcount_gb", "mmc_rx_octetcount_gb_h")
	if !okTx || !okRx {
		return 0, 0, false
	}
	return tx, rx, true
}

func ethtoolCounter(stats map[string]uint64, lowKey, highKey string) (uint64, bool) {
	low, ok := stats[lowKey]
	if !ok {
		return 0, false
	}
	high, hasHigh := stats[highKey]
	if !hasHigh {
		return low, true
	}
	if high > (math.MaxUint64-low)>>32 {
		return 0, false
	}
	return high<<32 + low, true
}

```

The AI also changed the bitwise slightly. Here's the justification it gave. I haven't checked the driver implementation so not sure if it's correct.

> `agent/network_counters_linux.go:101` combines the high and low values with bitwise OR. NVIDIA’s driver accumulates both registers independently into 64-bit software fields, so the low value can exceed 32 bits; overlapping bits are then lost. Use checked arithmetic equivalent to (high << 32) + low. See the Jetson driver’s accumulator implementation (https://github.com/OE4T/linux-tegra-5.10/blob/oe4t-patches-l4t-35.5.0/nvidia/drivers/net/ethernet/nvidia/nvethernet/nvethernetrm/osi/core/mgbe_mmc.c).

Issue #1345, opened 2025-10-29
[Bug]: S.M.A.R.T isnt working after upgrade to 0.15.1


Mateusz779:
### Component

Agent

### Description

After upgrading the agent to 0.15.1, the S.M.A.R.T. statistics stop working and display only the following data:

<img width="1079" height="314" alt="Image" src="https://github.com/user-attachments/assets/c03b69d8-fa7e-4168-8375-7a961a94c686" />
 

### Expected Behavior

Agents running on 0.15.0 work fine with hub on both version 0.15.1 and 0.15.0.

<img width="1103" height="515" alt="Image" src="https://github.com/user-attachments/assets/c7416d95-a5cf-4b22-b786-7ccaccdca067" />

### Steps to Reproduce

1. After upgrading the agent to version 0.15.1, while the hub remains on 0.15.0 or 0.15.1, the result is the same.

### Category

Metrics

### Affected Metrics

Storage

### OS / Architecture

linux/amd64(agent), linux/amd64(hub)

### Beszel version

0.15.1

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

henrygd:
Thanks, can anyone with this issue please post the output of this on the agent system?

```bash
docker exec beszel-agent smartctl --scan
```

henrygd:
Also, try using the `SMART_DEVICES` env var to fix this:

```dotenv
SMART_DEVICES=/dev/sda:sat,/dev/sdb:sat,/dev/sdc:sat,/dev/sdd:sat,/dev/sde:sat
```

sotasan:
I have the exact same issue as @Mateusz779 and this is my output:

![Image](https://github.com/user-attachments/assets/1e06e0ab-7f58-47f7-ab15-c21083df7ab2)

Mateusz779:
@henrygd 
Here is the command result, after adding to docker compose to the ENV what you sent, the command result is the same, but in the panel the disks are shown correctly

<img width="433" height="79" alt="Image" src="https://github.com/user-attachments/assets/cd6eff12-771f-40fd-943d-dc86860874b5" />

henrygd:
Thank you both! It looks like `smartctl` wrongly reports them as `scsi` in the scan.

Previously we didn't handle SCSI so I think it defaulted to `sat`. Let me think about how to handle it.

In the meantime, use the `SMART_DEVICES` env var or downgrade agents to 0.15.0.

[...]

PR #2102, opened 2026-06-27
fix(agent): honor explicit SMART_DEVICES type hint instead of scan-detected type

### LINKED ISSUE #2072: [Bug]: `SMART_DEVICES` type hint ignored — agent uses `smartctl --scan` result instead, causing USB drive SMART collection to fail
### Welcome!

### Component

Agent

### Problem Description

When `SMART_DEVICES` is configured with an explicit device type (e.g. `/dev/sda:scsi`), the agent ignores the specified type and instead uses the type returned by its internal `smartctl --scan` call. For USB-attached drives where the enclosure's bridge chip does not support SAT/ATA passthrough, `--scan` returns `-d sat`, which fails with exit status 2. The correct type (`:scsi`) works when called manually, but the agent never uses it.

---

### Additional details
`sudo smartctl --scan` output:
```
/dev/sda -d sat # /dev/sda [SAT], ATA device
/dev/nvme0 -d nvme # /dev/nvme0, NVMe device
```

`/dev/sda` device details:
Seagate Ultra Touch HDD (USB-attached)
`smartctl -d scsi /dev/sda` → exit status 0, full SMART data returned
`smartctl -d sat /dev/sda` → exit status 2, no data

`/dev/nvme0` device details:
Corsair MP600 Mini (direct NVMe)
Works correctly, included here for contrast

smartctl version: 7.3 (2022-02-28)

### Related issue
#1791 is superficially similar — it also involved the sat parser being used despite SCSI being the correct type — but describes a different failure mode: the startup poll worked correctly while manual hub-triggered refreshes produced zeros, because the sat parser was applied unconditionally on refresh. That issue was fixed in v0.18.5.
The present bug is distinct in two ways. First, it affects the regular polling cycle, not just manual refreshes. Second, the type hint in `SMART_DEVICES` is actively ignored: the agent reads the config correctly (`INFO SMART_DEVICES value=/dev/sda:scsi`), then immediately overrides it with the `--scan` result (`DEBUG smartctl scan name=/dev/sda type=sat protocol=ATA`) and uses sat for every subsequent query. The fix for #1791 did not address this code path.
The issue may share a root cause with #1791 — both suggest the agent does not reliably honor the device type determined at scan time when making the actual smartctl call — but the reproduction path and affected code path appear different enough to warrant a separate report.

### Expected Behavior

The type specified in `SMART_DEVICES` (e.g. `:scsi`) should be passed to smartctl when querying that device, overriding the scan-detected type. The purpose of the `device:type` syntax in `SMART_DEVICES` is precisely to handle cases where auto-detection produces the wrong result.

### Steps to Reproduce

1. Attach a USB hard drive whose bridge does not support SAT/ATA passthrough (`smartctl --scan` reports it as `-d sat`, but `-d sat` fails with exit status 2)
2. Confirm the correct type works manually: `smartctl -H -d scsi /dev/sda` succeeds
3. Set `SMART_DEVICES=/dev/sda:scsi` in the agent service file
4. Restart the agent and wait for the SMART collection interval
5. Observe in agent logs: the scan result (sat) is used instead of the configured type (scsi), and collection fails

### OS / Architecture

aarch64-linux-6.12.62+rpt-rpi-2712

### Beszel version

0.18.7

### Installation method

Binary

### Configuration

```yaml
[Unit]
Description=Beszel Agent Service
Wants=network-online.target
After=network-online.target

[Service]
Environment="PORT=45876"
Environment="KEY=ssh-XXX"
Environment="TOKEN=XXX"
Environment="HUB_URL=https://beszel.XXX.XXX"
Environment="EXTRA_FILESYSTEMS=sda1"
Environment="SMART_DEVICES=/dev/sda:scsi,/dev/nvme0:nvme"
ExecStart=/opt/beszel-agent/beszel-agent
User=beszel
Restart=on-failure
RestartSec=5
StateDirectory=beszel-agent
KeyringMode=private
LockPersonality=yes
ProtectClock=yes
ProtectHome=read-only
ProtectHostname=yes
ProtectKernelLogs=yes
ProtectSystem=strict
RemoveIPC=yes
RestrictSUIDSGID=true
AmbientCapabilities=CAP_SYS_RAWIO CAP_SYS_ADMIN
CapabilityBoundingSet=CAP_SYS_RAWIO CAP_SYS_ADMIN

[Install]
WantedBy=multi-user.target
```

### Hub Logs

```json
N/A — the failure occurs entirely in the agent before any data reaches the hub. The SMART section for /dev/sda never appears in the UI.
```

### Agent Logs

```shell
Startup and SMART collection with LOG_LEVEL=debug and SMART_INTERVAL=1m:

INFO  Data directory path=/var/lib/beszel-agent
DEBUG 0.18.7
DEBUG smartctl path=/usr/sbin/smartctl err=<nil>
...
INFO  SMART_INTERVAL duration=1m0s
INFO  SMART_DEVICES value=/dev/sda:scsi,/dev/nvme0:nvme
DEBUG smartctl scan name=/dev/sda type=sat protocol=ATA
DEBUG smartctl scan name=/dev/nvme0 type=nvme protocol=NVMe
INFO  smartctl failed device=/dev/sda err="exit status 2"
DEBUG smartctl collect failed device=/dev/sda err="exit status 2"

Note that /dev/nvme0 succeeds because nvme matches both the configured and scan-detected type. The bug manifests specifically when the configured type differs from what --scan returns. The log line DEBUG smartctl scan name=/dev/sda type=sat shows the scan result overriding the :scsi hint from SMART_DEVICES.
```

BootstrapperSBL:
Fixes #2072.

### Problem
When `SMART_DEVICES` specifies an explicit device type (e.g. `/dev/sda:scsi`), the agent parses and resolves the type correctly, but `smartctlArgs` intentionally omits the `-d` flag for `scsi`/`ata` — a workaround for scan *misdetection* (#1345). That workaround also discards an **explicit** user-provided type, so smartctl falls back to auto-detection and picks the wrong type (`sat`). On USB drives whose bridge chip doesn't support SAT/ATA passthrough, `-d sat` fails with exit status 2 and SMART collection never succeeds, even though `/dev/sda:scsi` was configured precisely to avoid that.

### Fix
Track whether a device type came from an explicit `SMART_DEVICES` `:type` hint (`DeviceInfo.explicitType`) and always pass it to smartctl via `-d`. Scan-detected `scsi`/`ata` types are still left off so smartctl can auto-detect them, preserving the #1345 behavior. The flag is threaded through `mergeDeviceLists` so it survives both config merges and verified-device rescans.

### Before / After
For `SMART_DEVICES=/dev/sda:scsi` where `smartctl --scan` reports `/dev/sda` as `sat`:

- **Before:** `smartctl -a --json=c /dev/sda` → smartctl auto-detects `sat` → exit status 2, no data.
- **After:** `smartctl -d scsi -a --json=c /dev/sda` → correct type used, data collected.

### Tests
- `TestSmartctlArgsExplicitType` — explicit `scsi`/`ata` get `-d`; scan-detected `scsi` still doesn't.
- `TestSmartDevicesExplicitTypeFlowsToSmartctlArgs` — full parse → merge → args regression for this issue.
- `TestMergeDeviceListsPreservesExplicitTypeAcrossRescan` — the explicit flag survives a rescan.

All three fail without the fix and pass with it. `go test -tags=testing ./...` for the smart suite passes; `gofmt`/`go vet` clean.

### DIFF
--- a/agent/smart.go
+++ b/agent/smart.go
@@ -55,6 +55,11 @@ type DeviceInfo struct {
 	typeVerified bool
 	// parserType holds the parser type (nvme, sat, scsi) that last succeeded.
 	parserType string
+	// explicitType reports whether Type came from an explicit ":type" hint in
+	// SMART_DEVICES. Such a type is a deliberate user override and must always be
+	// passed to smartctl via -d, even for scsi/ata where a scan-detected type is
+	// otherwise left off (see smartctlArgs and issue #1345).
+	explicitType bool
 }
 
 // deviceKey is a composite key for a device, used to identify a device uniquely.
@@ -251,8 +256,9 @@ func (sm *SmartManager) parseConfiguredDevices(config string) ([]*DeviceInfo, er
 		}
 
 		devices = append(devices, &DeviceInfo{
-			Name: name,
-			Type: devType,
+			Name:         name,
+			Type:         devType,
+			explicitType: devType != "",
 		})
 	}
 
@@ -558,7 +564,9 @@ func (sm *SmartManager) smartctlArgs(deviceInfo *DeviceInfo, includeStandby bool
 		deviceType = strings.ToLower(deviceInfo.Type)
 		parserType = strings.ToLower(deviceInfo.parserType)
 		// types sometimes misidentified in scan; see github.com/henrygd/beszel/issues/1345
-		if deviceType != "" && deviceType != "scsi" && deviceType != "ata" {
+		// An explicit SMART_DEVICES ":type" hint is a deliberate override, so always
+		// pass it through; otherwise scsi/ata are left off so smartctl can auto-detect.
+		if deviceType != "" && (deviceInfo.explicitType || (deviceType != "scsi" && deviceType != "ata")) {
 			args = append(args, "-d", deviceInfo.Type)
 		}
 	}
@@ -663,6 +671,9 @@ func mergeDeviceLists(existing, scanned, configured []*DeviceInfo) []*DeviceInfo
 		target.Type = prev.Type
 		target.typeVerified = true
 		target.parserType = prev.parserType
+		if prev.explicitType {
+			target.explicitType = true
+		}
 	}
 
 	// applyConfiguredMetadata updates a matched device with any configured
@@ -675,6 +686,9 @@ func mergeDeviceLists(existing, scanned, configured []*DeviceInfo) []*DeviceInfo
 			existingDev.Type = newType
 			existingDev.typeVerified = false
 			existingDev.parserType = normalizeParserType(newType)
+		}
+		if configuredDev.explicitType {
+			existingDev.explicitType = true
 		}
 		if configuredDev.InfoName != "" {
 			existingDev.InfoName = configuredDev.InfoName
--- a/agent/smart_test.go
+++ b/agent/smart_test.go
@@ -392,6 +392,81 @@ func TestSmartctlArgs(t *testing.T) {
 	)
 }
 
+// TestSmartctlArgsExplicitType verifies that an explicit SMART_DEVICES type hint
+// is always passed to smartctl via -d, while a scan-detected scsi/ata type is
+// still left off so smartctl can auto-detect it (see issue #1345).
+func TestSmartctlArgsExplicitType(t *testing.T) {
+	sm := &SmartManager{}
+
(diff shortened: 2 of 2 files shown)


henrygd (approved):
Looks great, thanks very much for your work!

PR #2107, opened 2026-07-01
fix(scripts): update agent env vars on reinstall instead of skipping

### LINKED ISSUE #2097: [Bug]: Env variables not updated
### Welcome!

### Component

Agent

### Problem Description

On the agents, when you want to redeploy beszel-agent, the new variabls token, key and hub_url  are not rewriten!

Tested on Windows 11 with nssm and Debian 13 with systemd

### Expected Behavior

Thoses variables should be overwritten when we copy/paste the new config from beszel hub!

### Steps to Reproduce

1. Configure a host on one beszel instance
2. Add a second beszel instance
3. copy new config from second beszel instance and paste on agent
4. Agent not beeing reconfigured for second instance

### OS / Architecture

windows 11 (agent), debian 13 ( agent), debian 13 + docker (hub)

### Beszel version

latest

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json
no log
```

### Agent Logs

```shell
no log
```

svenvg93:
## 📃 Description

Fixes a bug where re-running the install script to reconfigure the agent (e.g. pointing it at a different hub) would silently skip updating `KEY`, `TOKEN`, `HUB_URL`, and `PORT` because the existing service file was detected and creation was skipped entirely.

Closes #2097

### ✏️ Changed

- **systemd**: Instead of skipping when the service file already exists, update the four `Environment=` lines in-place with `sed -i`, preserving any user customizations (e.g. `EXTRA_FILESYSTEMS`)
- **FreeBSD**: Always overwrite the `env` file on reinstall since it only contains beszel-managed variables
- **Alpine OpenRC**: Update the `export` lines in-place with `sed -i` when the init.d file already exists
- **OpenWRT procd**: Replace the `procd_set_param env` line in-place with `sed -i` when the init script already exists
- **Windows (NSSM)**: When the service already exists with the correct binary path, update `AppEnvironmentExtra` with the new values instead of returning early without changes

### DIFF
--- a/supplemental/scripts/install-agent.ps1
+++ b/supplemental/scripts/install-agent.ps1
@@ -9,8 +9,20 @@ param (
     [string]$NSSMPath = "",
     [switch]$ConfigureFirewall,
     [ValidateSet("Auto", "Scoop", "WinGet")]
-    [string]$InstallMethod = "Auto"
+    [string]$InstallMethod = "Auto",
+    # Set automatically from $PSBoundParameters below, or forwarded through an elevated relaunch.
+    # Used so a reinstall only overwrites Token/Url/Port on an existing service if the caller
+    # actually asked to change them, instead of wiping them with their unset defaults.
+    [switch]$TokenProvided,
+    [switch]$UrlProvided,
+    [switch]$PortProvided
 )
+
+if (-not $Elevated) {
+    $TokenProvided = $PSBoundParameters.ContainsKey('Token')
+    $UrlProvided = $PSBoundParameters.ContainsKey('Url')
+    $PortProvided = $PSBoundParameters.ContainsKey('Port')
+}
 
 # Check if required parameters are provided
 if ([string]::IsNullOrWhiteSpace($Key)) {
@@ -312,7 +324,10 @@ function Install-NSSMService {
         [string]$HubUrl = "",
         [Parameter(Mandatory=$true)]
         [int]$Port,
-        [string]$NSSMPath = ""
+        [string]$NSSMPath = "",
+        [switch]$TokenProvided,
+        [switch]$UrlProvided,
+        [switch]$PortProvided
     )
     
     Write-Host "Installing beszel-agent service..."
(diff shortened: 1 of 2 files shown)


henrygd:
Sorry, first time I've used that copilot review and not sure if it's accurate. Don't have time to review at the moment but will try to get back to it soon.

svenvg93:
> Sorry, first time I've used that copilot review and not sure if it's accurate. Don't have time to review at the moment but will try to get back to it soon.

No worries , I will check it out when I’m back from holiday. Also not 100% sure this is something we should handle in the install script 🤷🏻‍♂️

henrygd:
Thanks for your dedication Sven. No need to work on anything here while on holiday :)

svenvg93:
Hi @henrygd, 

I had a look at the Copilot and made some adjusted with help of Claude. As I don't have knowlegde of all the diffrenent OS types :)

mews-se:
Ran into this yesterday on three machines that already had the agent installed, in case it helps with the "should the installer handle this" question from above.

The install command copied out of Add System printed `installed successfully` on all three without touching the unit file. On one the agent kept its old token and sat in a `401` reconnect loop every ten seconds; on another `TOKEN=` was simply left empty. Nothing in the output hinted that the values had been ignored, so from the hub side it just looks like the machine never comes up. Running a patched script that does what you do here fixed all three immediately.

One small thing while you're in there: the replacement side of the `sed` expressions takes the values raw, so `&`, `|` or a backslash in a value breaks the substitution — `|` is the delimiter, `&` expands to the whole match, and a backslash is eaten:

```
value=http://h/?a=1&b=2  ->  Environment="HUB_URL=http://h/?a=1Environment="HUB_URL=old"b=2"
value=http://h/a|b       ->  sed: bad flag in substitute command: 'b'
value=back\slash         ->  Environment="HUB_URL=backslash"
```

Keys and normal hub URLs don't contain those characters, so it's belt and braces rather than a bug report. Escaping first covers it:

```sh
escaped=$(printf '%s' "$value" | sed 's/[\\&|]/\\&/g')
```

henrygd (approved):
I pushed a small change to improve escaping and I think it's good to go at this point. Thank you!

PR #2126, opened 2026-07-09
fix(hub): re-dial SSH-pull agents when a connection silently dies

### LINKED ISSUE #2041: [Bug]: Some SSH-pull agents silently go dead per-agent; hub never re-dials until restart (v0.18.7)
## Summary

On v0.18.7 with the SSH-pull connection model, some agent connections silently go dead and the hub **never re-dials them until the hub process is restarted**. The dead agents continue to show as `up` in the system list (because their last `info` snapshot is recent enough), so the bug is invisible from the UI's status indicator — only the per-system detail charts (which read from `system_stats`) show the gap.

The behavior is **asymmetric across agents**: in a 5-agent fleet, some agents flow at ~70% coverage while others sit at ~20% with multi-hour silences, and which agents end up in each bucket is variable. Restarting the hub (`docker restart`) momentarily restores all 5 agents (every one re-dials successfully within ~30 s), then 2 of 5 typically drift back to "dead-but-up" within minutes.

The hub log does not record any disconnect / error / reconnect attempt for the dead agents. The agent-side logs also go quiet — for one of our agents, the agent log shows the hub stopped initiating SSH connections for **~13 consecutive hours** even though the agent's listener was up the entire time, the network was reachable, and the agent's own process was healthy.

## Environment

- **Hub:** `henrygd/beszel:latest` (image label `0.18.7`) under Docker, `network_mode: host`, on macOS 15 arm64. PocketBase WAL mode. Hub healthy (`/api/health` → 200) throughout.
- **Agents — 5 total, deliberately mixed:**
  - 2× **macOS arm64 native** binary (`beszel-agent 0.18.7`) under `launchd`
  - 1× **macOS arm64 via Docker** (`henrygd/beszel-agent:latest`, `--network host`)
  - 1× **Linux x86_64 baremetal** via systemd (binary from the official install script)
  - 1× **Linux x86_64** (Synology DSM) via Docker (`--network host`)
- **Connection model:** SSH-pull. Hub dials out to each agent's listen port using its `id_ed25519`; no per-agent token.
- All on the same private LAN subnet. Hub can reach every agent listen port at all times (verified with `nc` from inside the hub container while in host-net mode — including during the "dead-but-up" windows).
- All 5 agents on `0.18.7`. No version mismatch.

## Observed pattern (the asymmetric part)

Hourly histogram from `system_stats` over a representative 2-hour window. Healthy = ~60 1m-rows/hour. Agent identifiers anonymized; OS/deploy method preserved because that's the relevant axis:

| Agent | OS / deploy | Rows hour N | Rows hour N+1 | Coverage |
|---|---|---|---|---|
| A | macOS arm64, Docker `--network host` | 51 | 34 | ~71% |
| B | macOS arm64 native binary (launchd) | 46 | 32 | ~65% |
| C | Linux x86_64 native binary (systemd) | 33 | 28 | ~51% |
| D | Linux x86_64 Docker (Synology DSM) `--network host` | 14 | 15 | **~24%** |
| E | macOS arm64 native binary (launchd) | 14 | 11 | **~21%** |

Two agents (D and E in this snapshot) are at ~20% coverage — i.e. effectively dead for most of every hour, with the hub never noticing. The other three vary from gappy to mostly healthy. **There is no OS pattern** (a macOS native and a Linux Docker share the worst slot here; another macOS native is mid-pack; the macOS Docker is best). The pairing reshuffles across cycles — it isn't always the same two agents that end up dead.

Current per-system staleness query at the moment of filing:

```sql
SELECT s.name,
       s.status,
       MAX(ss.created) AS latest,
       ROUND((julianday('now') - julianday(MAX(ss.created))) * 1440.0, 1) AS minutes_ago
  FROM systems s
  LEFT JOIN system_stats ss ON ss.system = s.id
 GROUP BY s.id;
```

Sample output (all five report `status=up`):

```
Agent A   up   <ts>   0.8 min ago
Agent B   up   <ts>   0.7 min ago
Agent C   up   <ts>   0.7 min ago
Agent D   up   <ts>  19.7 min ago     ← stale, but UI shows up
Agent E   up   <ts>  23.6 min ago     ← stale, but UI shows up
```

## Logs

**Hub** (`docker logs beszel`):

```
YYYY/MM/DD hh:mm:ss Server started at http://0.0.0.0:8090
├─ REST API:  http://0.0.0.0:8090/api/
└─ Dashboard: http://0.0.0.0:8090/_/
```

Nothing further. No disconnect, no SSH error, no redial attempt for the dead agents.

**Agent side (macOS native binary):** the agent log records each `INFO SSH connected addr=<hub-ip>:<port>` line as the hub initiates a pull. Excerpt from one of the worst-affected agents (anonymized timestamps):

```
2026/05/26 00:07:27 INFO SSH connected addr=<hub>:59357
2026/05/26 00:17:27 INFO SSH connected addr=<hub>:61028
2026/05/26 00:27:27 INFO SSH connected addr=<hub>:62694
2026/05/26 13:48:32 INFO SSH connected addr=<hub>:49259    ← 13h21m gap; hub stopped dialing
2026/05/26 13:52:37 INFO SSH connected addr=<hub>:49989    ← right after a manual hub restart
2026/05/26 14:27:38 INFO SSH connected addr=<hub>:55965
2026/05/26 14:47:39 INFO SSH connected addr=<hub>:59205
2026/05/26 15:12:40 INFO SSH connected addr=<hub>:61450
```

For 13+ hours the hub simply did not dial this agent — and there is no corresponding log line on the hub side explaining why. After the manual restart at ~13:52, the hub started dialing again at intervals, but the per-dial cadence is irregular (10 min, 35 min, 20 min, 25 min) instead of the expected 1-minute cadence implied by the 1m `system_stats` collection interval.

The agent's own process was running and listening on its port the entire time (`launchctl list` + `ps`). Network reachability from the hub host to the agent port was fine throughout (`nc -vz` succeeded). It is purely the hub-side dial scheduling that stalled.

## What we ruled out

- **`#531` / `#796` (gopsutil/macOS temperature-sensor null-byte corruption).** Closed in v0.18.7. Set `SENSORS=""` on both macOS-native agents to be safe. The stall reproduces, and crucially **both Linux agents also stall** at times — so this is not the temp-sensor bug.
- **Docker bridge networking.** Hub is in `network_mode: host`. `nc -vz` from inside the hub container reaches every agent listen port at every point in the cycle.
- **macOS sleep / power events on the hub host.** `pmset -g log` shows no sleep events at the stall windows.
- **macOS firewall.** Off on hub and macOS agents.
- **DNS / NAT.** Pure private-LAN dials, IPs not hostnames, no NAT or proxy.
- **Version mismatch.** Hub and all agents on `0.18.7`.
- **Per-OS pattern.** Both macOS (native + Docker) and Linux (native + Docker) agents end up in the "dead-but-up" bucket. Reshuffles across cycles.

## Workaround pattern (also evidence)

We have a `launchd` watchdog that polls `system_stats` freshness every 5 min and runs `docker restart beszel` when fewer than 3 of 5 agents have a row in the last 5 minutes. It works in the sense that the restart always re-dials all 5 — but it also reveals the bug's character: the watchdog spent ~2.5 hours straight today logging `OK: 3 systems fresh in last 5m` while two specific agents were effectively dead the whole time. The hub considered itself healthy; the UI considered the agents up; only a `LEFT JOIN system_stats` revealed the stall.

When the watchdog does fire `docker restart beszel`:

1. Within ~30 s, all 5 agents re-dial successfully (visible in agent logs).
2. For 1–2 ticks, all 5 systems have fresh `system_stats` rows.
3. Within 5–10 minutes, 2 of the 5 quietly drift back to "dead-but-up". The hub does not redial them.
4. The cycle repeats — sometimes the same 2 agents, sometimes different ones.

## Hypothesis

This looks like per-agent state on the hub that, once a connection enters a particular "silently failed" state, never re-enters the dial scheduler. The fact that:

- The hub log shows no error or redial attempt for the affected agents,
- A full hub restart reliably forces a fresh dial to every agent,
- Which agents end up "stuck" varies cycle-to-cycle,
- Network reachability is fine throughout,

suggests the hub is holding a per-system "connection alive" belief that is not being invalidated when the data path goes silent. A TCP keepalive / read-side dead-peer detection on the SSH pull path that would invalidate that belief, plus a redial scheduler that re-attempts on invalidation, would seem to address this.

## Asks

1. Is there a hub-side keepalive / read-timeout / dead-peer-detection knob in v0.18.7 we can tune to force a redial on a silent connection failure? I could not find one in the docs.
2. If not, would you be open to adding one? Even a simple "if no rows received from agent X in N collection intervals → tear down and redial" would seem to close the loop based on what we're seeing.
3. Is there per-system state I can inspect (in `data.db` or via the API) that would tell us *why* the hub is no longer dialing a given agent? I checked the `systems` table; nothing obvious distinguishes a "dialing fine" agent from a "stuck not-dialing" one besides the `updated` timestamp.
4. Is WebSocket-push the recommended path for a LAN-stable but long-running fleet? (We attempted WS-push on a macOS agent and the agent's launchd-spawned binary was blocked from dialing outbound to a LAN IP by macOS Local Network privacy / TCC — separate issue, happy to file separately. SSH-pull is otherwise a great fit for our fleet.)

Happy to provide more diagnostics — longer watchdog log, packet captures of a stall transition, `tcpdump` on either side, sqlite dumps, anything else useful. Thanks for the project.

TowyTowy:
## 📃 Description

On the SSH-pull connection model, a per-system SSH connection could go half-open (a dead peer that never sends RST/FIN) or an agent could accept the session but never write a response. The hub's data read (fetchDataViaSSH → runSSHOperation) had no deadline, so it blocked forever. Because StartUpdater runs update() synchronously on its ticker, the blocked read froze the whole per-system goroutine — the ticker's later ticks were dropped, no error was returned (so the system stayed "up"), and the agent was never re-dialed until the hub process was restarted.

This matches #2041: agents silently drift to "dead-but-up", the hub logs no disconnect/redial, restart temporarily fixes everything, and which agents stall reshuffles between cycles.

### Fix
- Bound each SSH data exchange with a timeout (`runWithTimeout`). On timeout the connection is torn down (unwinding the blocked read) and a retryable error is returned, so the next collection tick re-dials. The bound (20s) is kept under the 60s collection interval so a stalled agent recovers within one cycle.
- Enable TCP keep-alive on dialed SSH connections as a backstop so the OS eventually detects a genuinely dead idle peer.

### 🔧 Fixed

- SSH-pull agents that silently went dead are now detected and re-dialed instead of staying "up" with stale stats until the hub restarts (#2041).

Fixes #2041

### DIFF
--- a/internal/hub/systems/ssh_timeout_test.go
+++ b/internal/hub/systems/ssh_timeout_test.go
@@ -1,0 +1,56 @@
+//go:build testing
+
+package systems
+
+import (
+	"errors"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestRunWithTimeout covers the guard added for issue #2041: the per-system SSH
+// data exchange must never block the updater indefinitely on a dead connection.
+func TestRunWithTimeout(t *testing.T) {
+	t.Run("returns the operation result when it completes before the timeout", func(t *testing.T) {
+		synctest.Test(t, func(t *testing.T) {
+			wantErr := errors.New("boom")
+			onTimeoutCalled := false
+
+			retry, err := runWithTimeout(10*time.Second, func() (bool, error) {
+				return true, wantErr
+			}, func() { onTimeoutCalled = true })
+
+			assert.True(t, retry, "should return the operation's retry value")
+			assert.Equal(t, wantErr, err, "should return the operation's error")
+			assert.False(t, onTimeoutCalled, "onTimeout must not fire when the op completes")
+		})
+	})
+
+	t.Run("times out and tears down the connection when the op blocks", func(t *testing.T) {
+		synctest.Test(t, func(t *testing.T) {
+			// unblock simulates a half-open connection: the op is stuck reading a
+			// response that never arrives until the connection is torn down.
+			unblock := make(chan struct{})
+			onTimeoutCalled := false
+			start := time.Now()
+
+			retry, err := runWithTimeout(5*time.Second, func() (bool, error) {
+				<-unblock
+				return false, nil
+			}, func() {
+				onTimeoutCalled = true
+				close(unblock) // tearing down the connection releases the blocked read
+			})
+
+			assert.Equal(t, 5*time.Second, time.Since(start), "should return exactly at the timeout")
+			assert.True(t, retry, "a timeout should be retryable so the next tick re-dials")
+			assert.Error(t, err, "a timeout must surface an error so the system is set down")
+			assert.True(t, onTimeoutCalled, "onTimeout must fire so the dead connection is closed")
+
+			synctest.Wait() // ensure the released op goroutine exits cleanly
+		})
+	})
+}
--- a/internal/hub/systems/system.go
+++ b/internal/hub/systems/system.go
@@ -623,10 +623,17 @@ func (sys *System) runSSHOperation(timeout time.Duration, retries int, operation
 			continue
 		}
 
-		retry, opErr := func() (bool, error) {
+		// Bound the whole operation. A half-open TCP connection (a dead peer that
+		// never sends RST/FIN) or a wedged agent that accepts the session but
+		// never writes a response would otherwise block the read forever. Because
+		// StartUpdater runs update() synchronously on its ticker, that stalls the
+		// per-system updater indefinitely with no error and no re-dial until the
+		// hub is restarted (issue #2041). On timeout we tear down the connection
+		// so the blocked read unwinds and the system is re-dialed on the next tick.
+		retry, opErr := runWithTimeout(sshOperationTimeout, func() (bool, error) {
 			defer session.Close()
 			return operation(session)
-		}()
+		}, sys.closeSSHConnection)
 
 		if opErr == nil {
 			return nil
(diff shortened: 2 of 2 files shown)


henrygd (approved):
Looks good, thanks for your work!

Issue #1953, opened 2026-04-25
[Bug]: Container CPU% inflated ~3.5x for all containers inside rootless Podman pods


niclas197:
### Welcome!

- [x] I have read the [Documentation](https://beszel.dev/guide/getting-started)
- [x] I have checked the [Common Issues Guide](https://beszel.dev/guide/common-issues) and my problem was not mentioned there.
- [x] I have searched open and closed issues and my problem was not mentioned before.
- [x] I have verified I am using the latest version available. You can check the latest release [here](https://github.com/henrygd/beszel/releases).

### Component

Agent

### Problem Description

CPU% values for all containers inside a Podman pod are inflated by a 
consistent factor of ~3.5x compared to `podman stats`. Standalone containers 
(not in a pod) show accurate values.

I've tested this across multiple pods and images. The cpu-inflation affects all pod 
containers regardless of image, entrypoint, or container user.

Disclaimer: I've used Claude to investigate.

### Expected Behavior

CPU% in Beszel should match `podman stats` AVG CPU% for pod containers, 
as it does for standalone containers.

### Steps to Reproduce

1. Run beszel-agent as a rootless Podman container with cgroup delegation 
   enabled and the Podman socket mounted
2. Create a Podman pod with multiple containers (e.g. a Quadlet-based pod)
3. Compare Beszel CPU% with `podman stats` for containers inside the pod and outside

Example data:

| Container | In Pod | Beszel CPU% | `podman stats` AVG% | Factor |
|-----------|--------|-------------|----------------------|--------|
| paperless-broker | yes | 9.90% | 2.58% | 3.8x |
| paperless-db | yes | ~9% | 2.40% | 3.8x |
| karakeep | yes | 7.25% | 2.04% | 3.6x |
| karakeep-meilisearch | yes | 0.62% | 0.17% | 3.6x |
| paperless | yes | 1.55% | 0.62% | 2.5x |
| immich-server | yes | 0.95% | 0.28% | 3.4x |
| forgejo | no | 0.72% | 0.72% | 1.0x |
| navidrome | no | 0.26% | 0.26% | 1.0x |

Investigation ruled out:
- CPU quota (`cpu.max` is `max 100000` for all containers)
- cgroup runtime overhead (negligible, < 0.01% of total CPU time)
- Container user (root, named user, UID:GID all affected equally)
- Entrypoint (tini, /init, direct binary all affected equally)
- Core count (factor ~3.5 does not correspond to 6 cores)

[...]

PR #2131, opened 2026-07-12
fix(agent): use time-based CPU calc for Podman containers

### LINKED ISSUE #2049: [Bug]: Podman container CPU chart appears to show cgroup activity share rather than host CPU utilization
### Welcome!

### Component

Hub & Agent

### Problem Description

The “Podman CPU Usage” chart appears to show a percentage that is not comparable to the host-wide “CPU Usage” chart.

I found #1953, which looks related, but I think this is a separate UI / metric-semantics issue. #1953 is about inflated CPU values for containers inside Podman pods. This report is not primarily about Podman pods, and not primarily about disagreement with `podman stats`. It is about how the Podman container CPU percentage is presented.

On my Podman host, the host-wide CPU chart can show below 2% CPU usage, while the Podman CPU chart shows a container around 16.65%.

Screenshot showing the mismatch between host CPU usage and Podman container CPU usage for the same time range:

Image

Direct host checks showed the VPS was basically idle:

```text
load average: ~0.09, 0.16, 0.12
vmstat: mostly 90-100% idle
podman stats: umami around ~0.8% at steady state
```

After looking at the code and the Podman stats payload, I think the Podman chart is effectively showing something closer to a container’s share of Podman/cgroup-accounted CPU activity, not its share of total host CPU capacity.

Beszel calculates Linux container CPU here:

https://github.com/henrygd/beszel/blob/v0.18.7/internal/entities/container/container.go#L54-L64

```go
cpuDelta := s.CPUStats.CPUUsage.TotalUsage - prevCpuContainer
systemDelta := s.CPUStats.SystemUsage - prevCpuSystem
return float64(cpuDelta) / float64(systemDelta) * 100.0
```

For Podman, `cpu_stats.system_cpu_usage` is populated from `cgroups.SystemCPUUsage()`:

https://github.com/containers/podman/blob/6ea0abf2d330fe4d6f2714734796c055fd095b69/pkg/api/handlers/compat/containers_stats_linux.go#L89-L116

That function sums cgroup `cpu.stat usage_usec` values:

https://github.com/containers/podman/blob/6ea0abf2d330fe4d6f2714734796c055fd095b69/vendor/go.podman.io/common/pkg/cgroups/cgroups_linux.go#L441-L469

So the displayed value seems closer to:

```text
container_cpu_delta / cgroup_accounted_cpu_delta
```

rather than:

```text
container_cpu_delta / total_host_cpu_capacity_over_the_interval
```

The metric is still useful for attribution, but the current title, percent unit, and stacked chart presentation make it look like a container is consuming much more VPS CPU than it actually is.

### Expected Behavior

For a chart titled “Podman CPU Usage” with percent units, I would expect the values to be comparable to the host-wide CPU chart, or for the UI to clearly indicate that the value is a relative container CPU share rather than absolute host CPU utilization.

Possible fixes:

1. If the intended meaning is “percentage of host CPU capacity”, normalize Podman container CPU against elapsed wall-clock time and `online_cpus`, for example:

```text
container_cpu_delta / (read_time_delta * online_cpus) * 100
```

2. If the current calculation is intentional and mainly useful for attribution, rename or describe the chart as something like:

```text
Podman CPU Share
Container CPU Share
Share of Container CPU Activity
```

3. Avoid stacking the CPU chart, or make the tooltip clearly distinguish per-container values from the stacked total.

### Steps to Reproduce

1. Run Beszel agent on a Linux host using Podman.
2. Start one or more ordinary Podman containers.
3. Open the system page in Beszel.
4. Compare the host-wide “CPU Usage” chart with the “Podman CPU Usage” chart.
5. Observe that a container can show a noticeably higher percentage, for example around `16.65%`, while host CPU usage is below `2%`.
6. On the host, verify actual host load with commands such as:

```bash
uptime
vmstat 1 5
podman stats --no-stream
```

In my case, the host was mostly idle while the Podman chart visually suggested much higher container CPU usage.

### OS / Architecture

Linux / amd64

### Beszel version

0.18.7

### Installation method

Other (please describe above)

### Configuration

```yaml
Runtime: Podman (agent) Docker (hub)

Podman version observed:


podman version 5.7.0


Host has 2 vCPUs.

Relevant agent-side setup:

- Beszel agent runs as a Podman container.
- Podman socket is mounted for container stats.
```

### Hub Logs

```json
No relevant hub errors observed.
```

### Agent Logs

```shell
No relevant agent errors observed.
```

polter-rnd:
## 📃 Description

Podman populates `system_cpu_usage` from cgroup `cpu.stat` rather than `/proc/stat`, so it reflects only cgroup-tracked CPU activity instead of total host CPU capacity. This caused container CPU percentages to be inflated and incomparable to the host-wide CPU chart.

Fix uses elapsed wall-clock time × `online_cpus` as the denominator for Podman containers, matching the approach already used for Windows. Falls back to the existing `system_cpu_usage` path if `online_cpus` is absent.

Fixes #2049

### ➕ Added

- `OnlineCPUs` field to `CPUStats` struct (maps to `online_cpus` in the Docker/Podman stats API response)
- `CalculateCpuPercentPodman` method on `ApiStats` for time-based CPU calculation: `cpuDelta / (elapsedNs × onlineCPUs) × 100`

### 🔧 Fixed

- Podman container CPU % now reflects actual host CPU utilization rather than share of cgroup-accounted activity

### DIFF
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -535,11 +535,18 @@ func (dm *dockerManager) updateContainerStats(ctr *container.ApiInfo, cacheTimeM
 	// Get previous CPU values
 	prevCpuContainer, prevCpuSystem := dm.getCpuPreviousValues(cacheTimeMs, ctr.IdShort)
 
-	// Calculate CPU percentage based on platform
+	// Calculate CPU percentage based on platform.
+	// Podman reports system_cpu_usage from cgroup cpu.stat (not /proc/stat), so it reflects
+	// only cgroup-tracked activity rather than total host capacity. Use a time-based method
+	// instead so the result is comparable to host CPU utilization. See:
+	// https://github.com/henrygd/beszel/issues/2049
 	var cpuPct float64
 	if dm.isWindows {
 		prevRead := dm.lastCpuReadTime[cacheTimeMs][ctr.IdShort]
 		cpuPct = res.CalculateCpuPercentWindows(prevCpuContainer, prevRead)
+	} else if dm.usingPodman && res.CPUStats.OnlineCPUs > 0 {
+		prevRead := dm.lastCpuReadTime[cacheTimeMs][ctr.IdShort]
+		cpuPct = res.CalculateCpuPercentPodman(prevCpuContainer, prevRead)
 	} else {
 		cpuPct = res.CalculateCpuPercentLinux(prevCpuContainer, prevCpuSystem)
 	}
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -544,11 +544,18 @@ func (dm *dockerManager) updateContainerStats(ctr *container.ApiInfo, cacheTimeM
 	// Get previous CPU values
 	prevCpuContainer, prevCpuSystem := dm.getCpuPreviousValues(cacheTimeMs, ctr.IdShort)
 
-	// Calculate CPU percentage based on platform
+	// Calculate CPU percentage based on platform.
+	// Podman reports system_cpu_usage from cgroup cpu.stat (not /proc/stat), so it reflects
+	// only cgroup-tracked activity rather than total host capacity. Use a time-based method
+	// instead so the result is comparable to host CPU utilization. See:
+	// https://github.com/henrygd/beszel/issues/2049
 	var cpuPct float64
 	if dm.isWindows {
 		prevRead := dm.lastCpuReadTime[cacheTimeMs][ctr.IdShort]
 		cpuPct = res.CalculateCpuPercentWindows(prevCpuContainer, prevRead)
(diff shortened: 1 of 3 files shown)


svenvg93:
Does this also fixes #1953 by any chance?

polter-rnd:
@svenvg93 I did not explicitly test this scenario, but I expect it to be fixed as well. You can try these binaries built with the fix: https://github.com/polter-rnd/beszel/releases/tag/v0.18.8

henrygd (approved):
Thank you! I'll get it in the release today. I did add a small guard re: #2205.

PR #2132, opened 2026-07-13
feat(site): add 15 and 30 minute chart ranges

fishandsheep:
AI-assisted implementation.

## Summary
- Add 15-minute and 30-minute chart ranges backed by stored one-minute metrics.
- Localize both range labels across all supported locales.
- Hide average/maximum controls for one-minute-backed ranges.
- Refresh selected time-range labels immediately after a language change.

## Validation
- `bun run build`
- `bunx biome check src/components/charts/chart-time-select.tsx`
- `git diff --check`


svenvg93:
Thanks for your work, much appreciated! Just one small note, could you omit the changes to the locale files from the PR?
https://beszel.dev/guide/developer-guide#exclude-generated-locale-files

PR #2138, opened 2026-07-15
fix: add fallback for CPU model detection on MIPS

JanDziaslo:
## 📃 Description

gopsutil's `cpu.Info()` does not parse the `cpu model` field from `/proc/cpuinfo`, which is the only source of CPU model names on MIPS. This PR adds a fallback that reads `/proc/cpuinfo` directly and combines `cpu model` (e.g. `MIPS 1004Kc V2.15`) with `system type` (e.g. `MediaTek MT7621 ver:1 eco:3`) into a complete identifier.

The fallback only triggers when gopsutil returns an empty ModelName, so x86/ARM/other architectures are unaffected.

## 📖 Documentation

N/A

### ➕ Added

- Fallback CPU model detection for MIPS architectures via direct `/proc/cpuinfo` parsing
- `getCpuModelFromCpuinfo()` and `parseCpuModel()` functions in `agent/system.go`
- Tests for MIPS cpuinfo parsing (8 test cases)

### ✏️ Changed

- `agent/fingerprint.go`: use the same fallback in `generateFingerprint()`

### 🔧 Fixed

- Empty CPU model name on MIPS devices (e.g. ASUS RT-AX53U with MediaTek MT7621)

### 🗑️ Removed

N/A

## 📷 Screenshots

N/A — no UI changes


rickydebojeet:
Waiting for this to get merged!

henrygd (approved):
Looks good, thanks very much for your work!

PR #2153, opened 2026-07-23
fix(agent): remove unsynchronized hubVersions cache causing concurrent map write panic

### LINKED ISSUE #2128: [Bug] Agent crash: fatal error: concurrent map writes in getHubVersion (unsynchronized hubVersions map)
### Component

Agent

### Problem Description

The agent crashes with `fatal error: concurrent map writes` in `(*Agent).getHubVersion` when the hub opens multiple SSH sessions concurrently (legacy/SSH polling mode). In my case an unstable network path (Tailscale over CGNAT) caused the hub to reconnect frequently; when several sessions arrived in the same second, the agent panicked. It crashed twice within a minute (systemd restarted it each time).

Looking at `agent/server.go` in v0.18.7, `hubVersions` is a package-level map that is lazily initialized and written without any synchronization:

```go
func (a *Agent) getHubVersion(sessionId string, sessionCtx ssh.Context) semver.Version {
	if hubVersions == nil {
		hubVersions = make(map[string]semver.Version, 1)
	}
	hubVersion, ok := hubVersions[sessionId]
	...
	hubVersions[sessionId] = hubVersion  // server.go:118 — concurrent write
	return hubVersion
}
```

`handleSession` runs in a separate goroutine per SSH session (gliderlabs/ssh `session.go:260`), so two simultaneous sessions race on the map. A `sync.Mutex`/`sync.Map` (or keying per-connection instead of globally) would fix it.

Secondary observation: since session IDs are unique per session, entries are never evicted, so the map also grows unboundedly on long-running agents with frequent hub reconnects.

### Expected Behavior

Concurrent hub SSH sessions should not crash the agent.

### Steps to Reproduce

1. Run hub + agent in legacy SSH polling mode (agent without working WebSocket connection).
2. Have the hub open several SSH sessions to the agent at nearly the same time (easily triggered by a flaky network path causing reconnect churn; likely reproducible with parallel manual SSH connections using the hub key).
3. Agent panics with `concurrent map writes` in `getHubVersion`.

### OS / Architecture

Debian 12 / aarch64 (Le Potato)

### Beszel version

0.18.7 (hub and agent)

### Installation method

Binary

### Configuration

Agent as systemd service (binary install script). Hub polls agent over Tailscale in legacy SSH mode on port 45876.

### Agent Logs

```
2026/07/09 07:29:50 INFO SSH connected addr=100.79.78.16:38510
2026/07/09 07:29:57 INFO SSH connected addr=100.79.78.16:43662
2026/07/09 07:29:57 INFO SSH connected addr=100.79.78.16:38566
2026/07/09 07:29:57 INFO SSH connected addr=100.79.78.16:38554
fatal error: concurrent map writes
goroutine 4345952 [running]:
internal/runtime/maps.fatal({0x497d90?, 0x1e8d88?})
	/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.1.linux-amd64/src/runtime/panic.go:1181 +0x20
github.com/henrygd/beszel/agent.(*Agent).getHubVersion(0x2378ee716d48?, {0x2378ee9b6900, 0x40}, {0x4c5b00, 0x2378ee929500})
	/home/runner/work/beszel/beszel/agent/server.go:118 +0x1b4
github.com/henrygd/beszel/agent.(*Agent).handleSession(0x2378ee700008, {0x4c5fd8, 0x2378ee5c51e0})
	/home/runner/work/beszel/beszel/agent/server.go:132 +0x7c
github.com/gliderlabs/ssh.(*session).handleRequests.func1()
	/home/runner/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.8/session.go:261 +0x34
created by github.com/gliderlabs/ssh.(*session).handleRequests in goroutine 4345949
	/home/runner/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.8/session.go:260 +0x4ac
beszel-agent.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
```

(Second crash 28 seconds later had an identical stack trace.)

Note: #221 reported the same fatal error message in 2024, but that was a different code path (`getDockerStats`) and was fixed via #216 — this one is in `getHubVersion`.

svenvg93:
## 📃 Description

`getHubVersion` cached the parsed hub SSH client version in a package-level `map[string]semver.Version`, keyed by session ID, written without any synchronization. Since `handleSession` runs in its own goroutine per SSH session, two sessions arriving close together (e.g. during hub reconnect churn on a flaky network) race on the map write and crash the agent with `fatal error: concurrent map writes`.

The cache also provided no benefit: session IDs are unique per session and `getHubVersion` is only ever called once per session, so no entry was ever re-read — it just grew unbounded for the lifetime of the agent process.

This removes the shared map entirely and extracts the hub version directly from the SSH client version string on each call, eliminating both the race and the leak.

Fixes #2128

### 🔧 Fixed

- Agent could crash with `fatal error: concurrent map writes` in `getHubVersion` when the hub opened multiple SSH sessions concurrently (e.g. legacy/SSH polling mode with frequent reconnects).

### 🗑️ Removed

- Unbounded, session-keyed `hubVersions` cache in the agent's SSH server — it was never actually hit on repeat lookups and grew indefinitely on long-running agents with frequent hub reconnects.

### DIFF
--- a/agent/server.go
+++ b/agent/server.go
@@ -29,9 +29,6 @@ type ServerOptions struct {
 	Keys    []gossh.PublicKey // SSH public keys for authentication
 }
 
-// hubVersions caches hub versions by session ID to avoid repeated parsing.
-var hubVersions map[string]semver.Version
-
 // StartServer starts the SSH server with the provided options.
 // It configures the server with secure defaults, sets up authentication,
 // and begins listening for connections. Returns an error if the server
@@ -99,24 +96,15 @@ func (a *Agent) StartServer(opts ServerOptions) error {
 	return a.server.Serve(ln)
 }
 
-// getHubVersion retrieves and caches the hub version for a given session.
-// It extracts the version from the SSH client version string and caches
-// it to avoid repeated parsing. Returns a zero version if parsing fails.
-func (a *Agent) getHubVersion(sessionId string, sessionCtx ssh.Context) semver.Version {
-	if hubVersions == nil {
-		hubVersions = make(map[string]semver.Version, 1)
-	}
-	hubVersion, ok := hubVersions[sessionId]
-	if ok {
-		return hubVersion
-	}
-	// Extract hub version from SSH client version
+// getHubVersion extracts the hub version from the SSH client version string
+// for a given session. Returns a zero version if parsing fails.
+func (a *Agent) getHubVersion(sessionCtx ssh.Context) semver.Version {
 	clientVersion := sessionCtx.Value(ssh.ContextKeyClientVersion)
 	if versionStr, ok := clientVersion.(string); ok {
-		hubVersion, _ = extractHubVersion(versionStr)
+		hubVersion, _ := extractHubVersion(versionStr)
+		return hubVersion
 	}
-	hubVersions[sessionId] = hubVersion
-	return hubVersion
+	return semver.Version{}
 }
 
 // handleSession handles an incoming SSH session by gathering system statistics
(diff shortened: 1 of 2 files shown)


henrygd:
Thanks Sven!!

PR #2159, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published images

kobihikri:
Hi, and thanks for Beszel.

`.github/workflows/docker-images.yml` publishes to both Docker Hub and `ghcr.io`, but the pushed manifests carry no provenance or SBOM attestation. Someone pulling either the hub or the agent image cannot check that it was built by this workflow, from this repository, at that tag.

Beszel's agent is installed on every machine being monitored, which makes it one of the more widely-placed images a self-hoster runs. That breadth is the argument for wanting its origin to be verifiable.

The change is two lines on the build step:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so a single build covers both registries it pushes to. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats: `mode=max` records build arguments (`provenance: true` gives a smaller record if any have been sensitive), and attestations add an extra manifest to the index, which both Docker Hub and GHCR support.

No SLSA level claimed.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.


henrygd (approved):
Great, thank you!

PR #2167, opened 2026-07-30
feat(ui): persist theme preference in user settings

milanobrtlik:
## 📃 Description

Addresses the theme half of #2142.

The theme was only stored in `localStorage`, so it was lost in private windows, on other devices, and after clearing site data. It is now persisted to `user_settings` alongside the other display preferences, so it follows the account after login.

`localStorage` is still written, because the inline script in `index.html` needs it to set the theme class before React mounts, and because the toggle has to keep working on the login page where there is no account yet.

No migration is needed — `user_settings.settings` is a free-form JSON field.

Rebased onto current `main`. The only conflict was the `./utils` import in `lib/api.ts`, which #2310 had also touched; resolved to `{ chartTimeData, debounce, THEME_STORAGE_KEY }`.

A few details worth flagging for review:

- The stored value is only applied when it is defined, so logging out (which resets user settings) does not reset the theme.
- Users who already picked a theme get it copied to the database once on load, so an existing preference is not silently lost.
- `saveSettings` in the settings layout is now a toast wrapper around a new `saveUserSettings` in `lib/api.ts`. The toggle needs to save without a toast, and importing from the lazily loaded settings route would have pulled it into the main bundle.

## On verification

There are no automated tests covering this, and the behaviour it exists for — surviving a private window or a different device — cannot be checked from the source alone, so I ran it against a real hub build and checked each path end to end:

| scenario | result |
| --- | --- |
| toggle the theme in the UI | `"theme": "light"` written to `user_settings` |
| empty `localStorage` + valid session (new device, private window) | theme read from the database, applied, and written back to `localStorage` |
| theme only in `localStorage`, absent from the database (pre-existing user) | database seeded once from the local value; the choice is not lost |
| log out | theme retained, confirming the `undefined` guard behaves as intended |

One caveat I should flag rather than leave to be discovered: with an empty `localStorage` there is a brief flash of the default theme before `user_settings` arrives over the network. The inline script in `index.html` runs before there is anything to read, so this is inherent to fetching the preference remotely rather than something introduced here — but it is visible, and worth knowing about. Writing back to `localStorage` means it only happens on the first load on a given device.

Note: `internal/site/package.json` has no `check`/`check:fix` script, so I ran the local Biome binary directly over the touched files, plus the full `bun run build`. Biome reports a formatter error in `lib/utils.ts` (the favicon SVG, lines 112-126) and an ineffective `biome-ignore` in `settings/layout.tsx:100`; both reproduce on an unmodified `main` and are untouched here. `lingui extract` produced no catalog changes.

### ➕ Added

- Theme preference is saved to user settings, so it persists in private windows and across browsers.

### DIFF
--- a/internal/site/src/components/routes/settings/layout.tsx
+++ b/internal/site/src/components/routes/settings/layout.tsx
@@ -14,7 +14,7 @@ import { lazy, useEffect } from "react"
 import { $router } from "@/components/router.tsx"
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card.tsx"
 import { toast } from "@/components/ui/use-toast.ts"
-import { pb } from "@/lib/api"
+import { saveUserSettings } from "@/lib/api"
 import { $userSettings } from "@/lib/stores.ts"
 import type { UserSettings } from "@/types"
 import { Separator } from "../../ui/separator"
@@ -36,18 +36,7 @@ const HeartbeatSettings = lazy(heartbeatSettingsImport)
 
 export async function saveSettings(newSettings: Partial<UserSettings>) {
 	try {
-		// get fresh copy of settings
-		const req = await pb.collection("user_settings").getFirstListItem("", {
-			fields: "id,settings",
-		})
-		// update user settings
-		const updatedSettings = await pb.collection("user_settings").update(req.id, {
-			settings: {
-				...req.settings,
-				...newSettings,
-			},
-		})
-		$userSettings.set(updatedSettings.settings)
+		await saveUserSettings(newSettings)
 		toast({
 			title: t`Settings saved`,
 			description: t`Your user settings have been updated.`,
--- a/internal/site/src/components/theme-provider.tsx
+++ b/internal/site/src/components/theme-provider.tsx
@@ -1,6 +1,10 @@
+import { useStore } from "@nanostores/react"
 import { createContext, useContext, useEffect, useState } from "react"
+import { pb, saveUserSettings } from "@/lib/api"
+import { $userSettings } from "@/lib/stores"
+import { THEME_STORAGE_KEY } from "@/lib/utils"
 
-type Theme = "dark" | "light" | "system"
+export type Theme = "dark" | "light" | "system"
 
 type ThemeProviderProps = {
 	children: React.ReactNode
@@ -23,10 +27,11 @@ const ThemeProviderContext = createContext<ThemeProviderState>(initialState)
 export function ThemeProvider({
 	children,
 	defaultTheme = "system",
-	storageKey = "ui-theme",
+	storageKey = THEME_STORAGE_KEY,
 	...props
 }: ThemeProviderProps) {
 	const [theme, setTheme] = useState<Theme>(() => (localStorage.getItem(storageKey) as Theme) || defaultTheme)
+	const { theme: savedTheme } = useStore($userSettings, { keys: ["theme"] })
 
 	useEffect(() => {
 		const root = window.document.documentElement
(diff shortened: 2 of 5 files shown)


svenvg93:
Hi! 

Thanks for your work, out of interest; would it worth it to combine it with the open PR for the other changes in #1831. 
Now there are two PRs with the same idea with different settings that are saved. If you want to add the other settings to your PR, I will close mine.

milanobrtlik:
Hi Sven, thanks for the offer, and for taking it so lightly given your PR has been open much longer than mine.

I would rather not move your work into mine. You have four months in #1831 and it should stay yours. But I agree that two PRs with the same idea is not useful for review, so let me suggest a different shape: treat them as one thing in two parts, not as two competing PRs.

#1831 is the base. It covers most of the issue, so it should keep `closes #2142`. Mine only adds the theme on top. The two only really overlap in `saveSettings` in `settings/layout.tsx`, so whichever lands first, I am happy to rebase the other one over it. Nothing you need to do right now.

On that overlap: you add a `quiet` parameter, while I moved the body into a new `saveUserSettings` in `lib/api.ts` and left `saveSettings` as a thin wrapper that only adds the toast. Please take that part from my branch if you like it, I think it solves two things in yours:

- With `quiet`, failed saves disappear completely. The `console.error` in the catch block is commented out, so a silent save that fails leaves no trace at all, not even in the console. `saveUserSettings` just throws and lets the caller decide, e.g. `.catch(console.error)` for the silent ones.
- `systems-table.tsx` and `use-system-data.ts` import `saveSettings` from `@/components/routes/settings/layout`, which pulls the settings route module (Card, toast, lingui macros, sidebar-nav) into the main bundle. That was my reason for putting the function in `lib/api.ts` instead.

Three more things I noticed while reading the diff. Take them or leave them:

1. The localStorage writes are gone. `useBrowserStorage` wrote the value back on every change; the new `useState` + `JSON.parse(localStorage.getItem(...))` only reads. `updateUserSettings()` in `main.tsx` is not awaited before render, so a component mounts with the localStorage value and then jumps to the server value when it arrives. Without the write, that fallback stays frozen at whatever it was before this lands, so the jump happens on every page load instead of once. Keeping the write would avoid it.

2. `sortMode` was in `sessionStorage` on purpose (`useBrowserStorage<SortingState>("sortMode", ..., sessionStorage)`) and becomes permanent and account-wide here. `statusFilter` is similar: someone who filters to "down" once will get an incomplete list on their next login, possibly on another device, with no obvious cause. Both may be exactly what you intend, but since they change behaviour that looks deliberate, it may be worth stating them explicitly in the description so henrygd does not have to infer it from the diff.

3. Minor: `saved` is in the dependency array of the large `useMemo` in systems-table, so the toolbar re-renders twice per change.

If you would rather do it the other way round, just say so.

svenvg93:
Hi Milan, thanks for taking the time to write that up, those were solid points, especially the `saveUserSettings` split and the local storage write-back regression.

I've updated my branch to match:

- Moved the save logic into `saveUserSettings` in `lib/api.ts`, with `saveSettings` in `layout.tsx` now just a thin toast wrapper around it, same as yours. The silent callers (`systems-table.tsx`, `use-system-data.ts`) now call `saveUserSettings` directly with `.catch(console.error)` instead of routing through the settings-route wrapper, so no bundle pull-in and no more silently swallowed failures.
- Restored the `localStorage`/`sessionStorage` write-back in the `cols`/`viewMode`/`sortMode`/`grid`/`displayMode` handlers that `useBrowserStorage` used to do — good catch, that was a real regression.

Since we're both on `saveUserSettings` now, whichever of us lands first, the other should rebase cleanly on top — `theme` and my other settings fields are just separate additions to `UserSettings`. Let me know if anything still looks off.

PR #2179, opened 2026-08-03
feat: Add cumulative disk read/write totals to Disk I/O sheet

### LINKED ISSUE #2174: [Feature]: Disk I/O cumulative read/write
### Welcome!

### Component

Hub

### Description

In the "view more" menu of disk I/O, it would be helpful to view cumulative read and cumulative write of a disk. This would be very similar to the cumulative upload and cumulative download already available in the bandwidth "view more" menu.

### Motivation / Use Case

This would make it easier to estimate progress on a large read operation. For example, a borg check operation which reads the entire borg archive.

svenvg93:
## 📃 Description

Adds cumulative disk read/write totals to the Disk I/O "view more" sheet, mirroring the existing "Cumulative Upload/Download" charts already available for network bandwidth. Closes #2174.

The agent already tracked cumulative read/write byte counters internally (used to compute the per-interval read/write rate) but never sent them to the hub. This change exposes those totals over the wire and renders them as two new charts.

### ➕ Added

- Cumulative Read / Cumulative Write charts in the Disk I/O "view more" sheet, for both the root filesystem and any configured extra filesystems.
- `Stats.DiskIOTotal` field (agent → hub) with cumulative read/write bytes for the root disk, and `FsStats.TotalRead`/`TotalWrite` are now serialized for extra filesystems.

### ✏️ Changed

- Hub's stats averaging (`AverageSystemStatsSlice`) treats the new cumulative fields like the existing cumulative network totals: takes the `max` across the aggregation window instead of averaging/summing.

## 📷 Screenshots

Scherm­afbeelding 2026-08-03 om 20 19 58


henrygd (changes requested):
Thank you Sven! Looks good overall. However, I think `ReadBytes` / `WriteBytes` from `disk.IOCounters()` are cumulative since boot, not since the agent started.

Could we change the comments and chart descriptions to something like "Cumulative data read" / "Cumulative data written” instead? I don't think we necessarily need to mention 'since boot' either, because it should be fairly obvious when it resets to zero.

Otherwise everything looks good to me.

svenvg93:
> Thank you Sven! Looks good overall. However, I think `ReadBytes` / `WriteBytes` from `disk.IOCounters()` are cumulative since boot, not since the agent started.
> 
> Could we change the comments and chart descriptions to something like "Cumulative data read" / "Cumulative data written” instead? I don't think we necessarily need to mention 'since boot' either, because it should be fairly obvious when it resets to zero.
> 
> Otherwise everything looks good to me.

yeah make sense, I change the description to the suggestion :)

henrygd (approved):
Gracias. I did make a few small changes. Moved the charts to the bottom of the sheet, synced the tooltip, and added "since boot" to the description after thinking about it a bit more.

aProfessionalFrog:
Huge thanks for this guys!

PR #2180, opened 2026-08-03
fix(agent): read /proc/uptime on linux instead of sysinfo(2)

### LINKED ISSUE #1948: [Bug]: LXC Containers show uptime of host
### Welcome!

### Component

Hub

### Problem Description

When the agent is intalled on LXC containers hosted on proxmox, on the dashboard, the reported uptime is that of the host and not the container:

Image

Image

Image

### Expected Behavior

Should be the uptime of the container, rather than the host

### Steps to Reproduce

Install an agent on a proxmox host and view the uptime column on the hub dashboard. Compare output of the: `uptime` command, executed on the host and the lxc

### OS / Architecture

Debian Linux 13 on Hub and Agent LXC Proxmox 8.4.19 on the Host

### Beszel version

0.18.7

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

alecrubin:
Fixes #1948.

`host.Uptime()` is `unix.Sysinfo()` in gopsutil, and lxcfs can't intercept a syscall, so every LXC container reports the host's uptime. `/proc/uptime` is virtualized correctly, it just wasn't being read.

This reads `/proc/uptime` on linux and falls back to `host.Uptime()` if the file is missing or unparseable. Other platforms go through a stub that calls `host.Uptime()` exactly as before, so nothing changes off linux.

Built a small binary that calls both ways and ran it on my Proxmox 9 box (lxcfs 7.0.0-pve1), containers had been up about an hour:

```
=== PVE host
host.Uptime()  [sysinfo(2), current] :   268189 s
/proc/uptime   [this PR]             :   268188 s

=== inside CT 404
host.Uptime()  [sysinfo(2), current] :   268191 s
/proc/uptime   [this PR]             :     3880 s

=== inside CT 403
host.Uptime()  [sysinfo(2), current] :   268192 s
/proc/uptime   [this PR]             :     3886 s
```

Same on bare metal, correct in containers.

Tests cover the parse (typical, zero, no trailing newline, single field, large value) and the three fallback paths (missing file, empty file, unparseable). `uptimeFilePath` is a package var so the tests can point it at a fixture, same as `mdraidSysfsRoot` does.

Ran `go build` and `go vet` for both linux and darwin, and the tests in a golang:1.26 container.

### DIFF
--- a/agent/system.go
+++ b/agent/system.go
@@ -256,7 +256,7 @@ func (a *Agent) getSystemStats(cacheTimeMs uint16) system.Stats {
 	a.systemInfo.MemPct = systemStats.MemPct
 	a.systemInfo.DiskPct = systemStats.DiskPct
 	a.systemInfo.Battery = systemStats.Battery
-	a.systemInfo.Uptime, _ = host.Uptime()
+	a.systemInfo.Uptime, _ = getUptime()
 	a.systemInfo.BandwidthBytes = systemStats.Bandwidth[0] + systemStats.Bandwidth[1]
 	a.systemInfo.Threads = a.systemDetails.Threads
 
--- a/agent/uptime_linux.go
+++ b/agent/uptime_linux.go
@@ -3,6 +3,7 @@
 package agent
 
 import (
+	"math"
 	"os"
 	"strconv"
 	"strings"
@@ -32,7 +33,11 @@ func getUptime() (uint64, error) {
 		return host.Uptime()
 	}
 	seconds, err := strconv.ParseFloat(fields[0], 64)
-	if err != nil || seconds < 0 {
+	if err != nil ||
+		math.IsNaN(seconds) ||
+		math.IsInf(seconds, 0) ||
+		seconds < 0 ||
+		seconds >= 1<<64 {
 		return host.Uptime()
 	}
 	return uint64(seconds), nil
--- a/agent/uptime_linux.go
+++ b/agent/uptime_linux.go
@@ -1,0 +1,39 @@
+//go:build linux
+
+package agent
+
+import (
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/shirou/gopsutil/v4/host"
+)
+
+// uptimeFilePath is a variable so tests can point it at a fixture.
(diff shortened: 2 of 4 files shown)


henrygd (changes requested):
Thanks a lot, this looks like a good approach.

One change before merge: validate the parsed float before converting it. `strconv.ParseFloat` accepts `NaN`, infinities, and values beyond the `uint64` range without returning an error, which can yield a huge bogus uptime instead of using the fallback.

```go
seconds, err := strconv.ParseFloat(fields[0], 64)
if err != nil ||
	math.IsNaN(seconds) ||
	math.IsInf(seconds, 0) ||
	seconds < 0 ||
	seconds >= 1<<64 {
	return host.Uptime()
}
return uint64(seconds), nil
```

And if possible, please add tests covering `NaN`, `+Inf`, negative values, and an out-of-range value such as `1e20`.

alecrubin:
> Thanks a lot, this looks like a good approach.
> 
> One change before merge: validate the parsed float before converting it. `strconv.ParseFloat` accepts `NaN`, infinities, and values beyond the `uint64` range without returning an error, which can yield a huge bogus uptime instead of using the fallback.
> 
> ```go
> seconds, err := strconv.ParseFloat(fields[0], 64)
> if err != nil ||
> 	math.IsNaN(seconds) ||
> 	math.IsInf(seconds, 0) ||
> 	seconds < 0 ||
> 	seconds >= 1<<64 {
> 	return host.Uptime()
> }
> return uint64(seconds), nil
> ```
> 
> And if possible, please add tests covering `NaN`, `+Inf`, negative values, and an out-of-range value such as `1e20`.

Thanks for the review. Done in 666b6f61b6aee620c7284f345aeed77d9133685b, took your validation as-is. Added fallback tests for `NaN`, `+Inf`, `-Inf`, a negative value, and `1e20`. `ParseFloat` parses all of those without an error so each one hits its own guard instead of the `err != nil` path. Ran the tests in a golang:1.26 container, all passing.

henrygd (approved):
Perfect, thank you! Appreciate your work.

PR #2197, opened 2026-08-08
Add per-sensor temperature alert thresholds

VACInc:
## What changed

- keeps the existing any-sensor temperature alert behavior as the default
- adds an individual-sensors mode with selectable probe-specific sliders and numeric inputs
- persists an optional per-probe threshold map through the API, cache, and collection schema
- evaluates immediate and multi-minute alerts against each selected probe's own threshold
- identifies the triggering probe in notifications and clarifies per-probe active alerts
- adds migration, API, evaluator, regression, schema, UI, and catalog coverage

## Why

Different temperature probes have different safe operating ranges. A single shared threshold can be too noisy for one sensor while being too permissive for another.

## Validation

- `go test -tags=testing ./internal/alerts -run '^TestTemperatureAlertPerSensorThresholds' -count=1`
- `go test -tags=testing ./internal/alerts -run '^TestUserAlertsApi$' -count=1`
- `go test -tags=testing ./internal/alerts -run '^TestSystemAlertsOneMin$|^TestSystemAlertsTwoMin$' -count=1`
- `go test -tags=testing ./internal/hub -run '^TestCollectionRulesDefault$|^TestCollectionRulesShareAllSystems$' -count=1`
- `bunx biome check src/components/alerts/alerts-sheet.tsx src/components/active-alerts.tsx src/lib/alerts.ts src/types.d.ts --max-diagnostics=100`
- `bunx lingui compile`
- `npm run build`
- `codex review --uncommitted`, no correctness findings

The repository's aggregate alert test run can intermittently hit the existing asynchronous `System.StartUpdater` teardown nil dereference. All affected and neighboring test groups pass independently above.

### DIFF
--- a/internal/alerts/alerts.go
+++ b/internal/alerts/alerts.go
@@ -74,6 +74,7 @@ type SystemAlertData struct {
 	count        uint8
 	min          uint8
 	mapSums      map[string]float32
+	mapCounts    map[string]uint8
 	descriptor   string // override descriptor in notification body (for temp sensor, disk partition, etc)
 }
 
--- a/internal/alerts/alerts_api.go
+++ b/internal/alerts/alerts_api.go
@@ -19,11 +19,12 @@ func UpsertUserAlerts(e *core.RequestEvent) error {
 	userID := e.Auth.Id
 
 	reqData := struct {
-		Min       uint8    `json:"min"`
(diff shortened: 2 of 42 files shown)


VACInc:
Opened against upstream in error; this change belongs in our internal repository.

Issue #1711, opened 2026-02-03
[Feature]: Docker Contianer Image Update Checker


lh01-lft:
### Welcome!

- [x] I have searched open and closed feature requests to make sure this or similar feature request does not already exist.
- [x] This is a feature request, not a bug report or support question.

### Component

Hub

### Description

All running docker containers get indexed and their repo gets sequentially checked for newer images

<img width="1552" height="987" alt="Image" src="https://github.com/user-attachments/assets/5ae9ea6f-199c-47f1-b6e8-07244839ac84" />

### Motivation / Use Case

to monitor outdated components. Maybe also consider to build it for system services of linux?

whedro:
+1

That's a good idea... I use Beszel for light container monitoring, and I would appreciate this feature. I don't need to add additional monitoring with this functionality. ;)

zenfokus:
+1 

That's a great idea. It would also be useful to have regex filtering, like in WUD, so that you can specify certain tags and exclude others. Maybe using labels as well?

iammarxg:
+1

Would love this. Beszel in the past few months has added many features that I used to use Dozzle and Dockpeek for. Only thing it is missing now, is the ability to update containers. I currently use Dockpeek for updating containers. I have already retired Dozzle since I no longer have a need for it, thanks to all who contributed to Beszel's development.

Dockpeek implements the update feature in a way where you have to manually check for updates, and then manually update each container. It's a bit buggy though to be honest. Would be nice if you could setup automated update checks and then the option to either update automatically or manually.

brunob45:
I started looking into this, so I'll just dump what I found here.

In order to query the latest digest (image id) from the registry, one must first get an auth token, then query the http header of the relevant image

```sh
NAME="library/alpine"
TAG="latest"

TOKEN=$(curl -s "https://auth.docker.io/token?service=registry.docker.io&scope=repository:$NAME:pull" | jq -r .token)

curl -s -I -H "Authorization: Bearer $TOKEN" \
    https://registry-1.docker.io/v2/$NAME/manifests/$TAG \
    -o manifest.txt

DIGEST=$(cat manifest.txt | grep "docker-content-digest" | sed -e 's/^[^:]*:\s*//')
```

Then, we can compare with the digest of the current image. If it is different, an update is available.
Of course, different registries have different url and auth requirements, so it won't be as easy.

As far as I've seen, the agent only has access to the container's info, not the images themselves.
[...]

PR #2211, opened 2026-08-10
feat(agent): Add docker image update available flag

brunob45:
## 📃 Description

Query the local docker image digest and compare it to the latest entry on https://hub.docker.com. An update is available when the values don't match. Back-end implementation towards fixing #1711.

## 📖 Documentation

### ➕ Added

- Query the local image digest for each running container
- Query the remote image digest and compare
- Update indicator `↑` near the image name
- Dependencies:
  - `github.com/distribution/reference`

## 📷 Screenshots

image

### DIFF
--- a/agent/docker.go
+++ b/agent/docker.go
@@ -260,7 +260,7 @@ func (dm *dockerManager) checkImageUpdate(ctr *container.ApiInfo) bool {
 
 	// get registry digest
 	repoDigest, err := getRegistryDigest(registry, repository, tag)
-	if err != nil {
+	if repoDigest == "" || err != nil {
 		return false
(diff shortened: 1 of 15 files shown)


brunob45:
rebased branch onto main

henrygd (approved):
Thanks, merging now. Let me know if my changes broke anything on your end.

brunob45:
Thanks for merging my work! I've tested the latest commit, but I found that after updating a container, the "UpdateAvailable" flags takes a long time to turn off.
Instead of caching the boolean flag, would it be possible to cache the SHA key and compare more frequently with the current image instead?

Issue #1100, opened 2025-08-28
[Feature]: Consider making "All Systems" a paginated list


chrisdeeming:
### Component

Hub

### Describe the feature you would like to see

I'd like for there to be some kind of limit, or pagination, on the Hub homepage.

### Motivation / Use Case

This might be, for now, a little bit niche, but I have a large quantity of hosts that I am monitoring through Beszel and the page has a tendency to go blank and require a refresh after a period of time. I'm not certain, but I strongly suspect the realtime nature of the page might be causing memory issues. I'm currently monitoring 435 systems!

### Describe how you would like to see this feature implemented

Either:

1. A paginated list with an option of how many systems per page
2. Or a persistent filter to show only the top X hosts

Personally, I doubt I'd be navigating through too many pages deep, so arguably option 2 sounds the best. I'm mostly going to have the list sorted by CPU / memory usage so seeing, say, only the top 50 hosts seems fine, though I appreciate all of these solutions might have complexities with filtering, and ordering in realtime.

### Category

Metrics

### Affected Metrics

CPU

henrygd:
I reckon your suspicion is correct. The project was originally intended for hobbyists, so I didn't account for that amount of systems. I'll figure out a solution because at this point we've outgrown the original goal.

What sort of resource usage are you seeing with the hub having so many systems? Is it still within a reasonable range?

chrisdeeming:
Honestly Hank I feel like I've way over overspecced the server. It's minimal. Barely a difference from when I had just a couple. It's just the client side that's struggling on that one page.

henrygd:
Let me know if this is any better in 0.12.6. I made the table a virtualized list so it only needs to render the systems in view.

svenvg93:
+1 to make it paginated, It looks cleaner then an endless list of the systems.

chrisdeeming:
> Let me know if this is any better in 0.12.6. I made the table a virtualized list so it only needs to render the systems in view.

Yes it's way better. I've had the tab open for 12 hours or so at this point and it's still running nice.

From my perspective, I'd say the issue is closed. I understand why people may still want pagination but this solves the issue for me as the page doesn't crash, I can filter and sort and scroll the full list without issue.

henrygd:
Perfect, let me know if this reoccurs or you find any other performance issues.

I'll close the issue since the original problem seems to be resolved.

I agree a paged table might be slightly nicer than the scrolling list. It's not super high priority, but I'll play around with it a bit more.

PR #2216, opened 2026-08-13
Adjust containers table height properties

erusso7:
Remove max height restriction from containers table.

## 📃 Description

This PR just removes the restriction on the height of the containers-table. It's a bit annoying the double scroll (main page + container list)

#### Current version:
Screenshot 2026-08-12 at 16 45 48

#### Proposed version after this PR
Screenshot 2026-08-12 at 16 46 06


henrygd (commented):
Thanks, however we want to keep the tables virtualized or paged for performance reasons. See #1100.

PR #2223, opened 2026-08-16
feat(agent): monitor Intel Arc (xe) GPUs via nvtop

miloszkolber:
## Problem

Xe kernel driver is not supported by `intel_gpu_top`, so such GPUs report no data at all. An alternative, nvtop is not included in docker builds.

## Solution

- Detect xe devices via `DRIVER=xe` in `/sys/class/drm/card*/device/uevent`.
- Skip `intel_gpu_top` when an xe device is present and let the existing `nvtop` collector handle the GPU (nvtop >= 3.2 supports the xe driver).
- Add `nvtop` to the Intel agent image.

Behavior is unchanged when `GPU_COLLECTOR` is set explicitly. Auto-detection on xe systems now resolves to nvtop, equivalent to `GPU_COLLECTOR=nvtop`.

## Testing

- `go build`, `go test -tags='testing no_ui' ./agent/...`
- Cross-compiled the agent for darwin/arm64, freebsd/amd64, windows/amd64.
- Live-tested on an Arc Pro B50: with `pid: host`

## Notes

- nvtop reports aggregate GPU usage/temp/power/memory, not per-engine utilization.
- For Intel GPU utilization to show up, the agent container needs the host PID namespace (`pid: host`). Without it, nvtop reports 0% usage (temperature, power, and VRAM still work).
- On mixed systems (xe + NVIDIA/AMD), the primary vendor collector runs and nvtop does not, so the xe card may not be reported (nvtop is last-resort).

### DIFF
--- a/agent/gpu.go
+++ b/agent/gpu.go
@@ -122,6 +122,7 @@ type gpuCapabilities struct {
 	hasAmdSysfs     bool
 	hasTegrastats   bool
 	hasIntelGpuTop  bool
+	hasXe           bool
 	hasNvtop        bool
 	hasMacmon       bool
 	hasPowermetrics bool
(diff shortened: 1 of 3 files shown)


henrygd:
Thank you! Looks good, but just to clarify, did you verify that `nvtop` is working properly in Docker environment?

miloszkolber:
Yes, verified on my live docker setup with Arc Pro B50. Power consumption, usage, VRAM and temperature from `nvtop` are picked up properly.

henrygd (approved):
Great, thanks! I fixed a pretty simple merge conflict that shouldn't break anything. 

There was another intel GPU related change added earlier today, so let me know if anything is not working as intended after merge 👍

PR #2237, opened 2026-08-18
fix(hub): guard SSH client against concurrent close race

### LINKED ISSUE #2157: [Bug]: Hub can panic in createSessionWithTimeout during concurrent SSH client close
### Welcome

- [x] I have read the documentation.
- [x] I have checked the Common Issues Guide.
- [x] I searched open and closed issues for this stack trace and found no match.
- [x] I verified that `0.18.7` is the latest release.

### Component

Hub

### Problem Description

A production Beszel hub crashed with a nil pointer dereference while creating an SSH session. Docker restarted it automatically about seven seconds later.

The stack points to `createSessionWithTimeout`:

```text
2026-07-04T12:33:43.271945416Z panic: runtime error: invalid memory address or nil pointer dereference
2026-07-04T12:33:43.331368787Z [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7b962e]

goroutine 169920 [running]:
golang.org/x/crypto/ssh.(*Client).NewSession(...)
    /go/pkg/mod/golang.org/x/crypto@v0.49.0/ssh/client.go:135
github.com/henrygd/beszel/internal/hub/systems.(*System).createSessionWithTimeout.func1()
    /app/internal/hub/systems/system.go:675
created by github.com/henrygd/beszel/internal/hub/systems.(*System).createSessionWithTimeout
    /app/internal/hub/systems/system.go:674
```

The source appears to contain a time-of-check/time-of-use race:

1. `createSessionWithTimeout` checks `sys.client == nil` before starting a goroutine.
2. The goroutine later evaluates `sys.client.NewSession()`.
3. `closeSSHConnection` can concurrently call `sys.client.Close()` and set `sys.client = nil`.

Relevant source:

- `createSessionWithTimeout`: https://github.com/henrygd/beszel/blob/v0.18.7/internal/hub/systems/system.go#L661-L690
- `closeSSHConnection`: https://github.com/henrygd/beszel/blob/v0.18.7/internal/hub/systems/system.go#L692-L701

The same access pattern is still present on `main` at the time of filing. Capturing the client pointer before starting the goroutine, together with appropriate synchronization around replacement/close, may prevent the nil dereference. This root-cause assessment is an inference from the stack and current source.

### Expected Behavior

An agent disconnect or concurrent SSH connection close should make session creation return an error. It should not terminate the hub process.

### Steps to Reproduce

The crash is intermittent and I have not reproduced it deterministically:

1. Run the hub with multiple systems using SSH-connected agents.
2. Keep the hub running while agents disconnect/reconnect or SSH operations time out.
3. Concurrent session creation and connection cleanup can enter the race window described above.
4. Observe the hub panic in `ssh.(*Client).NewSession` and restart.

The production process had been running normally before this event. No database I/O or integrity error was present.

### OS / Architecture

Debian 13 / linux/amd64

### Beszel version

0.18.7

### Installation method

Docker

### Configuration

```yaml
services:
  beszel:
    image: henrygd/beszel:0.18.7
    restart: unless-stopped
    volumes:
      - ./data:/beszel_data
      - ./socket:/beszel_socket
```

### Hub Logs

Included above.

### Agent Logs

No corresponding agent-side panic was found.

svenvg93:
## 📃 Description

Fixes a nil pointer dereference panic in the hub caused by a time-of-check/time-of-use race on the SSH client. `createSessionWithTimeout` checked `sys.client == nil` and then, in a separate goroutine, read `sys.client` again to call `.NewSession()`. Meanwhile `closeSSHConnection()` — triggered independently from a different goroutine when a system is paused or deleted — could close and nil out `sys.client` in between those two reads, causing `ssh.(*Client).NewSession` to panic on a nil receiver and crash the hub process.

The fix adds a mutex guarding the `client` field and routes all reads/writes through new `getClient`/`setClient` helpers, and has `createSessionWithTimeout` capture a stable local reference before spawning its goroutine so the check-then-use is race-free.

## 📖 Documentation

N/A — internal bug fix, no user-facing behavior or docs changes.

### 🔧 Fixed

- Fixed a race condition where a concurrent SSH connection close (e.g. pausing or deleting a system while it was mid-update) could nil out the SSH client between a nil-check and its use, causing the hub to panic with a nil pointer dereference in `createSessionWithTimeout` (closes #2157).

## 📷 Screenshots

N/A — backend-only concurrency fix, no UI changes.

### DIFF
--- a/internal/hub/systems/system.go
+++ b/internal/hub/systems/system.go
@@ -9,6 +9,7 @@ import (
 	"math/rand"
 	"net"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -38,6 +39,7 @@ type System struct {
 	Port           string                  `db:"port"`
 	Status         string                  `db:"status"`
 	manager        *SystemManager          // Manager that this system belongs to
+	clientMu       sync.Mutex              // Guards client against concurrent close/replace
 	client         *ssh.Client             // SSH client for fetching data
 	sshTransport   *transport.SSHTransport // SSH transport for requests
 	data           *system.CombinedData    // system data from agent
@@ -434,7 +436,7 @@ func (sys *System) request(ctx context.Context, action common.WebSocketAction, r
 	err := sys.sshTransport.RequestWithRetry(ctx, action, req, dest, 1)
 	// Keep legacy SSH client/version fields in sync for other code paths.
 	if sys.sshTransport != nil {
-		sys.client = sys.sshTransport.GetClient()
+		sys.setClient(sys.sshTransport.GetClient())
 		sys.agentVersion = sys.sshTransport.GetAgentVersion()
 	}
 	return err
@@ -476,8 +478,8 @@ func (sys *System) ensureSSHTransport() error {
 		})
 	}
 	// Sync client state with transport
-	if sys.client != nil {
-		sys.sshTransport.SetClient(sys.client)
+	if client := sys.getClient(); client != nil {
+		sys.sshTransport.SetClient(client)
 		sys.sshTransport.SetAgentVersion(sys.agentVersion)
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


henrygd (commented):
Thank you Sven. This addresses the same issue as #2277, and after comparing the two approaches, I’m going to go with #2277 because it’s a smaller change and preserves the existing client reset behavior on failed SSH dials. Appreciate your work in any case!

PR #2238, opened 2026-08-18
fix(agent): discover fans on legacy hwmon parent devices

IlyaMur:
## 📃 Description

Fix Linux fan discovery for legacy hwmon drivers that register a hwmon class device but expose `fan*_input`, labels, and the chip name on the parent device instead of directly under `/sys/class/hwmon/hwmonN`.

This affects `applesmc` on Intel Macs. Beszel 0.18.8 sees the hwmon device, but misses the fan because the RPM file is exposed on the parent platform device.

The change keeps the existing discovery path as-is and only falls back to `hwmonN/device` when no `fan*_input` files are present in `hwmonN` itself. Labels and chip names are then read from the same directory as the discovered fan inputs.

Tested on real hardware:

- MacBook Air 13-inch Mid 2013 (`MacBookAir6,2`)
- Ubuntu Server 26.04 LTS
- Linux 7.0.0-29-generic
- `applesmc`

Before the fix, Beszel showed no fan data even though Linux reported the fan at `/sys/devices/platform/applesmc.768/fan1_input`.

After the fix, the agent reports the sensor as `applesmc_Exhaust` and the fan chart updates correctly in the Beszel UI.

A regression test covering this legacy `applesmc` layout is included.

### 🔧 Fixed

- Discover fan sensors exposed on legacy hwmon parent devices such as `applesmc`.
- Read fan labels and chip names from the same sysfs directory used for the discovered fan inputs.
- Add test coverage for the legacy `applesmc` hwmon layout.

## 📷 Screenshots

No UI code changes. The fix was verified end-to-end on a real 2013 MacBook Air, where the `applesmc_Exhaust` fan appears and reports live RPM in the existing Fans chart.

image

### DIFF
--- a/agent/fans.go
+++ b/agent/fans.go
@@ -72,14 +72,30 @@ func discoverHwmonFans(root string) ([]fanSensor, error) {
 	var sensors []fanSensor
 	for _, entry := range entries {
 		chipDir := filepath.Join(root, entry.Name())
-		chipName := utils.ReadStringFile(filepath.Join(chipDir, "name"))
+		sensorDir := chipDir
+		inputs, _ := filepath.Glob(filepath.Join(sensorDir, "fan*_input"))
+
+		// Some legacy hwmon drivers (notably applesmc) register a hwmon class
+		// device but create fan attributes on the parent platform device. In
+		// sysfs that parent is exposed through hwmonN/device.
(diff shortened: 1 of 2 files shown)


henrygd (approved):
Great, thanks a lot!

PR #2241, opened 2026-08-19
fix(agent): strip invalid UTF-8 from battery names

### LINKED ISSUE #2240: [Bug]: Invalid UTF-8 in a battery name stops all data collection for the system (hub reports a misleading EOF)
### Welcome!

### Component

Hub & Agent

### Problem Description

After the agent auto-updated to **0.18.8**, one system (a laptop, the only host in my fleet with a
battery) went permanently `down` and never recovered. Hub logs only this, once:

```json
{"err":"EOF","system":"<laptop>"}
```

The network is fine, and the agent is fine: the hub reaches port 45876 every minute (the agent logs
two `SSH connected` per minute, forever), and connecting manually from the hub host with the hub's
own key returns a complete payload. No stats row is ever written.

The `EOF` is misleading. The real error is `cbor: invalid UTF-8 string`, thrown away by a fallback
path in the hub.

**Root cause**

1. On this ThinkPad, the battery model name reported by firmware contains an invalid UTF-8 byte:

   ```
   # od -c /sys/class/power_supply/BAT0/model_name
   0000000   L   N   V   -   5   B   1   1   K   6   3   0   2   4   @ 320
   0000020  \n
   ```

   (trailing `0xD0`, a lone continuation byte — DMI junk, present since the machine was new)

2. The multi-battery monitoring added in 0.18.8 uses that string **verbatim** as a map key.
   `agent/battery/battery_linux.go` reads `model_name` with no validation, and
   `agent/system.go:138` does `systemStats.Batteries[device.Name] = device.Percent`
   (`Batteries map[string]uint8` → `cbor:"37,keyasint"`).

3. `fxamacker/cbor`'s **encoder does not validate UTF-8**, so the agent encodes and sends it happily,
   with no error on the agent side. The **decoder defaults to `UTF8RejectInvalid`**, so the hub's
   decode fails. Minimal proof:

   ```go
   b, err := cbor.Marshal(map[string]uint8{"LNV-5B11K63024@\xd0": 99}) // err == nil
   var m map[string]uint8
   err = cbor.Unmarshal(b, &m)                                        // cbor: invalid UTF-8 string
   ```

4. **The hub then masks that error.** In `fetchDataViaSSH` (`internal/hub/systems/system.go`), when
   the `AgentResponse` decode fails, execution falls through to a *second* decode against the same,
   already-drained `stdout`:

   ```go
   var resp common.AgentResponse
   if decErr := cbor.NewDecoder(stdout).Decode(&resp); decErr == nil && resp.SystemData != nil {
       ...
       return false, nil
   }
   // decErr is discarded here
   var decodeErr error
   if sys.agentVersion.GTE(beszel.MinVersionCbor) {
       decodeErr = cbor.NewDecoder(stdout).Decode(sys.data)  // stream already consumed -> EOF
   }
   ```

   So the logged error is always `EOF`, never the actual decode failure. With `retries = 1` this
   repeats twice per collection cycle, which is why the agent logs two `SSH connected` per minute
   forever.

This is the same failure class as #1755, in a new place.

**Note:** the SSH transport is only what makes the error *unreadable*. The WebSocket transport
(`internal/hub/transport/websocket.go`) also uses default-mode `cbor.Unmarshal`, so WS agents on
such a machine should fail too — presumably with the accurate error message.

**Why it is easy to miss:** connecting to the agent by hand with plain `ssh` and the hub's key
*works* and shows the bad byte, because OpenSSH's client version string is not `beszel_x.y.z`, so
the agent falls back to the legacy **JSON** path — and `encoding/json` silently replaces invalid
bytes with U+FFFD. Only the CBOR path the hub actually uses breaks.

```
"bats":{"LNV-5B11K63024@�":99}
```

### Expected Behavior

1. The agent should sanitize strings that come from firmware before putting them into a
   CBOR-encoded payload. `normalizeBatteries()` in `agent/battery/battery.go` looks like the
   natural place — it is already the shared name-normalization step for Linux, Windows and
   Darwin, so `strings.ToValidUTF8` there covers every platform in one line. One bad byte from
   firmware should not take a host off monitoring entirely.

   I have opened a PR for this part. Verified on the affected hardware: with the patched agent
   the system came back up 45 seconds after restart, having been down for 11 hours, and the
   name is now reported as `"bats":{"LNV-5B11K63024@":99}`.

2. The hub should not report `EOF` when the real error is a decode failure. In `fetchDataViaSSH`,
   the first decode error should be kept and returned/logged instead of falling through to a second
   decode of an already-consumed stream.

3. Optionally, hub-side decoders could use `UTF8DecodeInvalid` as defense in depth, so a single
   malformed string degrades one field instead of dropping the whole payload.

### Steps to Reproduce

On any Linux host, without needing a machine with a quirky battery:

1. Create a replacement model name containing an invalid UTF-8 byte and bind-mount it over the
   sysfs file the agent reads:

   ```sh
   printf 'FAKEBAT\xd0\n' | sudo tee /etc/fake_model_name > /dev/null
   sudo mount --bind /etc/fake_model_name /sys/class/power_supply/BAT0/model_name
   ```

   (On a host with no battery, bind-mount a fake `power_supply` tree containing
   `BAT0/{type,capacity,status,model_name}` over `/sys/class/power_supply` instead.)

2. Run agent 0.18.8 and add the system to a 0.18.8 hub in SSH pull mode.

3. Watch the hub: the system goes `down` within a minute with `{"err":"EOF"}`, and no
   `system_stats` record is ever written. The agent logs two `SSH connected` per minute and
   reports no error of its own.

4. Undo with `sudo umount /sys/class/power_supply/BAT0/model_name`.

### OS / Architecture

linux/amd64 (agent, Ubuntu 24.04), linux/amd64 (hub, Docker)

### Beszel version

0.18.8 (both hub and agent)

### Installation method

Binary (agent) / Docker (hub)

### Configuration

```yaml
# agent: /etc/systemd/system/beszel-agent.service (SSH pull mode, no token)
Environment="PORT=45876"
Environment="KEY=ssh-ed25519 AAAA..."
ExecStart=/opt/beszel-agent/beszel-agent
User=beszel
```

### Hub Logs

```json
{"created": "2026-08-18 17:15:55.980Z", "level": 4, "message": "Session closed. Retrying...", "data": {"err": "EOF", "host": "<agent-ip>", "port": "45876"}}
{"created": "2026-08-18 17:15:56.054Z", "level": 8, "message": "System down", "data": {"err": "EOF", "system": "<laptop>"}}
```

Nothing is logged after that — the failure repeats silently every minute.

### Agent Logs

```shell
# agent updated itself to 0.18.8 at 01:15:05 local time; the hub marked the system
# down 50 seconds later. Since then, twice a minute, forever, with no error:
Aug 19 11:06:55 beszel-agent[1162366]: 2026/08/19 11:06:55 INFO SSH connected addr=<hub-ip>:39112
Aug 19 11:06:56 beszel-agent[1162366]: 2026/08/19 11:06:56 INFO SSH connected addr=<hub-ip>:39124
Aug 19 11:07:55 beszel-agent[1162366]: 2026/08/19 11:07:55 INFO SSH connected addr=<hub-ip>:48710
Aug 19 11:07:56 beszel-agent[1162366]: 2026/08/19 11:07:56 INFO SSH connected addr=<hub-ip>:48722
```

toomore:
## 📃 Description

Fixes #2240.

Battery names come from firmware and are not guaranteed to be valid UTF-8. On a ThinkPad here,
`/sys/class/power_supply/BAT0/model_name` ends in a lone `0xD0`:

```
# od -c /sys/class/power_supply/BAT0/model_name
0000000   L   N   V   -   5   B   1   1   K   6   3   0   2   4   @ 320
0000020  \n
```

Since 0.18.8 that string is used verbatim as a key in `SystemStats.Batteries`. `fxamacker/cbor`'s
encoder does not validate UTF-8, so the agent sends it without complaint, but the decoder defaults
to `UTF8RejectInvalid`, so the hub cannot decode the payload. The result is that the system records
no stats at all and stays down permanently — downgrading the agent was the only way out.

This sanitizes the name in `normalizeBatteries()`, which is the shared normalization step for the
Linux, Windows and Darwin collectors, so a single line covers all of them. A name consisting only
of invalid bytes becomes empty and falls through to the existing `Battery N` fallback, so that case
stays sane too.

The added test fails before the change and passes after it.

Verified on the machine that triggers the bug: with the patched agent it came back up 45 seconds
after restart, having been down for 11 hours, and the battery is now reported as
`"bats":{"LNV-5B11K63024@":99}`.

The hub-side half of #2240 — `fetchDataViaSSH` decoding an already-drained stream a second time and
therefore logging `EOF` instead of the real decode error — is deliberately not touched here, since
that is your retry logic and there is more than one reasonable way to fix it. Happy to send a
follow-up if you would like it addressed.

### 🔧 Fixed

- Battery names containing invalid UTF-8 no longer stop the hub from decoding the agent's payload.

### DIFF
--- a/agent/battery/battery.go
+++ b/agent/battery/battery.go
@@ -33,7 +33,10 @@ var errNoBatteries = errors.New("no readable batteries")
 func normalizeBatteries(batteries []Battery) []Battery {
 	nameCounts := make(map[string]int, len(batteries))
 	for i := range batteries {
-		name := strings.TrimSpace(batteries[i].Name)
+		// Names come from firmware (e.g. sysfs model_name) and are not guaranteed to
+		// be valid UTF-8. Invalid bytes are rejected when the hub decodes the CBOR
+		// payload, which drops every metric for the system, so strip them here.
+		name := strings.ToValidUTF8(strings.TrimSpace(batteries[i].Name), "")
 		if name == "" {
 			name = "Battery " + strconv.Itoa(i+1)
 		}
--- a/agent/battery/battery.go
+++ b/agent/battery/battery.go
@@ -36,7 +36,7 @@ func normalizeBatteries(batteries []Battery) []Battery {
 		// Names come from firmware (e.g. sysfs model_name) and are not guaranteed to
 		// be valid UTF-8. Invalid bytes are rejected when the hub decodes the CBOR
 		// payload, which drops every metric for the system, so strip them here.
-		name := strings.ToValidUTF8(strings.TrimSpace(batteries[i].Name), "")
+		name := strings.TrimSpace(strings.ToValidUTF8(batteries[i].Name, ""))
 		if name == "" {
 			name = "Battery " + strconv.Itoa(i+1)
 		}
--- a/agent/battery/battery_test.go
+++ b/agent/battery/battery_test.go
@@ -2,6 +2,7 @@ package battery
 
 import (
 	"testing"
+	"unicode/utf8"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -32,4 +33,16 @@ func TestPrimarySelection(t *testing.T) {
 func TestNormalizeBatteriesFallbackNames(t *testing.T) {
 	bats := normalizeBatteries([]Battery{{}, {}, {Name: "Mouse"}, {Name: "Mouse"}})
 	assert.Equal(t, []string{"Battery 1", "Battery 2", "Mouse", "Mouse (2)"}, []string{bats[0].Name, bats[1].Name, bats[2].Name, bats[3].Name})
+}
+
+func TestNormalizeBatteriesStripsInvalidUTF8(t *testing.T) {
+	// Firmware occasionally reports names that are not valid UTF-8 (a ThinkPad
+	// reporting "LNV-5B11K63024@\xd0" in model_name is a real example).
+	bats := normalizeBatteries([]Battery{{Name: "LNV-5B11K63024@\xd0"}, {Name: "\xff\xfe"}})
+	assert.Equal(t, "LNV-5B11K63024@", bats[0].Name)
+	// A name made up entirely of invalid bytes falls back to the generic name.
+	assert.Equal(t, "Battery 2", bats[1].Name)
+	for _, b := range bats {
+		assert.True(t, utf8.ValidString(b.Name))
+	}
 }


henrygd (approved):
Thank you very much!

I did make a small change to swap order of ops like this, just in case there's whitespace between the name and invalid char. That ensures the whitespace is always trimmed.

```go
strings.TrimSpace(strings.ToValidUTF8(batteries[i].Name, ""))
```

Issue #1252, opened 2025-10-07
[Feature]: Installer should re-register a server


jimaek:
### Component

Hub

### Describe the feature you would like to see

If I run the the installer `curl -sL https://get.beszel.dev ` on a server that already has beszel installed it should 

1. Upgrade if not on latest version
2. De-register
3. Register with the new token

Currently I kept running into "fingerprint missmatch" errors and had to manually uninstall and install to fix that issue

### Motivation / Use Case

Simplifies debugging and problem fixing. Just re-run the universal adoption command and all previous problems will be fixed.

### Describe how you would like to see this feature implemented

Add additional checks to the installer to check if beszel is already installed and essentially uninstall and install again

### Category

Installation

### Affected Metrics

Other

henrygd:
Added to roadmap. If you run into this situation again, try rotating the fingerprint on `/settings/tokens`. This will allow the agent to register a new one.

Assuming you're on Linux, the agent should save the fingerprint in `/var/lib/beszel-agent` and always reuse it. If you continuously get `fingerprint mismatch` errors even with a saved token, let me know as that definitely shouldn't happen.

And if you ever want to uninstall an agent, the easiest way is to run the install script with the `-u` flag.

jimaek:
Thanks! I wasn't able to figure out the reason for the errors so I decided to reinstall. That's when this issue came up. I did use the -u option so it was simple :)

Would also be great to have universal tokens that never expire to use in deployment scripts

PR #2242, opened 2026-08-19
Apply -k, -t, -url and -p to an existing service configuration

mews-se:
## 📃 Description

Running the install command from the hub's "Add System" dialog on a host that already has an
agent installed leaves the old `TOKEN` in place, and the script still reports success.

Service files are only written when they don't exist yet:

```sh
if [ ! -f /etc/systemd/system/beszel-agent.service ]; then
  ...
  Environment="TOKEN=$TOKEN"
  ...
else
  echo "Systemd service file already exists. Skipping creation."
fi
```

That heredoc is the only place `TOKEN`, `KEY` and `HUB_URL` reach the service, so on an existing
install the flags are dropped. The script then restarts the agent and finishes with "Beszel Agent
has been installed successfully!", while the agent logs this every 10 seconds:

```
WARN WebSocket connection failed err="unexpected status code: 401"
```

Keeping the configuration on a plain upgrade looks deliberate — that's what "Upgrading existing
installation. Using existing service configuration." on line 562 is for. But that message only
prints when `-k` was omitted, and the Add System command always includes `-k` and `-t`, so the
user never sees it. From their side it looks like the hub is rejecting a valid token, and the
natural next step is to remove the system and add it again, which produces another token that is
ignored the same way plus a duplicate entry in the hub.

This adds `update_existing_config`, which writes the values that were actually passed into the
existing configuration and reports what changed:

```
Systemd service file already exists. Skipping creation.
Updated existing service configuration: TOKEN
```

Only the fields given on the command line are touched, and only the value part of each line, so
local customizations, comments and the `DeviceAllow` / sandboxing lines are preserved. An upgrade
run without flags behaves exactly as before. All four service formats are covered: systemd,
Alpine OpenRC, OpenWRT procd and the FreeBSD env file (which names the port `LISTEN` and quotes
`KEY` because it contains spaces).

`PORT` is handled separately. It has a default of `45876`, so it is never empty and cannot be
used to tell "flag given" from "flag omitted" — applying it unconditionally would overwrite a
custom port on every run. It is therefore only written when `-p` was actually passed.

Docker installs are unaffected; they configure the agent through environment variables and never
run this script.

Related: #1252 asks for full de-register/re-register on reinstall. This is narrower and doesn't
depend on that landing — it just stops the flags from being discarded without a word.

### Testing

- 22 unit test cases against fixtures in all four formats: values with spaces (`KEY`),
  sed-special characters (`& | \`), preserved customizations, a custom port surviving a run
  without `-p`, no-op when no flags are passed, no output when the value is unchanged, and a
  missing file.
- End-to-end on a systemd host with an agent already installed: ran the installer with a bogus
  token and confirmed the unit file was updated with the sandboxing lines untouched, then again
  with the correct token and confirmed the agent reconnected and the hub showed the system as up.
  Also set a custom port, ran the installer without `-p`, and confirmed the port survived.

### 🔧 Fixed

- Installing over an existing agent now applies `-k`, `-t`, `-url` and `-p` to the existing
  service configuration instead of silently ignoring them.

### DIFF
--- a/supplemental/scripts/install-agent.sh
+++ b/supplemental/scripts/install-agent.sh
@@ -97,6 +97,83 @@ ensure_trailing_slash() {
   fi
 }
 
+# Apply values passed on the command line to a service configuration that
+# already exists. Existing service files are never recreated, so -k, -t, -url
+# and -p would otherwise be dropped while the install still reported success,
+# leaving the agent running with a stale token.
+# Usage: update_existing_config <file> <systemd|openrc|procd|env>
+update_existing_config() {
+  config_file=$1
+  config_format=$2
+  config_changed=""
+
+  [ -f "$config_file" ] || return 0
+
+  for config_var in PORT KEY TOKEN HUB_URL; do
+    eval "config_value=\$$config_var"
+    [ -n "$config_value" ] || continue
+
+    # PORT always has a default value, so only apply it when -p was given
+    if [ "$config_var" = "PORT" ] && [ "$PORT_PROVIDED" != true ]; then
+      continue
+    fi
+
+    # the FreeBSD env file calls the port LISTEN
+    config_name=$config_var
+    if [ "$config_format" = "env" ] && [ "$config_var" = "PORT" ]; then
+      config_name="LISTEN"
+    fi
+
+    # escape characters that are special on the replacement side of sed
+    config_escaped=$(printf '%s' "$config_value" | sed 's/[\\&|]/\\&/g')
+
+    case "$config_format" in
+    systemd)
+      config_expr="s|^\(Environment=\"${config_name}=\)[^\"]*|\1${config_escaped}|"
+      ;;
+    openrc)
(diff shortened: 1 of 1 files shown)


svenvg93:
Is this not a dup of https://github.com/henrygd/beszel/pull/2107?

mews-se:
It is — I missed #2107 when I searched, sorry for the noise. Some timing though: that PR had been open since July, and you force-pushed the rework the day before I wrote mine and landed on the same `*_PROVIDED` flag design 🙂

Yours covers the same four service formats plus Windows, so closing this in favour of it. I left a reproduction over there in case it helps with the "should the installer handle this" question.

#2243 is a different one — the line printed at the end of the install, not the service config — so I'll leave that open.

svenvg93:
No worries, I understand 😄 Just want to make user we dont do double work :) 
I pushed some changes yesterday after the review of @henrygd.

Issue #982, opened 2025-07-25
[Bug]: Load average doesn't display if all zeros


davidemyers:
### Component

Hub

### Description

After upgrading the hub and two of my agents to 0.12.0, the load average values in the All Systems column disappear when the load on the monitored system is `load average: 0.00, 0.00, 0.00`. Also the Load Average graph for the system no longer displays.

When the load average increases the data is displayed again.

### Expected Behavior

Show the zeros.

### Steps to Reproduce

Have idle systems. :smile:

### Category

Charts & Visualization

### Affected Metrics

Other

### OS / Architecture

Ubuntu Server 24.04 on both Intel and Raspberry Pi 

### Beszel version

0.12.0

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

henrygd:
Thanks for pointing this out! Will be fixed in the next release 👍

nodesocket:
Also confirm, load average graphs just spins loading for all my hosts on `v0.12.0`.

henrygd:
Should be fixed now in 0.12.1. Let me know if you still see this after upgrading.

nodesocket:
Just upgraded to `0.12.1` in another cluster and when there is not enough data, i.e. default `24h` view still just spinning loading. However, drilling down to `1hr` shows the load average graph. I suspect this goes away when more load average data populates just wanted to mention it.

nodesocket:
Correction, it just showed the spinning loading for a minute or so after upgrade, now the load average graph is working as expected even for `24h` with only  a few data points.

PR #2245, opened 2026-08-19
fix(agent): round load average to two decimals

### LINKED ISSUE #2071: [Bug]: [Windows] Abnormal Load Average values returned
### Welcome!

### Component

Hub

### Problem Description

After installing the agent on Windows, the Load Average display is abnormal. However, after installing the agent on Linux, the display is normal.

### Expected Behavior

The Load Average should return standard float values with reasonable precision (e.g., 2 decimal places)

Image

```json
"la": [
0.08,
0.32,
0.26
]
```

### Steps to Reproduce

Here is the actual output I received:

Image

```json
"la": [
1.3667392689044936e-73,
7.324907119980117e-17,
9.154626702692115e-8
]
```

### OS / Architecture

windows/agent   docker/hub

### Beszel version

0.18.7

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

justadityaraj:
## 📃 Description

Fixes #2071.

`getSystemStats` stores every metric through `utils.TwoDecimals`, except the load averages, which were assigned straight from gopsutil:

```go
systemStats.Cpu = utils.TwoDecimals(cpuMetrics.Total)
systemStats.CpuBreakdown = []float64{utils.TwoDecimals(cpuMetrics.User), ...}
...
systemStats.LoadAvg[0] = avgstat.Load1   // not rounded
```

Linux hides this because `/proc/loadavg` is already two decimal places. Other platforms are not so tidy:

- Windows has no real load average, so gopsutil synthesises one as a decaying EWMA over the `\System\Processor Queue Length` counter. It approaches zero without reaching it, which is where the reporter's `1.3667392689044936e-73` comes from. The value is not wrong, it is just never rounded.
- macOS and BSD divide a fixed point value by `fscale`, giving numbers like `2.55322265625`.

The hub already treats two decimals as the canonical precision here, since `records.go` rounds the load average when averaging records. That left the raw agent records as the only place carrying full precision.

So this is not really Windows specific, it is every platform whose load average does not arrive pre-rounded. Windows is just where it looks alarming.

### 🔧 Fixed

- Load average is now rounded to two decimals, consistent with every other metric the agent reports.

## Steps to verify the change

I do not have a Windows box, so I verified on macOS, which has the same unrounded behaviour for the same reason.

Built a small probe from this module calling the same `load.Avg()` the agent calls, cross compiled for darwin/arm64, and ran it natively:

```
before_la: [2.55322265625, 2.7646484375, 2.90234375]
after_la:  [2.55, 2.76, 2.90]
```

`uptime` on the same machine at the same moment:

```
load averages: 2.55 2.76 2.90
```

Also ran:

- `go test -tags="testing no_ui" ./agent/...` — all green, unchanged from before the patch
- `gofmt -l agent/system.go` — clean
- `go vet ./agent/` — clean
- Cross compiled `./internal/cmd/agent` for windows/amd64, linux/amd64, darwin/arm64 and freebsd/amd64

One thing I checked because it looked like a risk: rounding means a genuinely idle Windows host now reports exactly `[0, 0, 0]`, which is close to #982. It does not come back. The `la` field is `cbor:"28,keyasint"` with no `omitempty`, and `json:"la,omitempty"` is a no-op on a `[3]float64`, so the array still serialises.

No test added. `getSystemStats` calls `load.Avg()` directly with no seam to inject a value, and `utils.TwoDecimals` already has its own tests, so a test here would either need a refactor larger than the fix or would only re-test the helper.

## Type

- [x] Fix

## Checklist

- [x] Title follows the conventional commit format
- [x] Tested locally
- [x] I have no unrelated changes in the PR

### DIFF
--- a/agent/system.go
+++ b/agent/system.go
@@ -171,9 +171,9 @@ func (a *Agent) getSystemStats(cacheTimeMs uint16) system.Stats {
 
 	// load average
 	if avgstat, err := load.Avg(); err == nil {
-		systemStats.LoadAvg[0] = avgstat.Load1
-		systemStats.LoadAvg[1] = avgstat.Load5
-		systemStats.LoadAvg[2] = avgstat.Load15
+		systemStats.LoadAvg[0] = utils.TwoDecimals(avgstat.Load1)
+		systemStats.LoadAvg[1] = utils.TwoDecimals(avgstat.Load5)
+		systemStats.LoadAvg[2] = utils.TwoDecimals(avgstat.Load15)
 		slog.Debug("Load average", "5m", avgstat.Load5, "15m", avgstat.Load15)
 	} else {
 		slog.Error("Error getting load average", "err", err)


henrygd (approved):
Thank you! Didn't know that was happening.

justadityaraj:
Anytime man, love beszel, would contribute as much as I can <3

PR #2256, opened 2026-08-23
fix(agent): carry Intel GPU averages forward between realtime samples

wassupluke:
## Description

On the live **1m** view, the **GPU Power Draw** chart and its legend for an Intel iGPU blank on roughly two of every three-to-four one-second refresh cycles, while every other chart renders normally. NVIDIA/AMD GPUs are unaffected.

**Root cause:** `calculateGPUAverage` treats all-zero temperature + memory as a suspended card and returns an empty `GPUData` (power omitted, since `p`/`pp` are `omitempty`). Intel's `intel_gpu_top` collector never reports temperature or memory, so that heuristic fires on every collection that lands *between* samples. Because `intel_gpu_top` samples every 3.3s (`intelGpuStatsInterval`) while the hub's realtime worker collects every 1s (`system_realtime.go`), most realtime collections have no new sample (delta count 0) and emit an empty GPU block. The frontend derives the power series and legend from the **latest** sample (`gpu-charts.tsx` → `gpuPowerKey`), so the panel blanks until the next 3.3s sample arrives.

NVIDIA/AMD report temperature even when idle, so the heuristic never fires for them and the last average is already carried forward, which is why those cards render smoothly on 1m.

**Fix:** gate the zero-return on non-engine (discrete) GPUs (`gpu.Engines == nil`) so Intel GPUs carry their last average forward during between-sample gaps, matching the existing NVIDIA/AMD behavior. Stored aggregates (1h+) were already correct because a 60s window always contains fresh samples; only the realtime path was affected. Adds a regression test (`carries Intel GPU average forward between samples`).

## Result

- Intel GPU **Power Draw** chart and legend no longer blank intermittently on the realtime (1m) view.

## Screenrecording

Before (Intel iGPU, 1m view GPU Power Draw flickers while other charts are steady):

> First shows the NVIDIA GPU Power Draw rendering as expected, then the Intel iGPU rendering with the graph drops.

https://github.com/user-attachments/assets/3bea84fb-65e0-4fba-a1e3-85de39a3d326

### DIFF
--- a/agent/gpu.go
+++ b/agent/gpu.go
@@ -361,12 +361,16 @@ func (gm *GPUManager) calculateGPUAverage(id string, gpu *system.GPUData, cacheK
 
 	// If no new data arrived
 	if deltaCount == 0 {
-		// If GPU appears suspended (instantaneous values are 0), return zero values
-		// Otherwise return last known average for temporary collection gaps
-		if gpu.Temperature == 0 && gpu.MemoryUsed == 0 {
+		// Only discrete GPUs report temp/memory, so treat all-zero as suspended (return zeros).
+		// Engine-based (Intel) GPUs don't, so carry the last average forward across sample gaps.
+		if gpu.Engines == nil && gpu.Temperature == 0 && gpu.MemoryUsed == 0 {
 			return system.GPUData{Name: gpu.Name}
 		}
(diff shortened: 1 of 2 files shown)


henrygd (approved):
Makes sense, thank you!

PR #2266, opened 2026-08-24
feat: add network probes (ICMP/TCP/HTTP/DNS) 

svenvg93:
## 📃 Description

Adds a **Network Probes** feature: agents can actively monitor external targets (ICMP ping, TCP port checks, HTTP requests, and DNS resolution) on a configurable interval, with results synced to the hub, aggregated over time, and surfaced per-system in a new Probes table and detail view with response-time and packet-loss charts.

### ➕ Added

- Network probes: agents run ICMP, TCP, HTTP, and DNS checks against configurable targets on a per-probe interval, using native ICMP sockets with a fallback to system `ping` where sockets aren't available
- Hub-side sync (full/incremental) and aggregation of probe results, including 1h rollups of avg/min/max response time and packet loss
- New per-system Network Probes table (add/edit/delete probes) and a probe detail sheet with response-time (avg/min/max) and packet-loss charts
- Realtime updates for probe results, matching the live-update behavior of existing system metrics
- Gap detection in probe charts, so periods with no data show as a break in the line instead of a misleading connected line across the gap

### ✏️ Changed

- Probes moved out of the system metrics tab into their own dedicated tab/page
- TCP probe latency no longer includes DNS resolution time, so it reflects the connection itself
- Chart filter parsing (comma/space term matching) consolidated into a shared utility instead of being reimplemented per chart

## 📷 Screenshots

Scherm­afbeelding 2026-08-24 om 19 49 23
Scherm­afbeelding 2026-08-24 om 19 49 31


henrygd:
Looking good, thanks for working on this!

I'm planning to release 0.19.0 in a day or two because we just added ZFS integration. Then I'll get back on this for 0.20.0. Don't worry about the merge conflicts.

niccfun:
I'm really looking forward to this feature. Thank you to the developers.

henrygd:
This is very close to release ready now. I just want to stress test it for a few more days and go through the diff again to make sure we don't mess anything up on merge.

svenvg93:
> This is very close to release ready now. I just want to stress test it for a few more days and go through the diff again to make sure we don't mess anything up on merge.

Thank you so much! 

Just a side question, I have a use case to add like a Ookla Speedtest to this as well later on. Do you think that will work in this model we have now for ICMP etc.

henrygd:
I think it would be best to keep speed testing separate from this. It's a good idea though.

From a quick look, there's a promising Go package for this here: https://github.com/showwin/speedtest-go

Though depending on how heavy it is, it may be better to make our own lightweight solution.

svenvg93:
> I think it would be best to keep speed testing separate from this. It's a good idea though.
> 
> From a quick look, there's a promising Go package for this here: https://github.com/showwin/speedtest-go
> 
> Though depending on how heavy it is, it may be better to make our own lightweight solution.

You mean keep it out beszel at all or just from this PR?  (which as already my plan) 

I had a quick look at the Go Package, there seems to be some deviation in the [results](https://github.com/showwin/speedtest-go#summary-of-experimental-results)  from what Ookla returns and the Go package. My idea was to use the Ookla CLI client which as a json output and store that.

henrygd:
> You mean keep it out beszel at all or just from this PR? (which as already my plan)

I think it's fine for Beszel, I just mean it should be built out separately from the network probe functionality. Though I suppose it would work similarly in that it would run in the background rather than while fetching other stats from the agent.

Ookla CLI would be fine I think, assuming it has servers that are reachable in China.

PR #2269, opened 2026-08-27
fix: GPU Power Draw chart renders full-width instead of half-width

### LINKED ISSUE #2077: [Bug]: "GPU Power Draw" graph uses 100% width in grid view
### Welcome!

### Component

Hub

### Problem Description

The graph for Gpu power draw is always displayed in full width on a machines single view (on display: default and on tabs). see these screenshots:

default:
Image

tabs:
Image

thats probably not intentional? :)

### Expected Behavior

It should have half width like all other graphs

### Steps to Reproduce

- use `GPU_COLLECTOR=macmon` on the agent (maybe unreleated)
- notice that the graph looks different (full width) compared to all other ones

### OS / Architecture

linux/amd64 (hub), macos/arm64 (agent)

### Beszel version

0.18.7

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

santhiprakash:
## 📃 Description

The GPU Power Draw chart on the system detail page always rendered at full width in the grid, on both the default layout and the GPU tab, instead of sharing half width with sibling charts.

**Root cause:** `ChartCard` applies `odd:last-of-type:col-span-full` so that a lone leftover chart in an odd-count grid stretches to fill the row instead of leaving a dangling gap. The GPU Power Draw chart is usually the last (and on the GPU tab, the *only*) chart in its grid container, so this selector always matched it and forced it full-width — even though it should behave like every other paired chart.

This is the same class of layout quirk already worked around for the GPU Usage chart in `GpuDetailCharts` (`className={cn(grid && "!col-span-1")}`). This PR applies the identical, already-established fix to `GpuPowerChart`.

**Fix:** pass `className={cn(grid && "!col-span-1")}` to `GpuPowerChart`'s `ChartCard`, forcing it back to a single grid column (half width) in grid display mode, overriding the `odd:last-of-type` full-width trick.

Fixes #2077

## 📖 Documentation

### 🔧 Fixed

- GPU Power Draw chart now renders at half width (matching sibling charts) instead of full width, on both the default and tabbed system-detail layouts.

## 📷 Screenshots

I could not generate a live screenshot in this environment (no GPU data / running hub instance to reproduce the rendered UI). Before/after description instead:

- **Before:** the GPU Power Draw card spans the full row width by itself on both the default grid layout and the GPU tab, as shown in the issue's screenshots.
- **After:** the GPU Power Draw card is constrained to a single grid column (`col-span-1`), matching the width of every other paired chart (CPU, Memory, etc.), the same way the GPU Usage chart already behaves.

Verified with `bun run build` (production Vite build succeeds) and `bunx biome check` on the touched file (no new lint issues introduced by this change).


henrygd (changes requested):
Unfortunately this causes another problem in the tabbed grid layout with three charts (see screenshot).

My fault for having a funky grid to begin with, but it probably needs to be refactored from a structural standpoint.

Image

santhiprakash:
Your screenshot matches the code path exactly: the power chart sat in its own one-chart grid above the detail grid, so once it was pinned to `col-span-1` nothing could fill the slot beside it. Agreed the two-grid split is the real problem — I've refactored the GPU tab structurally (ef23d09):

- Power draw, engines, and per-GPU usage/VRAM now render as direct children of a single `xl:grid-cols-2` grid, so the cards' existing `odd:last-of-type:col-span-full` parity rule flows across the whole tab — an odd chart count ends with the last chart spanning full width instead of stranding a half-empty row.
- The per-GPU `display: contents` wrappers are replaced with fragments. Those wrappers were what disabled the parity rule inside the detail grid (`nth-child`/`last-of-type` resolve against the wrapper, not the grid), and the usage card's `!col-span-1` was a no-op under them — both dropped.
- The power chart keeps the half-width force only so a *lone* power chart stays half-width (#2077).

Layouts in grid mode now: power only → half width; power + usage + VRAM → `[power | usage] / [VRAM — full]`; with engines → clean 2×2; multi-GPU → the last chart spans full only when the total count is odd, no half-empty rows in any combination. This also fixes a pre-existing hole on `main` for the engines + one-GPU case (engines + usage filled row 1, VRAM sat at half width with an empty slot).

Verification: `bunx biome check` on the two touched files — no new findings (one pre-existing `useTemplate` note in untouched code); `bun run build` (lingui extract + compile + vite production) green; branch merged with current `main` (ZFS/alerts). No visual check on my side — the three-chart GPU tab is the one to eyeball.

henrygd (approved):
Looks good, thanks!

PR #2272, opened 2026-08-27
fix(agent): don't read host CPU and memory totals from a Docker VM

### LINKED ISSUE #2098: [Bug]: macOS agent reports Docker VM memory as host total memory
### Confirmations

- [x] I have read the Documentation.
- [x] I have checked the Common Issues Guide and my problem was not mentioned there.
- [x] I have verified I am using the latest version available (0.18.7).

**Component:** Agent
**OS / Architecture:** darwin/arm64 (agent)
**Beszel version:** 0.18.7
**Installation method:** Binary (Homebrew)

---

## Problem Description

On macOS the agent runs natively (Homebrew binary) while Docker runs inside a Linux VM (OrbStack or Docker Desktop). In this setup the **system-details header reports the Docker VM's RAM as the host's total memory**, instead of the Mac's physical RAM.

On my machine:

- Physical RAM: **64 GiB** (`sysctl hw.memsize` = `68719476736`)
- OrbStack VM memory: **16 GiB** (`~/.orbstack/config`: `memory_mib: 16384`); `docker info` reports `MemTotal` = `16818024448` (≈ 15.66 GiB)
- System-details header **"Total"** shows **~16 GiB**, i.e. the Docker VM value
- The **memory chart, percentage and alerts are all correct (64 GiB)** — so the discrepancy is only in the static "Total" of the system details

### Root cause (from source, v0.18.7)

`refreshSystemDetails()` sets the total memory from the Docker daemon's `/info` response:

```go
// agent/system.go
a.systemDetails.MemoryTotal = hostInfo.MemTotal
if a.systemDetails.MemoryTotal == 0 {
    if v, err := mem.VirtualMemory(); err == nil {
        a.systemDetails.MemoryTotal = v.Total
    }
}
```

`hostInfo` comes from `dockerManager.GetHostInfo()`, which does `GET /info` against the Docker socket (`agent/docker.go`). On macOS, `docker info`'s `MemTotal` is the **Linux VM's** memory, not the host's. The correct value via `mem.VirtualMemory().Total` (which reads `hw.memsize` and the real page size) is only used as a fallback when `hostInfo.MemTotal == 0`, i.e. when no Docker daemon is reachable.

Note the asymmetry: `OsName`/`Kernel` are taken from Docker `hostInfo` **only on Linux**, but `MemoryTotal` (and cores/threads) are taken from Docker `hostInfo` on **all** OSes. On macOS/Windows, where Docker runs in a VM, the Docker host info does not describe the physical machine — so using it for the host's memory looks unintended.

The same likely affects Windows + Docker Desktop, and any setup where a natively-running agent talks to a Docker daemon whose `MemTotal` differs from the host (e.g. a memory-limited VM).

## Expected Behavior

On macOS (and other hosts where the agent runs natively), the system-details "Total" memory should reflect the physical host RAM (e.g. `mem.VirtualMemory().Total` / `hw.memsize`), consistent with the memory chart — not the Docker VM allocation.

## Steps to Reproduce

1. Apple Silicon Mac, e.g. 64 GiB physical RAM.
2. Run OrbStack (or Docker Desktop) with a VM memory limit different from the host RAM (here OrbStack `memory_mib: 16384` = 16 GiB), exposing the default socket `/var/run/docker.sock`. Have at least one container running so the Docker manager is active.
3. Install/run `beszel-agent` natively (Homebrew), with **no** `DOCKER_HOST` override.
4. Add the system to the hub. The system-details header "Total" shows the Docker VM value (~16 GiB), while the memory chart shows the correct physical 64 GiB.

(Consistent with the code, stopping the Docker daemon so the agent can no longer reach it makes the fallback path use `mem.VirtualMemory().Total`, which then reports the correct 64 GiB.)

## Possible fix

Prefer the OS-reported physical memory over the Docker `/info` value when the agent is not running inside the Docker host — e.g. only use `hostInfo.MemTotal` when `Os == Linux` (mirroring how `OsName`/`Kernel` are already handled), or prefer `mem.VirtualMemory().Total` when it is available and non-zero.

## Workaround (local)

If you run the agent natively on macOS and don't need per-container stats for that machine, you can stop the agent from querying Docker at all by setting an **empty `DOCKER_HOST=`** in its environment. With an empty (but set) `DOCKER_HOST`, `newDockerManager()` returns nil, so the agent never creates a Docker manager and `MemoryTotal` falls back to `mem.VirtualMemory().Total` — the correct physical RAM.

End-to-end (this part has a catch worth documenting):

1. Add `DOCKER_HOST=` (empty value) to the agent's environment, then restart the agent. Homebrew: append `DOCKER_HOST=` to `~/.config/beszel/beszel-agent.env` and `brew services restart`. (The Homebrew launcher does `set -a; source …env`, so the empty assignment is exported as a set-but-empty variable, which is what the nil check needs.)
2. **Refresh the hub's cached details.** An agent restart alone is not enough: the hub fetches a system's static details only once and caches them — `detailsFetched` in `internal/hub/systems/system.go` is set after the first successful fetch and is never reset, so the header keeps showing the originally-fetched value across agent reconnects. Restart the hub (or remove and re-add the system) so it re-fetches details on the next poll. Only then does the header update. (This caching is also why a future code fix wouldn't be visible until the hub re-fetches.)

**Trade-off:** with no Docker connection the agent no longer reports that machine's Docker container stats (per-container CPU/memory). Fine if you only want host metrics; not acceptable if you also want container monitoring on that host.

Verified end-to-end on a 64 GiB Apple Silicon Mac: with `DOCKER_HOST=` empty the agent stops querying Docker, and after a hub restart the system-details header changed from ~16 GiB to the correct 64 GiB — matching the live memory chart.

## Severity

Cosmetic — monitoring, percentages and alerts are unaffected. Filing for correctness/consistency, since the header contradicts the chart on the same machine.

justadityaraj:
## 📃 Description

Closes #2098

`refreshSystemDetails()` seeds `MemoryTotal` and `Threads` from the Docker daemon's `/info` response. That response describes the machine the daemon runs on, which is only this host when Docker shares its kernel. On macOS and Windows it runs inside a Linux VM, so the system-details header reports the VM's memory as the host total — 16 GiB on the reporter's 64 GiB Mac.

The CPU numbers are wrong the same way and for the same reason: `threads` comes from `hostInfo.NCPU`, and the lxc branch immediately below then clamps `cores` down to it, so a VM given fewer CPUs than the Mac has shrinks both values.

The fix only consults Docker's host info where the daemon runs natively. On macOS and Windows the existing fallbacks take over, and those read this host: `mem.VirtualMemory()` for memory and `cpu.Counts(true)` for threads.

Memory percentage, the memory chart and alerts were already correct, since they come from `mem.VirtualMemory()` rather than from Docker.

### 🔧 Fixed

- macOS and Windows agents no longer report the Docker VM's memory as the host's total memory
- macOS and Windows agents no longer report the Docker VM's CPU count as the host's cores/threads

## Verification

- `gofmt` clean on the changed file
- `go vet ./agent/...` clean for `GOOS=linux` and `GOOS=darwin`
- `go build ./agent/...` and `go test -c ./agent/` succeed for linux/amd64 and darwin/arm64
- `go test ./internal/...` passes for every package that builds on this Windows host; `internal/hub`, `internal/site` and the two `internal/cmd` packages fail setup here before and after the change, because `agent/sensors_windows.go` embeds a LibreHardwareMonitor artifact this checkout does not build

### DIFF
--- a/agent/system.go
+++ b/agent/system.go
@@ -32,7 +32,11 @@ func (a *Agent) refreshSystemDetails() {
 
 	if a.dockerManager != nil {
 		a.systemDetails.Podman = a.dockerManager.IsPodman()
-		hostInfo, _ = a.dockerManager.GetHostInfo()
+		// Docker's host info describes the machine its daemon runs on. On macOS and
+		// Windows that is a Linux VM, so its CPU and memory totals are not this host's.
+		if runtime.GOOS != "darwin" && runtime.GOOS != "windows" {
+			hostInfo, _ = a.dockerManager.GetHostInfo()
+		}
 	}
 
 	a.systemDetails.Hostname, _ = os.Hostname()


henrygd (approved):
Makes sense. Gracias!

PR #2277, opened 2026-08-29
fix(hub): don't read the SSH client after it is closed

### LINKED ISSUE #2157: [Bug]: Hub can panic in createSessionWithTimeout during concurrent SSH client close

justadityaraj:
## 📃 Description

Closes #2157.

`createSessionWithTimeout` checks `sys.client` for nil, then dereferences it a second time inside the goroutine that calls `NewSession`. Those two reads are not the same read. `update()` runs the SMART fetch in its own goroutine (`system.go:166`), so the updater goroutine can call `closeSSHConnection` and clear the field in between, and the session goroutine dereferences a nil client. That panic is not recovered anywhere, so it takes down the whole hub process.

The client field is now an `atomic.Pointer[ssh.Client]`, loaded once before the goroutine starts and cleared with `Swap`, matching the `atomic.Bool` fields already on the struct. `NewSession` on an already-closed client returns an error, which `runSSHOperation` already retries. `createSSHClient` still stores the dial result unconditionally, so a failed dial clears the field exactly as before.

### 🔧 Fixed

- Hub no longer panics with a nil pointer dereference when an SSH connection is closed while a session is being created.

## Verification

`TestCreateSessionDuringClose` runs session creation against `closeSSHConnection` and reproduces the reported crash on `main` with the same frames as the issue:

```
panic: runtime error: invalid memory address or nil pointer dereference
golang.org/x/crypto/ssh.(*Client).NewSession(0x0?)
	golang.org/x/crypto@v0.55.0/ssh/client.go:161 +0xe
github.com/henrygd/beszel/internal/hub/systems.(*System).createSessionWithTimeout.func1()
	internal/hub/systems/system.go:776 +0x45
created by ...createSessionWithTimeout in goroutine 9
```

It panicked on 3 of 3 runs before the change and passes after it.

- `go test -tags testing ./internal/hub/...`: `config`, `expirymap`, `heartbeat`, `systems`, `transport` and `ws` all pass. `internal/hub` itself does not build on this Windows host, before or after the change, because `agent/sensors_windows.go` embeds a LibreHardwareMonitor binary that is not checked in.
- `go vet -tags testing ./internal/hub/systems/...` and `gofmt -l` clean.
- I could not run `-race` here (no cgo toolchain on this machine), which is part of why the field moved to an atomic rather than a plain captured pointer.

### DIFF
--- a/internal/hub/systems/ssh_timeout_test.go
+++ b/internal/hub/systems/ssh_timeout_test.go
@@ -4,11 +4,13 @@ package systems
 
 import (
 	"errors"
+	"sync"
 	"testing"
 	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/crypto/ssh"
 )
 
 // TestRunWithTimeout covers the guard added for issue #2041: the per-system SSH
@@ -53,4 +55,39 @@ func TestRunWithTimeout(t *testing.T) {
 			synctest.Wait() // ensure the released op goroutine exits cleanly
 		})
 	})
+}
+
+// closedConn stands in for a connection whose peer has gone away: opening a
+// channel fails rather than succeeding, which is what NewSession does on a
+// client that closeSSHConnection has already closed.
+type closedConn struct{ ssh.Conn }
+
+func (closedConn) OpenChannel(string, []byte) (ssh.Channel, <-chan *ssh.Request, error) {
+	return nil, nil, errors.New("use of closed network connection")
+}
+
+func (closedConn) Close() error { return nil }
+
+// TestCreateSessionDuringClose covers issue #2157: the background SMART fetch
+// creates a session while the updater can be tearing the same connection down,
+// so session creation must not read the client field after it is cleared.
+func TestCreateSessionDuringClose(t *testing.T) {
+	for range 500 {
+		sys := &System{ctx: t.Context()}
+		sys.client.Store(&ssh.Client{Conn: closedConn{}})
+
+		var wg sync.WaitGroup
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			session, err := sys.createSessionWithTimeout(time.Second)
+			assert.Nil(t, session)
+			assert.Error(t, err, "a closed connection must surface an error, not a session")
+		}()
+		go func() {
+			defer wg.Done()
+			sys.closeSSHConnection()
+		}()
+		wg.Wait()
+	}
 }
--- a/internal/hub/systems/system.go
+++ b/internal/hub/systems/system.go
@@ -33,22 +33,22 @@ import (
 )
 
 type System struct {
-	Id             string                  `db:"id"`
-	Host           string                  `db:"host"`
-	Port           string                  `db:"port"`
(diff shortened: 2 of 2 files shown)


henrygd (approved):
Looks like a good solution, thanks for your work 👍

PR #2279, opened 2026-08-30
ui: hide gpu indicator for host without gpu

snh:
## 📃 Description

Hides the GPU indicator on the system overview page for hosts that lack reported GPU metrics.

### ✏️ Changed

- GPU indicators for hosts that lack reported GPU metrics are now hidden on the system overview page.


henrygd (approved):
Thanks, I agree it's better than showing 0%.

Issue #176, opened 2024-09-20
Agents on LXC Containers show RAM Usage of Proxmox Host


heeelga:
I have deployed the Beszel agent on several LXC containers on Proxmox which have different amounts of RAM. 
Unfortunately every agent is showing 32GB of RAM in the UI which is the total amount of RAM of the Proxmox host. RAM usage is also showing the exact same value for every agent.

I tried changing the corresponding LXCs to unprivileged/privileged, turned on and off Nesting without any effect.

CPU and disk usage seem to work fine.

Am I missing an option here?

heeelga:
PS: I just figured out that this only occurs when using the dockerized agent.

henrygd:
Thanks for the info, I'll look into it.

henrygd:
To clarify, the binary agent correctly shows the memory allocated to the LXC container?

But installing Docker in the container and running the agent with Docker shows the memory info for the host?

Could you please tell me what `free -h`  and `cat /sys/fs/cgroup/memory.max` output in the container?

heeelga:
Yes, exactly. The docker agent shows the memory for the host and the binary agent shows the correct memory allocated to the specific LXC container it is running on.

Sure thing. I checked on a exemplary LXC and the amount of memory (1G) is correct.

free -h:
                    total        used        free      shared  buff/cache   available
Mem:           1.0Gi       178Mi       659Mi        48Ki       186Mi       845Mi
Swap:          1.0Gi       122Mi       901Mi

cat /sys/fs/cgroup/memory.max:
max

moesix:
Also on Proxmox with a similar setup and ran into this same issue. Binary agent works though.

jlssmt:
I recognized now it's not only for memory, it's the same for CPU.
The Host + all LXC, running beszel inside docker, showing the same cpu and memory utilization.

At least for me, it's a show stopper.

nunobasto:
I also have proxmox with several lxcs with dfferent services and I added the beszel agent on each lxc console. When watching plex, the cpu monitoring for several services seems to be the same as the plex service. In plex lxc I have 3 cores allocated from 4, so the full proxmox server CPU is about 70%.

<img width="1025" alt="Image" src="https://github.com/user-attachments/assets/387db824-cfe5-449a-b435-44115f2097a1" />

The uptime kuma is always so low, but in Beszel it's more than 90%. The memory seems correct, though.

[...]

Issue #144, opened 2024-09-01
 Error getting container stats: container_name - invalid data


Malek-GHub:
Hello

I've installed the Beszel Hub and Agent on a raspberry pi 4, with the logging Container Dozzle
The Hub can see the raspberry Pi cpu, disk and memory etc, 

![image](https://github.com/user-attachments/assets/37af5e6e-f6dc-4846-b070-6675a866f189)

however,  I cannot see any container information and am getting the following error in the logs of the Beszel agent

```
beszel-agent logs

 2024/09/01 15:40:59 Found network interface: eth0 (5065362491 recv, 1064108764 sent)
2024/09/01 15:40:59 Starting SSH server on :45876
2024/09/01 15:41:03 Error getting container stats: beszel - invalid data
2024/09/01 15:41:03 Error getting container stats: beszel-agent - invalid data
2024/09/01 15:41:03 Error getting container stats: dozzle - invalid data
2024/09/01 15:41:56 Error getting container stats: beszel-agent - invalid data
2024/09/01 15:41:56 Error getting container stats: beszel - invalid data
2024/09/01 15:41:56 Error getting container stats: dozzle - invalid data
2024/09/01 15:42:56 Error getting container stats: beszel - invalid data
2024/09/01 15:42:56 Error getting container stats: beszel-agent - invalid data
```

This is my docker compose , and env file.   I know this is insecure but I just wanted it to work like this first before installing the socketproxy... 
Any Ideas would be greatly appreciated..

```
docker-compose.yaml

services:  
   beszel:
     image: henrygd/beszel
     container_name: beszel
     restart: unless-stopped  
     env_file:
       - malmeser.env     
     ports:
       - '8090:8090'
     volumes:
       - ./volumes/beszel/config:/beszel_data

   beszel-agent:
     image: "henrygd/beszel-agent"
     container_name: "beszel-agent"
     restart: unless-stopped
     network_mode: host
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro
     environment:
       PORT: 45876
       KEY: "ssh_key_here"
       FILESYSTEM: /dev/mmcblk02 # set to the correct filesystem for disk I/O stats

   dozzle: # Dozzle - Real-time Docker Log Viewer
     image: amir20/dozzle:latest
     container_name: dozzle
     restart: unless-stopped
     env_file:
       - malmeser.env
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro # Mounted as read-only       
[...]

PR #2285, opened 2026-09-02
fix(agent): correct memory/CPU/network stats for a dockerized agent inside an LXC

mleczakm:
## 📃 Description

Fixes the memory / CPU / network stats reported by the **agent running as a Docker container inside an LXC / Proxmox guest**. Two independent bugs, same underlying cause: a Docker container gets a fresh `procfs` and (on most LXC hosts) no delegated cgroup **memory** controller, so the kernel interfaces the agent reads describe the *physical host*, not the guest.

Addresses **#176** (the memory half — the dockerized-agent case that comment #2 there singles out). Also removes the spurious `bad memory stats` error (the one whose message links to #144) for the "no cgroup memory accounting" case, and stops it from blanking the rest of the Docker section.

Verified live on a 256 MiB LXC (Alpine + Docker, cgroup2, memory controller **not** delegated to nested containers):

| chart | before | after |
| --- | --- | --- |
| Memory usage | `126 GB` total, ~40 GB used (the Proxmox host) | `245.8 MB` total · `81.9 MB` used · `163.8 MB` cache |
| Docker CPU | blank | populated |
| Docker Network I/O | blank | populated |
| Disk usage / IO | already correct | unchanged |

### Root cause

`refreshSystemDetails()` already gets a correct memory total from the container runtime — `Docker/Podman /info` `MemTotal`, which the daemon reads from the *namespaced* `/proc/meminfo` the agent can't see — and uses it for the details header. But `getSystemStats()` takes `Mem` / `MemUsed` / `MemPct` straight from `gopsutil mem.VirtualMemory()` → `/proc/meminfo` → the host's total.

Separately, on that same host the Docker stats API returns an all-zero `memory_stats` (no `usage`) for every container, because the cgroup **memory** controller isn't in the LXC's `cgroup.subtree_control`. `calculateMemoryUsage` returned an error for that, and the error aborted the *entire* per-container sample in `updateContainerStats` — so Docker CPU % and Docker Network I/O disappeared too, not just Docker memory.

### How it's fixed

**System memory** — new `agent/cgroup.go` + `applyContainerMemoryLimit` in `agent/system.go`:

1. Gate on the container-runtime total: act only when `systemDetails.MemoryTotal` is present and <99 % of gopsutil's total. That mismatch is the signal that `/proc/meminfo` describes a bigger machine than the agent's real environment. A memory-limited agent on a normal host sees the same total from both sources and is left untouched.
2. If the cgroup **memory** controller is delegated (systemd Proxmox LXC, Kubernetes, `docker run --memory`), use `memory.max` / `memory.current` / `memory.stat` (v2) or the v1 equivalents for an exact limit and real `used` (`current − file` cache) + `buff/cache`.
3. Otherwise (Docker in an unprivileged LXC with no delegation) use the runtime total and scale the host `used` / `buff-cache` proportionally, so the chart is bounded and plausible instead of showing host GB. Exact `used` there needs cgroup memory delegation or the non-Docker agent — noted in a code comment.

**Docker container stats** — `calculateMemoryUsage` in `agent/docker.go`: an absent `memory_stats.usage` now returns `0, nil` ("accounting unavailable") instead of an error, so the container's CPU and network stats are still recorded. A real underflow (cache larger than usage) or an implausible value still errors as before.

### Not covered here

- **Per-container Docker memory** stays empty where the cgroup memory controller isn't delegated to containers — the kernel does no per-container accounting there, `docker stats` itself shows `0B / 0B`, so there is no data to read. Needs an infra change (`Delegate=yes` / `+memory` in the LXC's `subtree_control`) or the binary agent.
- System **CPU %** / load average inside such an LXC are still host-wide (`/proc/stat`, same fresh-procfs cause) — a separate part of #176, left for a follow-up.

### ➕ Added

- `agent/cgroup.go`: minimal cgroup v2 / v1 memory limit + usage reader.

### 🔧 Fixed

- Dockerized agent inside an LXC / Proxmox container reported the physical host's total and used memory instead of the container's (#176).
- Container memory limits from `docker run --memory` (or Kubernetes) are now reflected in the system memory chart.
- Docker CPU % and Docker Network I/O charts went blank on hosts where the Docker API reports no container memory (`bad memory stats`); memory is now treated as unavailable without dropping the rest of the sample.

## Verification

- `gofmt` clean; `go vet ./agent/...` clean; `go test ./agent/...` green.
- `GOOS=linux` and `GOOS=darwin` `go build ./agent/...` succeed.
- New tests: `TestReadCgroupMemory`, `TestApplyContainerMemoryLimit`; extended `TestCalculateMemoryUsage` / `TestMemoryStatsEdgeCases`.
- Live-tested by hot-swapping the built static binary into a running `henrygd/beszel-agent` container (`docker cp … :/agent && docker restart`) on the LXC above.

## 📷 Screenshots

Memory-usage tooltip after the fix: **Total 245.8 MB · Used 81.9 MB · Cache/Buffers 163.8 MB** (a 256 MiB LXC), instead of a flat ~40 GB line on a 126 GB axis.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agent/cgroup.go
+++ b/agent/cgroup.go
@@ -1,0 +1,120 @@
+package agent
+
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// cgroupMemoryStats holds the memory limit and usage of the cgroup the agent
+// runs in. It is the reliable source of container memory figures when the agent
+// runs inside an LXC / Proxmox container (where /proc/meminfo still describes the
+// physical host) or under a `docker run --memory` / Kubernetes limit.
+type cgroupMemoryStats struct {
+	limit   uint64 // hard memory limit in bytes
+	used    uint64 // working set: current usage minus reclaimable page cache
+	cache   uint64 // file-backed page cache in bytes
+	limitOK bool   // a finite limit is configured
+	usageOK bool   // per-cgroup usage accounting is exposed
+}
+
+// cgroup v1 reports "unlimited" as a page-aligned value close to the max int64.
+// Anything at or above this is treated as no limit.
+const cgroupV1Unlimited = uint64(1) << 62
+
+// cgroup file locations, kept as vars so tests can point them at fixtures.
+var (
+	cgroupV2Root = "/sys/fs/cgroup"
+	cgroupV1Root = "/sys/fs/cgroup/memory"
+)
+
+// readCgroupMemory reads the current cgroup's memory limit and usage, preferring
+// the cgroup v2 unified hierarchy and falling back to cgroup v1. Every field is
+// best-effort: callers must check limitOK / usageOK.
+func readCgroupMemory() (s cgroupMemoryStats) {
+	if readCgroupV2Memory(&s) {
+		return s
+	}
+	readCgroupV1Memory(&s)
+	return s
+}
+
+// readCgroupV2Memory populates s from the unified hierarchy. It returns true when
+// the v2 memory controller files are present (even if no limit is set), so the
+// caller knows not to try v1.
+func readCgroupV2Memory(s *cgroupMemoryStats) bool {
+	raw, err := os.ReadFile(cgroupV2Root + "/memory.max")
+	if err != nil {
+		return false
+	}
+	if limit, ok := parseUint(strings.TrimSpace(string(raw))); ok {
+		s.limit, s.limitOK = limit, true
+	}
+	if current, ok := readUintFile(cgroupV2Root + "/memory.current"); ok {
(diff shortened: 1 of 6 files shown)


boxwork-ctrl:
I appear to be hitting this as well. I have Beszel agents running in Docker inside separate LXC guests, and the systems are reporting essentially identical host-level RAM values rather than the resources assigned to each guest. This PR looks like it addresses exactly what I'm seeing. Happy to test the change against my setup if that would be helpful.

PR #2297, opened 2026-09-04
fix(agent): avoid reconnect loops during slow collection

### LINKED ISSUE #2294: [Bug]: 70s WebSocket deadline causes reconnect loops when collection exceeds 70s
### Welcome!

### Component

Agent

### Problem Description

After upgrading both the Hub and agents to `0.19.0`, some WebSocket agents periodically disconnect and immediately reconnect.

This is much easier to reproduce with the `beszel-agent:alpine` image. In my environment, the standard `0.19.0` agent does not show this behavior, while two Alpine-based agents running on PVE hosts repeatedly hit it.

Example logs:

```text
WARN Connection closed err="connection closed, code=1000, reason="
WARN Disconnected from hub
INFO WebSocket connected host=beszel.example.com
```

After checking the source code and collection timestamps in the database, this appears to be caused by a tight timing boundary:

* Hub collection interval: **60 seconds**
* WebSocket deadline: **70 seconds**
* Normal agents: maximum interval around **60.2 seconds**
* Affected PVE / Alpine agents: occasionally around **80 seconds**
* When the 70-second deadline is exceeded, the connection expires and the agent reconnects
* During reconnection, the previous connection is closed with WebSocket code `1000`

The 70-second deadline therefore leaves only about **10 seconds of tolerance** above the normal collection interval.

I also checked the current `main` branch and the deadline still appears to be 70 seconds.

The Hub is behind nginx with `proxy_read_timeout 360s`. The disconnects occur well before this timeout, and nginx access logs show new WebSocket connections corresponding exactly to the reconnect events.

The Alpine image may simply make this easier to trigger due to additional or slower metric collection. I have not confirmed that Alpine itself is the root cause; the underlying issue appears to be the narrow margin between the 60-second collection interval and the 70-second deadline.

### Expected Behavior

A healthy WebSocket agent should remain connected when a metric collection cycle occasionally takes slightly longer than 70 seconds.

The WebSocket deadline should provide sufficient tolerance above the normal 60-second collection interval.

For example, increasing the deadline from 70 seconds to 120 seconds, or deriving it from the collection interval, would tolerate occasional slow collection cycles while still detecting genuinely stalled connections.

### Steps to Reproduce

1. Run Beszel Hub `0.19.0`.
2. Run a `0.19.0` agent using the `beszel-agent:alpine` image on a host where metric collection can occasionally take longer than 70 seconds (PVE hosts reproduce this in my environment).
3. Connect the agent to the Hub using WebSocket/WSS.
4. Leave the agent running and observe its logs.
5. When a collection cycle exceeds approximately 70 seconds, observe:

```text
WARN Connection closed err="connection closed, code=1000, reason="
WARN Disconnected from hub
INFO WebSocket connected host=beszel.example.com
```

6. Compare the collection timestamps in the database. In my environment, unaffected agents stay around ~60 seconds, while the affected agents occasionally reach ~80 seconds.
7. The reconnect timestamps also correlate with new WebSocket requests in the nginx access log.

For reference, nginx is configured with:

```nginx
proxy_http_version 1.1;
proxy_set_header Upgrade $http_upgrade;
proxy_set_header Connection "upgrade";
proxy_read_timeout 360s;
```

The disconnect therefore occurs well before the reverse proxy timeout.

### OS / Architecture

Proxmox VE / Linux x86_64

### Beszel version

0.19.0

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

ryanchou1994:
## 📃 Description

Closes #2294

The agent currently resets its WebSocket deadline to 70 seconds, while the hub default collection interval is 60 seconds. A slow collection can exceed that window and make the agent reconnect even though the hub is still servicing it.

This change extends the deadline to 120 seconds and adds a regression test that keeps the slow-collection window from being reduced below two minutes.

## 📖 Documentation

No documentation changes are needed.

### 🔧 Fixed

- Prevent reconnect loops when a collection cycle takes longer than 70 seconds.

## 📷 Screenshots

N/A

## Verification

- `go test -tags=testing ./agent -run "Test(WebSocketClient|ConnectionManager|WebSocketDeadline)" -count=1`
- `go test -tags="testing no_ui" ./agent -skip TestDirectoryIsWritable -count=1`
- `go vet -tags=testing ./agent`
- `go build -o /tmp/beszel-agent ./internal/cmd/agent`
- `gofmt` and `git diff --check`

The full agent suite was also compared before and after this change; both runs retain the same container-only `TestDirectoryIsWritable/non-writable_directory` failure because the test container runs as root. No additional failure was introduced.

### DIFF
--- a/agent/client.go
+++ b/agent/client.go
@@ -25,7 +25,9 @@ import (
 )
 
 const (
-	wsDeadline = 70 * time.Second
+	// Keep the connection alive long enough for a slow collection cycle to
+	// finish before the hub considers the agent disconnected.
+	wsDeadline = 120 * time.Second
 )
 
 type caCertFileError struct {
--- a/agent/client_test.go
+++ b/agent/client_test.go
@@ -700,3 +700,11 @@ func TestGetToken(t *testing.T) {
 		assert.Equal(t, expectedToken, token, "Whitespace should be stripped from token file content")
 	})
 }
+
+func TestWebSocketDeadlineCoversSlowCollection(t *testing.T) {
+	const minimumDeadline = 120 * time.Second
+
+	if wsDeadline < minimumDeadline {
+		t.Fatalf("WebSocket deadline %s is shorter than the slow-collection window of %s", wsDeadline, minimumDeadline)
+	}
+}


henrygd (approved):
Thanks 👍

PR #2299, opened 2026-09-04
feat(agent): add Docker volume disk usage

### LINKED ISSUE #860: [Feature]: Docker Volume Disk Usage

Similarly how there is a separate docker cpu and memory usage chart, would love if there was one for docker volumes.

An additional chart called "Docker Volume Disk Usage" and would show the size of every docker volume directory

xDefcon:
## 📃 Description

As I needed a way of monitoring and showing the size of docker volumes, as per requested in #860 I'm opening this PR.

Adds a "Docker Volume Usage" chart showing the size of each Docker volume next to the existing container CPU, memory and network charts (Closes #860).

Sizes come from the Docker API's `/system/df` endpoint. That call can be slow especially on large volumes, so the collection is off by default and enabled with a new agent env var `DOCKER_VOLUME_INTERVAL` (a duration, minimum 1m). Agents that don't set it send nothing and do no extra work.

Collection runs in a background worker so the slow call never blocks gatherStats, and a failed request keeps the previous snapshot. 

Data is stored on `system.Stats` as `dv`, a map of volume name to size in GB, and averaged per volume.

A nice to have follow-up could be adding Alerts on volume size 😃 

## 📖 Documentation

[Docs PR here](https://github.com/henrygd/beszel-docs/pull/78)

### ➕ Added

- Docker volume usage chart, data collection is enabled with the `DOCKER_VOLUME_INTERVAL` agent environment variable

## 📷 Screenshots
image

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -190,6 +190,7 @@ func (a *Agent) gatherStats(options common.DataRequestOptions) *system.CombinedD
 		} else {
 			slog.Debug("Containers", "err", err)
 		}
+		data.Stats.DockerVolumes = a.dockerManager.volumeSizes()
 	}
 
 	// skip updating systemd services if cache time is not the default 60sec interval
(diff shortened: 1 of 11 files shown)


xDefcon:
Worth flagging: if /system/df keeps failing for whatever reason, republishing the last snapshot forever would draw a flat line that looks the same as volumes not changing. Should we keep the last successful reading for at most one failed refresh, or publish nothing straight away and show a gap?

albertorizzi:
it could be useful to see which conainer takes so space.

PR #2303, opened 2026-09-05
feat(site): display NVMe data units as human-readable GB/TB

matth3wdsouza:
## 📃 Description

Displays NVMe Data Units Written and Data Units Read S.M.A.R.T. values as human-readable GB/TB instead of raw unit counts, reusing the existing `formatCapacity` function. Only these two attributes are affected.

## 📖 Documentation

No documentation changes required.

### ✏️ Changed

- NVMe Data Units Written/Read now show as GB/TB in the S.M.A.R.T. details sheet.

### DIFF
--- a/internal/site/src/components/routes/system/smart-table.tsx
+++ b/internal/site/src/components/routes/system/smart-table.tsx
@@ -75,7 +75,12 @@ export const smartColumns: ColumnDef<SmartAttribute>[] = [
 		header: "Name",
 	},
 	{
-		accessorFn: (row) => row.rs || row.rv?.toString(),
+		accessorFn: (row) => {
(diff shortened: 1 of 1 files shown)


henrygd (changes requested):
Thanks! However it looks like NVMe SMART data units are defined as 1,000 × 512 bytes, so I think the conversion should use `units * 1000 * 512` rather than `units * 1024 * 512`.

Page 181: https://nvmexpress.org/wp-content/uploads/NVMe-NVM-Express-2.0a-2021.07.26-Ratified.pdf

matth3wdsouza:
You're right. Thanks for the correction!

Updated to use 1000 * 512 per the spec.

henrygd (approved):
Gracias 👍

PR #2307, opened 2026-09-06
fix(agent): fall back to -d sat when smartctl returns identity-only data

### LINKED ISSUE #2295: [Bug]: SATA disks that `smartctl --scan` reports as `-d scsi` silently record UNKNOWN (Synology DSM)
### Welcome!

- [x] I have verified I am using the latest version available.

### Component

Agent

### Problem Description

On Synology DSM, `smartctl --scan` reports every SATA drive as `-d scsi`. The agent trusts
that type, parses with the SCSI parser, finds no ATA data, and records **UNKNOWN / 0 °C /
0 hours** for every disk. Passing `-d sat` to the same binary returns everything.

The failure is silent. The agent logs nothing — no `no valid SMART data found`, no
`smartctl failed` — because from its point of view the scan succeeded and a parser ran. In
the hub the disks simply appear with no health, which is indistinguishable from "this
machine has no SMART support". That is what makes it worth reporting: an empty panel and a
working one look identical.

Same host, same smartctl 7.5 binary, only `-d` differs:

```
# smartctl --scan
/dev/sda -d scsi # /dev/sda, SCSI device
/dev/sdb -d scsi # /dev/sdb, SCSI device
/dev/sdc -d scsi # /dev/sdc, SCSI device
/dev/sdd -d scsi # /dev/sdd, SCSI device

# smartctl -j -a -d scsi /dev/sda
  ata_smart_attributes: absent
  smart_status.passed:  null
  temperature:          0

# smartctl -j -a -d sat /dev/sda
  ata_smart_attributes: present
  smart_status.passed:  true
  temperature:          38
  power_on_hours:       43953
```

Which lands in the hub as:

| device | model | type | state | temp | hours |
|---|---|---|---|---|---|
| /dev/sda | TOSHIBA MG03ACA200 | scsi | UNKNOWN | 0 | 0 |
| /dev/sdb | WDC WD40EFRX-68N32N0 | scsi | UNKNOWN | 0 | 0 |
| /dev/sdc | WDC WD6001FFWX-68Z39N0 | scsi | UNKNOWN | 0 | 0 |
| /dev/sdd | HGST HUH728080ALE600 | scsi | UNKNOWN | 0 | 0 |

`getSmartStatus(temperature, passed)` turns that into `UNKNOWN` — `passed` is false because
the SCSI parse found no status, and `temperature` is 0 — so the disks are reported as
having no data rather than as a collection failure.

For contrast, the same agent version on three Proxmox hosts reads their disks correctly, so
this is specific to how DSM's controller presents the drives to `--scan`.

### Expected Behavior

When the scan-reported device type yields no usable SMART data, retry with `-d sat` before
recording UNKNOWN — the drive is an ATA device behind a SAT layer, which is exactly what
`-d sat` is for. Failing that, log the fallthrough so it is visible that a parse produced
nothing, rather than silently storing an empty record.

### Steps to Reproduce

1. Install the agent on a Synology DSM NAS (DSM 7.1.1, kernel 3.10.108, x86_64) with SATA
   drives. A smartctl new enough for `--json` is required — DSM ships 6.5, so this used a
   statically built 7.5 on PATH.
2. Let the agent collect SMART.
3. Observe every disk in the hub with state UNKNOWN, 0 °C, 0 hours, and nothing in the
   agent log.
4. Run `smartctl -j -a -d sat /dev/sda` on the host and observe full attributes and PASSED.

### Workaround

`SMART_DEVICES=/dev/sda:sat,/dev/sdb:sat,/dev/sdc:sat,/dev/sdd:sat` pins the type
explicitly. Worth noting the workaround requires knowing the device list up front, which is
awkward to automate — we probe each `/dev/sd?` with `-d sat` at install time and build the
list from whatever answers.

### Environment

- Agent 0.19.0, hub 0.19.0
- Synology DS1813+ / DSM 7.1.1-42962 Update 9, kernel 3.10.108, x86_64
- smartctl 7.5 (static build; DSM's own 6.5 has no `--json`)

atirna:
## 📃 Description

On hosts where `smartctl --scan` reports SATA drives as `-d scsi` (Synology DSM, per #2295), the agent runs smartctl without `-d` (scan-detected scsi/ata types are left off so smartctl can auto-detect, #1345). That query returns identity fields but no health sections, and the parsers accepted it as valid data, so every disk was silently recorded as UNKNOWN / 0 °C / 0 hours, with nothing in the agent log.

This change does two things in the agent:

- **Reject identity-only output.** Each parser (SAT, SCSI, NVMe) now treats output that carries a serial number but no health data for its schema (no `ata_smart_attributes` table / devstat / temperature for ATA; no `smart_status.passed` field for SCSI; no health log for NVMe) as a failed collection instead of storing zeros.
- **Retry with `-d sat`.** When the scan-detected type is `scsi` or `ata` and the first query yields no valid data, the agent re-queries with `-d sat`. The drive is an ATA device behind a SAT layer, which is exactly what `-d sat` is for. The retry is skipped for explicit `SMART_DEVICES` `:type` hints, which remain deliberate user overrides (#2102). On success the device is marked verified with type `sat`, so later refreshes query it directly.

When both queries parse successfully but contain no usable health data, the agent logs `no valid SMART data found` and skips the device. Command errors retain the existing `smartctl failed` handling.

The two new fixtures (`scsi_identity.json`, `sat.json`) are representative payloads derived from the fields reported in #2295, not complete smartctl captures. They model identity-only output and SAT health data (PASSED, 38 °C, 43953 power-on hours).

### 🔧 Fixed

- SATA disks reported by `smartctl --scan` as `-d scsi` (e.g. Synology DSM) no longer record silent UNKNOWN / 0 °C / 0 h: the agent detects identity-only smartctl output, retries with `-d sat`, and reports the drive's real health data
- Drives where no parser finds health data are logged and skipped instead of being stored as zeroed records

## Verification

- `go test -tags='testing no_ui' ./agent/ -run 'Smart'`: all SMART tests pass, including the new regression tests for the identity-only and SAT fixture cases (`TestParseSmartOutputRejectsIdentityOnlyData`, `TestCollectSmartSatRetry`: sat recovery, sat-also-empty, explicit-type skip)
- Both new tests fail on the pre-fix code (verified by temporarily reverting `agent/smart.go`)
- Existing fixtures (`scsi.json`, `sda.json`, `nvme0.json`, `apple_nvme.json`) and devstat temperature tests unchanged and passing

Fixes #2295

### DIFF
--- a/agent/smart.go
+++ b/agent/smart.go
@@ -522,6 +522,26 @@ func (sm *SmartManager) CollectSmart(deviceInfo *DeviceInfo) error {
 
 	hasValidData := sm.parseSmartOutput(deviceInfo, output)
 
+	// If the query without an explicit type (scan-detected scsi/ata types are
+	// left off; see issue #1345) returned identity-only data, the drive is
+	// likely ATA behind a SCSI (SAT) layer. Re-query with -d sat before
+	// recording no data. See github.com/henrygd/beszel/issues/2295.
+	if !hasValidData && !deviceInfo.explicitType &&
+		(strings.EqualFold(deviceInfo.Type, "scsi") || strings.EqualFold(deviceInfo.Type, "ata")) {
+		satInfo := *deviceInfo
+		satInfo.Type = "sat"
+		satInfo.parserType = ""
+		satInfo.typeVerified = false
+
+		ctx4, cancel4 := context.WithTimeout(context.Background(), 15*time.Second)
+		defer cancel4()
+		output, err = exec.CommandContext(ctx4, sm.smartctlPath, sm.smartctlArgs(&satInfo, false)...).CombinedOutput()
+		hasValidData = sm.parseSmartOutput(&satInfo, output)
+		if hasValidData {
+			*deviceInfo = satInfo
+		}
+	}
+
 	// If NVMe controller path failed, try namespace path as fallback.
 	// NVMe controllers (/dev/nvme0) don't always support SMART queries. See github.com/henrygd/beszel/issues/1504
 	if !hasValidData && err != nil && isNvmeControllerPath(deviceInfo.Name) {
@@ -891,6 +911,15 @@ func (sm *SmartManager) parseSmartForSata(output []byte, deviceType string) (boo
 		return false, data.Smartctl.ExitStatus
 	}
 
+	// Identity-only output (serial, model, capacity but no ATA health data)
+	// means smartctl could not read SMART through this device type. Accepting
+	// it stores silent zeros; see github.com/henrygd/beszel/issues/2295.
+	if len(data.AtaSmartAttributes.Table) == 0 && data.Temperature.Current == 0 &&
+		!hasJSONValue(data.AtaDeviceStatistics) {
+		slog.Debug("no ATA SMART data", "device", data.Device.Name)
+		return false, data.Smartctl.ExitStatus
+	}
+
 	sm.Lock()
 	defer sm.Unlock()
 
@@ -1006,6 +1035,14 @@ func (sm *SmartManager) parseSmartForScsi(output []byte, deviceType string) (boo
 	// Skip virtual devices (e.g., Kubernetes PVCs, QEMU, VirtualBox, etc.)
 	if sm.isVirtualDeviceScsi(&data) {
 		slog.Debug("skipping smart", "device", data.Device.Name, "model", data.ScsiModelName)
+		return false, data.Smartctl.ExitStatus
+	}
+
+	// Identity-only output means smartctl could not read health data through
+	// this device type. Accepting it stores silent zeros; see
+	// github.com/henrygd/beszel/issues/2295.
+	if data.Temperature.Current == 0 && data.PowerOnTime.Hours == 0 && !data.SmartStatus.Passed {
+		slog.Debug("no SCSI health data", "device", data.Device.Name)
 		return false, data.Smartctl.ExitStatus
 	}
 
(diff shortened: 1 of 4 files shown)


henrygd (changes requested):
Thank you. Looks good overall, and the `-d sat` fallback makes sense.

One thing I’d like to change before merging: the new validity checks use decoded zero values, so `passed: false` is indistinguishable from `passed: null` / missing. That could make us reject real SMART data, including an actual failed SMART status.

Could we check whether the health section is present instead? For example:

```go
var health struct {
	SmartStatus json.RawMessage `json:"smart_status"`
	NVMeHealth  json.RawMessage `json:"nvme_smart_health_information_log"`
}

_ = json.Unmarshal(output, &health)

hasScsiHealth := hasJSONValue(health.SmartStatus)
hasNvmeHealth := hasJSONValue(health.NVMeHealth)
```

Or use pointer fields where we need to distinguish `false` / `0` from missing/null.

It’d also be good to add a regression test for `"smart_status": {"passed": false}`.

atirna:
the zero-value checks can't tell a real `"passed": false` from a missing section. pushed a presence check.

- NVMe only needs to know the section is there, so it reuses `hasJSONValue` on a `json.RawMessage`, same as the ATA path already does for `ata_device_statistics`.
- SCSI needs the value too, not just presence, so that one decodes through a pointer (`*struct{ Passed *bool }`). nil means absent or null, non-nil means reported.

with presence established, SCSI now derives the status directly instead of calling `getSmartStatus`. that helper falls back to `temperature > 0` before returning FAILED, which was a proxy for "did we really read health data", and on a drive reporting `passed:false` with no temperature it returned UNKNOWN. so a genuine failed disk showed as UNKNOWN. left `getSmartStatus` alone for the SATA path since its presence guard checks the ata attributes rather than `smart_status`, so it still cant tell a reported false from a decoded zero there.

regression is table driven, covers `passed:false`, `passed:true`, null and absent for scsi, plus the zero-valued/null/absent nvme cases. the two that matter (`scsi_failed`, `nvme_zero`) fail on the previous head and pass on this one:

```
--- FAIL: TestParseSmartOutputHealthPresence/scsi_failed
--- FAIL: TestParseSmartOutputHealthPresence/nvme_zero
```

full `./agent/...` run has the same 5 failures before and after my change (gpu collector tests, they want nvidia-smi/rocm-smi which this box doesnt have), so nothing new from me.

Issue #2296, opened 2026-09-04
[Bug]: SMART attribute 198 flags WARNING on raw value alone, ignoring normalised value and drivedb absence


ChrisonSimtian:
### Welcome!

- [x] I have read the [Documentation](https://beszel.dev/guide/getting-started)
- [x] I have checked the [Common Issues Guide](https://beszel.dev/guide/common-issues) and my problem was not mentioned there.
- [x] I have searched open and closed issues and my problem was not mentioned before.
- [x] I have verified I am using the latest version available.

### Component

Agent

### Problem Description

**Up front, so nobody wastes time reproducing it my way:** on my hardware this does not
currently surface, because `smartctl --scan` classifies the drive as `scsi` and it never
reaches `parseSmartForSata`. I am reporting the logic with a real drive as a
counter-example, not a broken dashboard. If that is too speculative to act on, please
close it — no argument from me.

The check added in #2275 (shipped in v0.19.0, `agent/smart.go`):

```go
if smartData.SmartStatus == "PASSED" && rawValue > 0 && (attr.ID == 5 || attr.ID == 197 || attr.ID == 198) {
    smartData.SmartStatus = "WARNING"
}
```

The goal from #2268 is right — "PASSED" really does hide developing damage, and reallocated
and pending sectors are good signals. The concern is narrower: **attribute 198 keyed on raw
value alone, with no corroboration**, on drives where smartmontools has no vendor table and
is therefore only *guessing* what 198 means.

A concrete drive, an HPE-branded SATA SSD, firmware `HPG2`:

```
ID# ATTRIBUTE_NAME          FLAG     VALUE WORST THRESH TYPE      UPDATED  WHEN_FAILED RAW_VALUE
  5 Reallocated_Sector_Ct   0x0033   100   100   036    Pre-fail  Always       -       0
187 Reported_Uncorrect      0x0032   100   100   000    Old_age   Always       -       0
196 Reallocated_Event_Count 0x0033   100   100   036    Pre-fail  Always       -       0
197 Current_Pending_Sector  0x0012   100   100   000    Old_age   Always       -       0
198 Offline_Uncorrectable   0x0010   100   100   000    Old_age   Offline      -       224
199 UDMA_CRC_Error_Count    0x003e   100   100   000    Old_age   Always       -       0

SMART overall-health self-assessment test result: PASSED
Device is: Not in smartctl database 7.5/5706
```

Four independent reasons 224 is not a bad-sector count here:

[...]

PR #2275, opened 2026-08-28
Warn when critical ATA SMART attributes are nonzero

justadityaraj:
Fixes #2268

### Problem
ATA SMART health remained `PASSED` when smartctl's overall assessment passed, even if critical attributes 5 (reallocated sectors), 197 (pending sectors), or 198 (offline uncorrectable sectors) had a nonzero raw value.

### Fix
Set the ATA drive health to `WARNING` after parsing one of those attributes with a nonzero raw value. Existing `FAILED` and `UNKNOWN` statuses remain unchanged.

### User impact
Users receive a degraded SMART health state and its existing alert path for these early disk-failure indicators, without affecting NVMe, SCSI, eMMC, or mdraid devices.

## 📖 Documentation

N/A - no user-facing command or configuration changes.

### 🔧 Fixed

- Warn when critical ATA SMART attributes have nonzero raw values.

## 📷 Screenshots

N/A - no UI changes.

## Verification

- `go test -tags testing ./agent -run '^TestParseSmartForSata' -count=1` passed, including the new synthetic smartctl JSON cases for attributes 5, 197, and 198 and the `FAILED`/`UNKNOWN` preservation cases.
- `go test -tags testing ./agent -count=1` reached unrelated Windows-only failures: Unix error-text and path-separator assertions in `TestGetToken`, `TestAddExtraFilesystemFolders`, and `TestAddPartitionExtraFs`.

PR #2308, opened 2026-09-07
fix(agent): avoid false warnings for vendor-specific ATA attributes

crimist:
## Issue

Beszel currently marks any non-zero raw ATA attributes 5, 197, or 198 as `WARNING`, even when the drive vendor uses those IDs for unrelated counters.

This can produce false SMART alerts, such as Crucial MX500 attribute 197 (`Current_Pending_ECC_Cnt`) briefly reporting `1` while SMART remains `PASSED` and no sectors are pending.

This causes me to get constant stream of false warnings via my Discord webhook.

> Disk /dev/sda (Crucial_CT525MX300SSD1) SMART status changed to WARNING

## The fix

To fix this the logic reads the the SMART attribute name when evaluating attributes 197 and 198 to determine whether they are relevant to drive health or unrelated vendor attributes.

The logic is based on this [database metadata](https://github.com/smartmontools/smartmontools/blob/main/drivedb/drivedb.h) so it should generalize to other drives / vendors. We can always tweak the logic if more reports come in. 

Related: #2296

### DIFF
--- a/agent/smart.go
+++ b/agent/smart.go
@@ -931,7 +931,7 @@ func (sm *SmartManager) parseSmartForSata(output []byte, deviceType string) (boo
 		if parsed, ok := smart.ParseSmartRawValueString(attr.Raw.String); ok {
 			rawValue = parsed
 		}
-		if smartData.SmartStatus == "PASSED" && rawValue > 0 && (attr.ID == 5 || attr.ID == 197 || attr.ID == 198) {
+		if smartData.SmartStatus == "PASSED" && rawValue > 0 && shouldWarnForAtaAttribute(data, attr) {
 			smartData.SmartStatus = "WARNING"
 		}
 		smartAttr := &smart.SmartAttribute{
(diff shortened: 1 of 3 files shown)


henrygd (commented):
Thanks for working on this. I think the name checks for 197/198 are a good improvement, but I’m hesitant to merge this as-is.

Issue #2296 shows that these raw SMART values are unreliable enough across vendors that we may still generate false warnings. In particular, attribute 5 is still treated as critical purely by ID/raw value, even though Crucial uses it for retired NAND blocks that may not indicate an actionable failure.

I think we should either revert the warning logic from #2275 for now, or apply a more conservative rule consistently across 5/197/198 rather than relying on raw value alone.

Sorry for lack of guidance, I really don't know what the best heuristic would be considering different vendor implementations. Making up our own decision engine for warnings may end up being more trouble than it's worth.

crimist:
Closing this in favour of a different solution: https://github.com/henrygd/beszel/issues/2296#issuecomment-5717354455

PR #2310, opened 2026-09-07
fix(site): redirect to login when JWT expires

### LINKED ISSUE #2293: [Bug]: user isn't kicked out of the dashboard when their session expires
### Welcome!

### Component

Hub

### Problem Description

If your session expires while using the hub there's no notification or login redirect. `1m` charts still update but not the others. 

### Expected Behavior

User is redirected to the login page when their session expires

### Steps to Reproduce

1. set auth token duration to something short
2. login
3. click on a system
4. wait for token to expire
5. change time filter, 1m continues to work, other charts are blank

### OS / Architecture

linux/amd64 (hub+agent)

### Beszel version

0.18.8

### Installation method

Binary

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

santhiprakash:
## Description

This PR fixes the silent half-broken dashboard state when the PocketBase JWT expires while the hub is open.

### Problem

`$authenticated` in `src/lib/stores.ts` is initialized once from `pb.authStore.isValid` and only updated on explicit `authStore.onChange` events. PocketBase's `onChange` does not fire when a JWT's `exp` claim passes, so the dashboard keeps rendering, the 1m realtime chart keeps streaming, and all other REST requests start returning 401 without a visible redirect to login.

### Triage

The auth store still reports `isValid: false` once the token expires, but `App` only listens to `onChange`. Because natural JWT expiration does not trigger `onChange`, `$authenticated` stays `true` and `Layout` keeps showing the dashboard.

### Fix

In `src/main.tsx`, following @henrygd's [review suggestion](https://github.com/henrygd/beszel/pull/2310#pullrequestreview-5133404890):

1. Register a `pb.afterSend` hook that runs a debounced `verifyAuth()` (100 ms, via the existing `debounce` util) whenever any API response returns 4xx while a token is set. The `auth-refresh` URL is excluded so `verifyAuth()`'s own failing refresh (which returns 401 when the token is no longer valid) can't loop.
2. `verifyAuth()` then calls `logOut()` on the failed refresh, which clears the auth store, fires `onChange`, and updates `$authenticated` so `Layout` renders `LoginPage` with a "Failed to authenticate" toast.

Also removed per the same review: the empty-records `verifyAuth()` heuristic in `systemsManager.refresh()` — the hook now catches the actual 401 from the systems fetch instead of inferring auth trouble from an empty list.

### Verification

- `bunx biome check src/main.tsx src/lib/systemsManager.ts` — clean
- `bunx tsc --noEmit` — clean
- `bun run build` inside `internal/site` — production build succeeds
- Loop-guard check against the installed `pocketbase` ^0.26.2 source: `afterSend` fires for every `send()` response (including 4xx, before the `ClientResponseError` throw) and its return value is used as the response data, so the hook returns `data` unchanged.

## Documentation

### Fixed

- Redirect to the login page when the JWT expires.

## Screenshots

No UI/UX changes to screenshot; the only visible behavior is the redirect to the login page on expiry.

Fixes #2293

AI-assisted. I wrote and verified this change.

### DIFF
--- a/internal/site/src/lib/api.ts
+++ b/internal/site/src/lib/api.ts
@@ -4,7 +4,7 @@ import { basePath } from "@/components/router"
 import { toast } from "@/components/ui/use-toast"
 import type { ChartTimes, UserSettings } from "@/types"
 import { $alerts, $allSystemsById, $allSystemsByName, $userSettings } from "./stores"
-import { chartTimeData } from "./utils"
+import { chartTimeData, debounce } from "./utils"
 
 /** PocketBase JS Client */
 export const pb = new PocketBase(basePath)
@@ -12,7 +12,7 @@ export const pb = new PocketBase(basePath)
 export const isAdmin = () => pb.authStore.record?.role === "admin"
 export const isReadOnlyUser = () => pb.authStore.record?.role === "readonly"
 
-export const verifyAuth = () => {
+const verifyAuth = () => {
 	pb.collection("users")
 		.authRefresh()
 		.catch(() => {
@@ -24,6 +24,23 @@ export const verifyAuth = () => {
 			})
 		})
 }
+
+const verifyAuthDebounced = debounce(verifyAuth, 100)
+
+// verify the session whenever any API request returns a 4xx response (e.g. an
+// expired JWT). The auth-refresh endpoint is excluded to avoid a loop, since
+// it returns 401 itself when the token is no longer valid.
+pb.afterSend = (response, data) => {
+	if (
+		(response.status === 401 || response.status === 403) &&
+		pb.authStore.token &&
+		!response.url.includes("auth-refresh")
+	) {
+		verifyAuthDebounced()
+	}
+	return data
+}
+
 
 /** Logs the user out by clearing the auth store and unsubscribing from realtime updates. */
 export function logOut() {
--- a/internal/site/src/lib/api.ts
+++ b/internal/site/src/lib/api.ts
@@ -41,7 +41,6 @@ pb.afterSend = (response, data) => {
 	return data
 }
 
-
 /** Logs the user out by clearing the auth store and unsubscribing from realtime updates. */
 export function logOut() {
 	$allSystemsByName.set({})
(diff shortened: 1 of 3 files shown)


henrygd (commented):
Thanks, however maybe we can simplify it by using [`pb.afterSend`](https://github.com/pocketbase/js-sdk#send-hooks) like this:

```ts
const verifyAuthDebounced = debounce(verifyAuth, 100)

pb.afterSend = function (response, data) {
  if (response.status >= 400 && pb.authStore.token && !response.url.includes("auth-refresh")) {
    verifyAuthDebounced()
  }
  return data
}
```

That way it automatically checks when any 4xx response is encountered. The URL check should avoid loops, since the auth refresh endpoint returns 401 itself when auth isn't valid.

We could also remove the `verifyAuth` check in `systemsManager.ts` with this change.

santhiprakash:
Done — switched to the `pb.afterSend` approach on [`fff74da`](https://github.com/henrygd/beszel/pull/2310/commits/fff74daeed6e25f61d584c3fcbf58848a1898c67).

Your read of the SDK was right, and I verified it against the installed `pocketbase` ^0.26.2 source before rewriting: `afterSend` is invoked for every `send()` response *before* the `status >= 400` throw, so failed responses do reach the hook, and its return value replaces the parsed data (hence the `return data`). The `auth-refresh` exclusion holds as you described — `verifyAuth()`'s own failing refresh 401s with `auth-refresh` in the URL, so no loop.

What changed on top of your snippet:

- Used the existing `debounce` util from `src/lib/utils.ts` for `verifyAuthDebounced`, and kept the hook at module scope in `main.tsx` so it isn't re-registered across `App` mount/unmount cycles (the login redirect unmounts `App`).
- Removed the empty-records `verifyAuth()` call in `systemsManager.refresh()` as you suggested — the hook now catches the actual 401 from the systems fetch rather than inferring auth trouble from an empty list.
- Wrote the hook as an arrow function (biome's `useArrowFunction` flags the `function` expression); semantics are identical.

Net −44/+14 vs the previous timeout/interval scheme.

Verification: `bunx biome check src/main.tsx src/lib/systemsManager.ts` clean · `bunx tsc --noEmit` clean · `bun run build` in `internal/site` succeeds.

henrygd (approved):
Thank you!

PR #2313, opened 2026-09-08
fix(agent): SKIP_GPU excludes GPU hwmon from temperatures and fans

miloszkolber:
## Problem

`SKIP_GPU=true` stops GPU collectors, but two other collection paths still open GPU sensor files on every poll:

1. **Temperatures** — gopsutil reads *every* `hwmon*/temp*_input` before results can be filtered, so `SENSORS=-xe*` filters only the reporting, not the reads.
2. **Fans** — `discoverHwmonFans` walks `/sys/class/hwmon` directly and ignores `SYS_SENSORS`, `SENSORS`, and `SKIP_GPU`.

On drivers like Intel Xe each sensor-file read is a runtime PM resume, so the card never idles even with `SKIP_GPU=true`.

## Fix

- `agent/sensors.go` — with `SKIP_GPU=true` (and no explicit `SYS_SENSORS`), point gopsutil at a shadow sysfs tree built in-process containing symlinks to non-GPU `hwmon` devices only. Only static `name` files are read while building it. An explicit `SYS_SENSORS` path is still respected as-is. GPU keys are additionally dropped from reported temperatures as a backstop.
- `agent/fans.go` — record the chip name during discovery and drop GPU chips *before* any `fan*_input` is opened.
- `agent/gpu.go` — `gpuHwmonChips` list (`xe`, `i915`, `amdgpu`, `radeon`, `nvidia`, `nouveau`) with exact-match helpers next to the existing `SKIP_GPU` handling, plus a log line. No behavior change unless `SKIP_GPU=true`.

## Verification

- New tests in `agent/sensors_test.go` / `agent/fans_test.go` (shadow contents, post-read filter both modes, fan filter incl. shared-cache non-mutation, `SKIP_GPU` env wiring). `go build`, `go vet`, `gofmt` clean.
- Built the agent and ran it live against a host with Intel Xe (`hwmon8`): shadow contains `hwmon0`-`hwmon7`, no `xe` temperatures or fans reported, and over a full 75s poll cycle PCI `0000:04:00.0` stayed `suspended` with `runtime_active_time` frozen — zero GPU wakeups while the agent stayed healthy and reporting.
- Full suite: all green except `TestGPUCapabilitiesAndLegacyPriority/no_gpu_tools_available`, which fails identically on clean `HEAD` (environmental — the test host has a real GPU).

## Notes

- The shadow is built once at agent startup, so hot-plugged devices need a restart to appear, and stale dirs can accumulate across restarts on bare-metal installs.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -243,7 +243,11 @@ func (a *Agent) gatherStats(options common.DataRequestOptions) *system.CombinedD
 // Start initializes and starts the agent with optional WebSocket connection
 func (a *Agent) Start(serverOptions ServerOptions) error {
 	a.keys = serverOptions.Keys
-	return a.connectionManager.Start(serverOptions)
+	err := a.connectionManager.Start(serverOptions)
+	if err != nil {
+		a.cleanupSensorShadow()
+	}
+	return err
 }
 
 func (a *Agent) getFingerprint() string {
--- a/agent/connection_manager.go
+++ b/agent/connection_manager.go
@@ -154,6 +154,7 @@ func (c *ConnectionManager) Start(serverOptions ServerOptions) error {
 func (c *ConnectionManager) stop() error {
 	_ = c.agent.StopServer()
 	c.closeWebSocket()
+	c.agent.cleanupSensorShadow()
 	return health.CleanUp()
 }
 
--- a/agent/fans.go
+++ b/agent/fans.go
@@ -12,7 +12,7 @@ import (
 )
 
 type fanSensor struct {
-	key, path string
+	key, path, chip string
 }
 
 var getFanSensors = newFanSensorCache(hwmonRoot)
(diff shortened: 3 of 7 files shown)


henrygd (changes requested):
Thanks for raising this. Looks good, but there is an edge case with `SYS_SENSORS`, which overrides the sysfs root gopsutil uses.

Right now, if it’s set, it bypasses the filtered shadow, so `SKIP_GPU=true` could still read GPU sensor files.

Could we use `common.HostSysWithContext(config.context)` to get the effective sysfs root, then build the shadow from that when `SKIP_GPU` is enabled?

```go
sysRoot := common.HostSysWithContext(config.context)
shadow, err := buildNonGpuSysShadow(sysRoot)
```

That should keep `SYS_SENSORS` working while still avoiding GPU hwmon reads.

miloszkolber:
Thanks — updated in e23c3a4. The shadow is now built from the effective `SYS_SENSORS` root, so `SYS_SENSORS` and `SKIP_GPU` work together. Added `TestSkipGpuShadowUsesSysSensorsRoot` to cover it.

henrygd (approved):
Thanks, looks good! I made one very small change to fall back to hwmonN/device/name, but otherwise I think we're good to merge.

PR #2315, opened 2026-09-08
feat(agent): report btrfs filesystems as storage pools

anaisbetts:
This PR adds btrfs storage volume support alongside ZFS support:

image
image

Also makes a few small copy changes to indicate it's not just ZFS any more (will need an i18n update)

### ➕ Added

- `agent/btrfs`: sysfs reader for btrfs filesystems (Linux; no-op elsewhere)
- Btrfs filesystems appear in pool usage/I/O charts, the pools table and health alerts

### ✏️ Changed

- `ZfsManager` sources are wrapped with `withBtrfs`, which appends btrfs entries and tolerates missing ZFS utilities on btrfs-only hosts
- User-facing "ZFS pool" wording (card title, chart descriptions, alert titles/history) now says "storage pool"; the two chart msgids changed, so locales need re-extraction


henrygd (commented):
Thanks, I agree it would be nice to add support for this in 0.19.x.

I like the design, there's maybe one or two small things I'd tweak, which I can do on my end. However, I tested with a RAID 1 pool and it's showing the combined device capacity (20GB) rather than real usable space (10GB).

I'll look further into it when I have a chance (probably tomorrow) 👍

anaisbetts:
@henrygd Edit away, thanks for having a look!

henrygd:
Merged. Thanks again for your work. Let me know if my changes broke anything on your end.

PR #2316, opened 2026-09-08
fix(agent): don't warn about unset HUB_URL in SSH-only mode

alecrubin:
An agent with only `KEY` set runs in SSH-only mode, where the hub dials the agent and nothing outbound is expected. `newWebSocketClient` still returns an error in that case because `HUB_URL` isn't set, and `ConnectionManager.Start` logs everything it returns at warn level, so those agents print this on every start:

```
WARN Error creating WebSocket client err="HUB_URL environment variable not set"
```

Nothing is actually wrong, it's a supported config being reported as a failure. I have eight agents set up this way and the line has been in their logs since the day I installed them.

This adds an `errNoHubURL` sentinel and logs that one case at debug. A malformed `HUB_URL` or a missing token still warns exactly as before, and the message text doesn't change, so the existing table test still covers it. The same block special-cases `caCertFileError` three lines up, so it follows that shape.

The only workaround right now is `LOG_LEVEL=error`, which also drops `INFO SSH connected` and `INFO SSH connection established`. Those are the lines you want when you're checking whether the hub is actually reaching an agent, so trading them away to silence this one is a bad deal.

The test asserts `errors.Is(err, errNoHubURL)`. Putting the old `errors.New` back makes it fail, so it isn't just matching on the message string.

Ran `gofmt`, `go vet`, and `go test -tags testing ./agent/...` on darwin. The GPU collector tests fail there with or without this change since nvidia-smi and friends aren't installed, and the failure set is identical both ways.


henrygd (changes requested):
Thanks, however we'd need to do something like this to make sure we show a warning when `DISABLE_SSH=true` and `HUB_URL` is unset. Otherwise there would be no connection method and no warning.

```go
disableSSH, _ := utils.GetEnv("DISABLE_SSH")
if errors.Is(err, errNoHubURL) && disableSSH != "true" {
	slog.Debug("WebSocket client not configured", "err", err)
} else {
	slog.Warn("Error creating WebSocket client", "err", err)
}
```

alecrubin:
Good catch, `DISABLE_SSH=true` with no `HUB_URL` would have left an agent with no connection method and nothing louder than debug saying so. Pushed your version.

One thing I checked before copying it: the `!= "true"` comparison is the same literal check `server.go` uses to decide whether the SSH server actually starts, so the gate and real SSH availability can't disagree on odd values. Someone setting `DISABLE_SSH=1` gets SSH enabled everywhere and the quiet log here, which is right since the hub can still dial in.

Reran `gofmt`, `go vet`, and `go test -tags testing ./agent/...` on darwin, same result as before.

henrygd (approved):
Looks good, thanks!

Issue #1098, opened 2025-08-28
[Feature]: Custom Email / Email Templates


princessskye:
### Component

Hub

### Describe the feature you would like to see

The notification Emails are a bit plain. I know its just as a notification, but i would really like to use custom HTML for the Notification Emails. 

That would be a really nice feature!

### Motivation / Use Case

It is not necessery but it would be nice to be able to customize Notification Emails

### Describe how you would like to see this feature implemented

Either as an UI Option or just a point in the right direction how to change that in the container / Supabase.

### Category

Settings & Configuration

### Affected Metrics

Other

svenvg93:
Thanks for taking the time to suggest this!

This issue is a duplicate/similar of an existing one, so we’ll be closing it to keep the discussion in a single place. You can follow progress and updates on the linked issue, and feel free to subscribe there or add any additional details that might help.

Thanks again for helping improve the project!

PR #2317, opened 2026-09-08
feat(alerts): customizable notification templates per alert type

### LINKED ISSUE #604: [Feature]: Customize notification message per notification type

I'd like to customize (i.e. add more details like a timestamp) some notifications that are sent.

Obviously this would need to be on a per notification type since the message content is different for each type.
Thanks

yusufgngor:
## 📃 Description

Adds user-customizable notification templates. Closes #604 (the HTML email request from #1098 can follow on top of this).

Users can edit the **title and body** of every notification kind — status, system metrics, container health, SMART, systemd services, storage pools (ZFS/Btrfs) — on the Notifications settings page. Templates use the same `{placeholder}` syntax as the PocketBase mail templates in the admin UI, plus an optional inline wording map (`{state:down=lost,up=restored}`): no template engine, no conditionals, every value pre-formatted. Unknown placeholders are left in place so typos are visible. An unchanged field keeps the built-in English text (title and body independently), so nothing changes for users who don't touch the page.

**Placeholders**
- Common: `{title}` `{message}` `{link}` `{link_text}` `{state}` `{time}` `{date}` `{clock}` `{time_iso}` `{timezone}` and `{system.name}` `{system.host}` `{system.port}` `{system.status}` `{system.hostname}` `{system.os}` `{system.kernel}` `{system.cpu}` `{system.cores}` `{system.arch}` `{system.agent_version}` `{system.uptime}` (from `systems` / `system_details`, read only when a template references `{system.`).
- System metrics: `{metric}` `{descriptor}` `{value}` `{unit}` `{threshold}` `{minutes}`; container: `{containers}` `{count}` `{logs}`; systemd: `{services}` `{count}`; SMART: `{device}` `{model}` `{old_state}` `{new_state}`; storage pools: `{pool}` `{old_health}` `{new_health}`.
- `{state}` carries `down/up`, `above/below`, `unhealthy/healthy`, `failed/recovered` or the new SMART / storage pool state. Any value can be reworded inline — `{state:down=lost,up=restored}` — which covers the "native language" part of #604 without conditionals.

**UI.** Each kind is a collapsible card whose fields show the text currently in use (a close placeholder equivalent of the built-in message), so users simply edit it; text equal to the built-in equivalent is not stored and Reset restores it. Placeholder chips insert at the cursor. Preview renders the template with sample data; Send test email delivers that preview.

**Time zone.** `{time}`/`{date}`/`{clock}` use a new per-user `notificationTimezone` setting and honour the existing 12h/24h preference. When unset, the hub's own time zone (`TZ`) is used. The Docker image is built with `-tags timetzdata` so zones resolve in the scratch image (+~450 KB there only; bare-metal builds keep using the OS zoneinfo).

**Link handling.** A custom body owns link placement (`{link}`), so the link is not appended a second time — for email and for Shoutrrr services without native link support. Services with native links (ntfy, lark, bark) still receive it as a parameter.

**API.** `POST /api/beszel/notification-templates/preview` `{kind, title, body, timezone, sendEmail}` renders the submitted template with sample data (the user's first system is used for `{system.*}`) and, with `sendEmail: true`, mails the result — to the account address for regular users, to the configured notification addresses for admins; read-only users cannot send. This doubles as the email test button that was missing.

**Validation.** `user_settings` create/update requests are validated: known template kinds, ≤ 2000 characters per field, loadable timezone. Rendered output is capped (500 / 8000 characters) so placeholder repetition cannot amplify a notification.

**Implementation notes.** All rendering logic lives in `internal/alerts/alerts_templates.go` (+ `alerts_templates_api.go`); `SendAlert` gained a single `applyUserTemplates` call and the alert builders pass `Kind`/`State`/`Vars` alongside the unchanged built-in text, so existing notifications are byte-identical when no template is set. The UI is a new `notification-templates.tsx` component wired into the notifications page with a few lines, which should rebase cleanly on #2005. Generated locale files are not included, as recommended in the developer guide; the new strings will flow through Crowdin. Tests cover rendering, the inline map, timezone, link handling, the preview API (roles) and the validation hook (`go test -tags 'testing no_ui' ./internal/alerts/`).

This is a sizeable change; happy to split it (plumbing → rendering + validation → preview endpoint → UI) or adjust the design (e.g. Go `text/template` instead of placeholders, admin-only) if you'd prefer something different.

## 📖 Documentation

henrygd/beszel-docs#80 — adds a "Custom templates" section to the notifications guide.

### ➕ Added

- Notification templates per alert kind with `{placeholder}` substitution and inline wording maps (`{state:down=lost,up=restored}`)
- Per-user `notificationTimezone` setting used for `{time}` / `{date}` / `{clock}` (hub time when unset)
- `POST /api/beszel/notification-templates/preview` (preview + test email)
- Settings UI: template cards prefilled with the active text, placeholder chips, preview, test email, timezone select
- Server-side validation of notification template settings
- Tests for rendering, the inline map, timezone, link handling, the preview API and the validation hook

### ✏️ Changed

- `AlertMessageData` carries `Kind`, `State`, `Vars` and `OmitLink`; the email/Shoutrrr link append is skipped when a custom body is used
- `Collapsible` accepts a `ReactNode` title (used for the "Customized" badge)
- Hub Docker image is built with `-tags timetzdata`
- Settings page shows the server's message when a save is rejected

## 📷 Screenshots

Settings → Notifications: timezone and one card per alert kind

Notification templates overview

Status card: prefilled text edited with an inline wording map, placeholder chips and the rendered preview (sample data)

Status template preview


henrygd (commented):
Thank you, this is a nice feature and the implementation makes sense. There are some things I'd want to tweak, but I can do that on my end.

I did rebase to main, so please pull again if you want to make any other changes.

Can I ask what the reasoning is for the test email cooldown? It seems unnecessary to me. I'm not sure why we'd limit this.

yusufgngor:
The idea actually started because I wanted an easier way to see which machine was causing an alert, specifically its IP address and operating system.
I'm using Beszel in an environment with many machines, and email is my primary alerting channel, so having those details directly in the email helps me react much more quickly without needing to open the dashboard first. 

As for the test email cooldown,it was an AI review,  in a multi-user setup any regular user can hit the preview endpoint and trigger a real email, so a small limit would stop someone from flooding the SMTP server by clicking the button repeatedly. If you think it is unnecessary I can remove it, or we can shorten it to a few seconds, whichever you prefer.

Thanks for creating and maintaining Beszel

henrygd:
Thanks, it looks like `readonly` users are excluded from sending test emails anyway, so I think it's fine to remove the cooldown.

One other thing as well, we added support for Btrfs storage pools yesterday, so the alerts are not ZFS specific anymore. The file is still called `alerts_zfs.go` but the verbiage is using 'Storage pool` now rather than 'ZFS pool'.

```go
title := fmt.Sprintf("Storage pool %s on %s: %s", newHealth, systemName, poolName)
message := fmt.Sprintf("Storage pool %s (%s) was first observed as %s", poolName, systemName, newHealth)
if oldSeverity > 0 {
	message = fmt.Sprintf("Storage pool %s (%s) health changed from %s to %s", poolName, systemName, oldHealth, newHealth)
}
```

If you can change those two things then I can pick it up from there.

yusufgngor:
Hi @henrygd, i did the changes you mentioned, you can check

Notes: 

- removed the 30 s test-email cooldown (only the read-only check remains)
- storage pool wording now matches `alerts_zfs.go` ("Storage pool …"); the template kind key stays `zfs` to match the file/alert naming, and the UI card is labelled "Storage pools (ZFS, Btrfs)"

While at it I also dropped the `notificationLabels` map (it had no UI and clashed with the inline `{state:down=…}` wording map, which covers the same need), so the settings JSON now only carries `notificationTemplates` and `notificationTimezone`. Rebased on `main`, single commit, tests green. Docs PR henrygd/beszel-docs#80 is updated to match.

henrygd:
Thank you. Let me get #2266 done first then I'll get back to this. I don't want to add any larger features before that gets across the finish line.

PR #2319, opened 2026-09-09
fix(agent): skip disk.PartitionsWithContext on Windows to prevent startup hang (#2289)

### LINKED ISSUE #2289: [Bug]: v0.19.0 Unexpected Status SERVICE_STOPPED In Response To START Control
### Welcome!

### Component

Agent

### Problem Description

Since updating the agent to v0.19.0 on a windows client pc NSSM is not able to start the Beszel service with an error: Unexpected Status SERVICE_STOPPED In Response To START Control

### Expected Behavior

In all previous releases all i had to do is stop the service via NSSM stop beszel, replace the agent executable and start the service again.

### Steps to Reproduce

1. stop the beszel agent via NSSM stop 
2. replace the agent executable with the latest one v0.19.0
3. start the agent via NSSM start

### OS / Architecture

Windows 8.1 x64

### Beszel version

0.19.0

### Installation method

Binary

### Configuration

```yaml
KEY=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
TOKEN=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
HUB_URL=http://172.18.0.13:8090
DISABLE_SSH=true
EXTRA_FILESYSTEMS=E:,Z:
SENSORS=CPU Package
NICS=vEthernet (Internal)
LHM=true
EXCLUDE_SMART=/dev/sda
SKIP_GPU=true
SKIP_SYSTEMD=true
```

### Hub Logs

```json

```

### Agent Logs

```shell
Nothing is registered in logs.
```

santhiprakash:
## Description

This PR fixes #2289 by avoiding gopsutil's `disk.PartitionsWithContext` on Windows during agent startup.

`disk.PartitionsWithContext` enumerates all volumes, including external USB and mapped network drives. On Windows, `GetVolumeInformationW` can hang indefinitely or panic when a volume is unresponsive or spinning down. This leaves the agent stuck before it can write any logs, which matches the `SERVICE_STOPPED` symptom reported when the agent is managed by NSSM.

The fix skips `disk.PartitionsWithContext` on Windows. The agent already has fallbacks to discover the root filesystem via `disk.IOCounters` (most active I/O device) and to register `EXTRA_FILESYSTEMS` entries via `disk.Usage`, so monitoring continues to work for the root device and any explicitly configured drives.

A regression test is added to verify that `initializeDiskInfo` does not call the partition discoverer on Windows and still registers a root filesystem.

### Fixed

- Agent no longer hangs or crashes on Windows when an unresponsive drive is present.

### Changed

- `disk.PartitionsWithContext` is skipped on Windows; `EXTRA_FILESYSTEMS` is now the preferred way to monitor non-root drives on Windows.

## Testing

- `go test -tags='testing no_ui' ./agent` passes.
- `go test -tags='testing no_ui' ./internal/cmd/agent ./agent/...` passes.
- `GOOS=windows GOARCH=amd64 go vet ./agent` passes.
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build -o /tmp/beszel-agent-windows.exe ./internal/cmd/agent` produces a valid `PE32+ executable (console) x86-64`.

Full `go test ./...` has a pre-existing environment/CI issue (`internal/site/dist` missing and a `systemd` CPU percent test that is sensitive to machine core count); these are unrelated to this change.

### DIFF
--- a/agent/disk.go
+++ b/agent/disk.go
@@ -15,14 +15,22 @@ import (
 	"github.com/shirou/gopsutil/v4/disk"
 )
 
+// getDiskPartitions is a test seam for disk.PartitionsWithContext.
+var getDiskPartitions = disk.PartitionsWithContext
+
+// skipDiskPartitionsOnWindows avoids calling gopsutil's disk.PartitionsWithContext
+// on Windows, where it can hang or panic on unresponsive volumes. See
+// https://github.com/henrygd/beszel/issues/2289 and PR #2047.
+var skipDiskPartitionsOnWindows = runtime.GOOS == "windows"
+
 // fsRegistrationContext holds the shared lookup state needed to resolve a
 // filesystem into the tracked fsStats key and metadata.
 type fsRegistrationContext struct {
-	filesystem         string // device part of optional FILESYSTEM env var
-	filesystemName     string // optional custom name from FILESYSTEM=device__name
-	isWindows          bool
-	efPath             string // path to extra filesystems (default "/extra-filesystems")
-	diskIoCounters     map[string]disk.IOCountersStat
+	filesystem     string // device part of optional FILESYSTEM env var
+	filesystemName string // optional custom name from FILESYSTEM=device__name
+	isWindows      bool
+	efPath         string // path to extra filesystems (default "/extra-filesystems")
+	diskIoCounters map[string]disk.IOCountersStat
 }
 
 // diskDiscovery groups the transient state for a single initializeDiskInfo run so
@@ -306,9 +314,13 @@ func (a *Agent) initializeDiskInfo() {
 	hasRoot := false
 	isWindows := runtime.GOOS == "windows"
 
-	partitions, err := disk.PartitionsWithContext(context.Background(), true)
-	if err != nil {
-		slog.Error("Error getting disk partitions", "err", err)
+	var partitions []disk.PartitionStat
+	var err error
+	if !skipDiskPartitionsOnWindows {
+		partitions, err = getDiskPartitions(context.Background(), true)
+		if err != nil {
+			slog.Error("Error getting disk partitions", "err", err)
+		}
 	}
 	slog.Debug("Disk", "partitions", partitions)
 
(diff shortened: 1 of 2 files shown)


henrygd:
Thanks for looking into this. Do we have confirmation that `disk.PartitionsWithContext` is actually causing #2289?

#2047 did reproduce a hang in this call, but that gopsutil bug was fixed upstream in 4.26.4, and v0.19.0 is already using 4.26.8. The linked issue has no logs or other indication that partition discovery is where startup fails.

It's also running Windows 8.1, which isn't supported by current Go versions, so there are other plausible causes for the v0.19.0 regression.

santhiprakash:
No — we can't confirm that, and after re-checking your points we agree the premise is unproven for v0.19.0.

The causal chain in this PR rested on the #2047 repro of `PartitionsWithContext` hanging, but that was against gopsutil v4.25.x — before v4.26.4 shipped "Fix infinite loop when failed to find path for volume on Windows" (shirou/gopsutil#2066). We checked `go.mod` at `v0.19.0` and it already pins v4.26.8, so that specific hang class should be gone. You're also right that #2289 has no logs tying startup to partition discovery, and the Windows 8.1 detail is the cleaner explanation: the module targets `go 1.27.1`, well past the Windows 8/8.1 support drop in Go 1.21, so a startup regression on that machine after the toolchain bump doesn't need `PartitionsWithContext` as the cause at all.

Given #2289 is resolved as a Go platform-support issue, we'd recommend closing this PR — the change would also move extra Windows drives behind `EXTRA_FILESYSTEMS` by default, which isn't a trade worth making without confirmed causality. Happy to close it on our side if you agree. If a hang in this call ever reproduces on a supported Windows version, the fallback approach here could be worth revisiting then.

santhiprakash:
Closing this after [#2289](https://github.com/henrygd/beszel/issues/2289) was closed as a Go platform-support matter (Windows 8.1 is no longer supported by the Go version used in v0.19.0).

As discussed above, we don't have confirmation that `disk.PartitionsWithContext` is the actual cause on v0.19.0, so the fallback change isn't justified on its own.

PR #2325, opened 2026-09-11
fix(zfs): skip zpool list when /dev/zfs unavailable inside the container

### LINKED ISSUE #2324: [Bug]: Alpine agent repeatedly logs zpool timeouts when /dev/zfs is not mapped
### Welcome!

### Component

Agent

### Problem Description

On a Proxmox host with no ZFS pools, beszel-agent:alpine still runs zpool list every minute.

The host itself returns immediately with:

no pools available

But inside the container, when /dev/zfs is not mapped, Beszel waits until its 10-second timeout:

DEBUG Storage pool stats unavailable backend=zfs err="zpool list: zpool timed out after 10s: context deadline exceeded"

This also correlates with the WebSocket reconnect issue I reported previously:

[https://github.com/henrygd/beszel/issues/2294](https://github.com/henrygd/beszel/issues/2294)

The documentation already states that /dev/zfs must be mapped for ZFS support, so missing ZFS data itself is expected.

The issue is that when /dev/zfs is not mapped, the Alpine agent retries zpool list every minute, waits for the full 10-second timeout each time, and repeatedly logs:

DEBUG Storage pool stats unavailable backend=zfs err="zpool list: zpool timed out after 10s: context deadline exceeded"

This creates unnecessary log noise and also delays collection.

It would be better to detect that /dev/zfs is unavailable and skip ZFS collection immediately, or only log the condition once instead of repeating it every minute.

### Expected Behavior

If /dev/zfs is unavailable inside the container, ZFS detection should be skipped instead of running zpool until timeout.

Alternatively, an option like DISABLE_ZFS=true would also solve this.

### Steps to Reproduce

Run beszel-agent:alpine on a Proxmox host with no ZFS pools.
Do not map /dev/zfs into the container.
Enable LOG_LEVEL=debug.
Run on the Proxmox host:

zpool list

It returns immediately with:

no pools available

Run inside the Beszel container:

docker exec beszel-agent zpool list

It hangs until the command timeout.

Observe the agent log:

DEBUG Storage pool stats unavailable backend=zfs err="zpool list: zpool timed out after 10s: context deadline exceeded"

Map /dev/zfs into the container and restart the agent.
Run again:

docker exec beszel-agent zpool list

It should now return immediately instead of timing out.

### OS / Architecture

linux/amd64(agent)

### Beszel version

0.19.0

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

santhiprakash:
## 📃 Description

On a Linux host without ZFS pools, the Alpine agent image still ships the `zpool` binary. If `/dev/zfs` is not mapped into the container, `zpool list` blocks for the full 10-second command timeout and then `PoolStats()` returns a wrapped `context deadline exceeded` error. Because `optionalPoolSource()` does not treat that as an absent backend, the agent retries every minute (each `poolStatsRefreshInterval` tick) and logs:

```
DEBUG Storage pool stats unavailable backend=zfs err="zpool list: zpool timed out after 10s: context deadline exceeded"
```

This mirrors the existing `PoolKernelStats()` pattern: on Linux, probe the kernel interface (`/dev/zfs`) first. If the device node is missing, return `ErrNoZfs`, which `optionalPoolSource()` already converts to a silent no-op backend. This removes both the 10-second hang and the repeating debug log.

Changes:
- `agent/zfs/zfs.go`: rename the cross-platform `PoolStats()` to private `poolStats()`.
- `agent/zfs/zfs_linux.go`: add `PoolStats()` with a fast `os.Stat("/dev/zfs")` probe; if absent, return `nil, ErrNoZfs`; otherwise delegate to `poolStats()`.
- `agent/zfs/zfs_freebsd.go`: add `PoolStats()` delegating directly to `poolStats()`.
- `agent/zfs/zfs_unsupported.go`: add `PoolStats()` returning `errors.ErrUnsupported`.
- `agent/zfs/zfs_linux_test.go`: add sabotage-first regression tests proving `PoolStats()` does not invoke `zpool list` when `/dev/zfs` is missing and does delegate when it is present.

Fixes #2324

## 📖 Documentation

### 🔧 Fixed

- Skip `zpool list` when `/dev/zfs` is not available inside the container, preventing the 10-second timeout and once-per-minute debug log on hosts without ZFS pools.

## 📷 Screenshots

No UI changes.

AI-assisted. I wrote and verified this change.


henrygd (approved):
Thank you! 👍

PR #2327, opened 2026-09-11
feat(hub): add TRUSTED_PROXY_IPS allowlist for TRUSTED_AUTH_HEADER

### LINKED ISSUE #2306: [Feature]: Optional TRUSTED_PROXY_IPS allowlist for TRUSTED_AUTH_HEADER
### Component

Hub

### Description

Follow-up to a private security advisory discussion (GHSA-247h-jh7m-2qc4). @henrygd suggested I raise this as a feature request, since the behavior is opt-in and intended rather than a bug.

When `TRUSTED_AUTH_HEADER` is set, the hub authenticates a request based solely on the value of that header:

```go
// internal/hub/api.go
if trustedHeader, _ := utils.GetEnv("TRUSTED_AUTH_HEADER"); trustedHeader != "" {
    se.Router.BindFunc(func(e *core.RequestEvent) error {
        return authorizeRequestWithEmail(e, e.Request.Header.Get(trustedHeader))
    })
}
```

The header is trusted regardless of the request's source. This is correct and secure when every request genuinely passes through the reverse proxy (Authelia/Authentik/etc.), which is the documented and intended setup. The docs are clear that this is an opt-in bypass of built-in auth and the user is responsible for wiring it up safely.

**Proposed enhancement:** add an optional `TRUSTED_PROXY_IPS` environment variable (comma-separated IPs/CIDRs). When set, the hub only honors `TRUSTED_AUTH_HEADER` if the request's source address is in the allowlist; otherwise it ignores the header and falls through to normal auth. When `TRUSTED_PROXY_IPS` is unset, behavior is unchanged (fully backward compatible).

Rough sketch:

```go
if trustedHeader, _ := utils.GetEnv("TRUSTED_AUTH_HEADER"); trustedHeader != "" {
    allowed := parseCIDRs(utils.GetEnv("TRUSTED_PROXY_IPS")) // empty => allow all (current behavior)
    se.Router.BindFunc(func(e *core.RequestEvent) error {
        if len(allowed) > 0 && !sourceAllowed(e.Request.RemoteAddr, allowed) {
            return e.Next() // ignore the header, use normal auth
        }
        return authorizeRequestWithEmail(e, e.Request.Header.Get(trustedHeader))
    })
}
```

### Motivation / Use Case

`TRUSTED_AUTH_HEADER` relies entirely on the deployment guaranteeing that the hub is never reachable except through the proxy. In practice that guarantee is easy to lose: a Docker network that also exposes the hub port, a host-level `-p 8090:8090`, or an SSRF primitive elsewhere on the network can all let a request reach the hub directly, at which point setting the header impersonates any user.

An optional source-IP allowlist gives operators a simple defense-in-depth control that keeps the header feature working exactly as today for correct setups, while failing safe if the hub ends up reachable off-proxy. It stays entirely opt-in, so it doesn't change anything for existing users who don't set it.

Happy to open a PR if you're open to the approach.

jmrplens:
## 📃 Description

Closes #2306. It follows the sketch @tikket1 gave there.

With `TRUSTED_AUTH_HEADER` set, the hub authenticates a request from the header alone, whichever address it comes from. That is right when every request passes through the reverse proxy, and not when the hub can also be reached directly, since anyone who can reach it can set the header themselves.

This adds `TRUSTED_PROXY_IPS`, a comma-separated list of IPs or CIDR ranges. When set, the header is only honored on requests whose peer address is in the list; other requests fall through to the normal authentication. When unset, nothing changes.

Details:

- The check uses the connection's `RemoteAddr`, not `e.RealIP()`: PocketBase's trusted proxy setting only decides which forwarded header names the client, and that header is exactly what a direct connection can forge. The list names the proxy itself. IPv4-mapped IPv6 entries are treated as IPv4, and zones on the peer address are ignored.
- Entries that do not parse are skipped with a warning on the console, the way `config.yml` problems are reported. A list with no valid entry trusts nobody, so a typo narrows the allowlist instead of widening it.
- The list is parsed once when the middleware is registered.

Tests: `TestTrustedHeaderProxyAllowlist` runs the same `/api/beszel/getkey` scenario as `TestTrustedHeaderMiddleware` with the allowlist set (the test client connects from `192.0.2.1`): a range that contains the peer and the exact address both authenticate, a range that does not and a list with no valid entry both get 401. Against the current middleware those last two return 200. `TestParseTrustedProxies` and `TestIsTrustedProxy` cover the parsing (single addresses, masked CIDRs, IPv4-mapped entries, blanks, invalid entries, the `BESZEL_HUB_` prefix) and the peer matching (IPv6, zones, missing port, garbage). `go test -tags='testing no_ui' ./internal/hub/...` passes.

Checked end to end with the hub in a container on a user-defined Docker network, `TRUSTED_AUTH_HEADER=X-Remote-Email` and a user with that email: without `TRUSTED_PROXY_IPS`, both the host (peer `172.22.0.1`) and another container on the network get 200 from `/api/beszel/getkey` with the header. With `TRUSTED_PROXY_IPS=172.22.0.1` the host still gets 200 and the other container gets 401. With `TRUSTED_PROXY_IPS=proxy.internal` everyone gets 401, and `TRUSTED_PROXY_IPS="proxy.internal, ::ffff:172.22.0.1"` logs `Ignoring invalid TRUSTED_PROXY_IPS entry "proxy.internal"` on the console and behaves like the second case.

## 📖 Documentation

henrygd/beszel-docs#82

### ➕ Added

- `TRUSTED_PROXY_IPS`: optional allowlist of peer addresses from which `TRUSTED_AUTH_HEADER` is honored.

### DIFF
--- a/internal/hub/api.go
+++ b/internal/hub/api.go
@@ -2,7 +2,11 @@ package hub
 
 import (
 	"context"
+	"fmt"
+	"log"
+	"net"
 	"net/http"
+	"net/netip"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


henrygd (approved):
Looks good, thanks very much!

PR #2328, opened 2026-09-11
fix(site): switch theme live when system preference changes

rajtslegr:
## 📃 Description

When theme is set to "system", changing the OS light/dark preference did not update the app until a hard refresh. `ThemeProvider` read `prefers-color-scheme` only once, when the theme was applied.

This PR subscribes to `prefers-color-scheme` changes and applies the resolved theme immediately. It also exposes `resolvedTheme` through context, so components can react to the active theme instead of the stored preference (the login page used `theme`, which treats "system" as dark).

### 🔧 Fixed

- System theme (OS light/dark changes) now applies live, without a page refresh.
- Login page border color now follows the resolved theme when "system" is selected.

## 📷 Screenshots

No visual changes beyond the correct colors being applied live.


henrygd (approved):
Looks good, thank you!

PR #2331, opened 2026-09-12
fix(agent): skip ZFS collection when /dev/zfs is not mapped

santhiprakash:
When the alpine agent container starts without /dev/zfs mounted, commands like zpool list and zfs list hang until the 10s command timeout and log a debug error on every collection cadence. Probe /dev/zfs once at startup and short-circuit pool metrics, dataset usage, and detail collection when it is absent, so the agent no longer pays the timeout cost every minute.\n\nFixes #2324\n\n## Test plan\n- [x] go test -tags='testing' ./agent/zfs/... ./agent\n- [x] go vet ./agent/...\n- [x] make build-agent

### DIFF
--- a/agent/storage_pool.go
+++ b/agent/storage_pool.go
@@ -3,9 +3,11 @@ package agent
 import (
 	"errors"
 	"log/slog"
+	"os"
 	"os/exec"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


santhiprakash:
Closing in favor of #2325. Both PRs address #2324; #2325 uses the narrower, platform-specific `agent/zfs` build-tag path that mirrors the existing `PoolKernelStats` / `ARCSize` split, so I am consolidating review there.

PR #2333, opened 2026-09-13
fix(site): discard pending history when switching to live charts

ChangkeunJ:
Switching to the 1-minute live chart while a historical request is pending can replace live points with old data when that request finishes. The effect returns before incrementing the request ID, so the previous response still passes the stale-request check. Moving the increment above the early return invalidates it when entering live mode too.

Reproduced in Chromium using the actual hook and controlled PocketBase responses: hold a 1-hour request, switch to 1 minute, deliver a live CPU sample of 20%, then complete the historical request with a 30-minute-old sample of 81%. The original code replaces 20% with 81%; this change keeps 20%. These are sample responses, without a running hub or agent.

The browser regression check, Biome check, catalog compilation, and Vite production build pass.


henrygd (approved):
Thanks!

PR #2345, opened 2026-09-15
Clarify temperature alert threshold direction

VACInc:
## Summary
- Keep the original temperature alert sentence structure.
- Replace `averaged` with `above` when the maximum threshold is triggered.
- Replace `averaged` with `below` when the alert recovers below the threshold.
- Leave non-temperature alert wording unchanged.

This makes the direction immediately clear to the end user without unnecessary capitalization or extra wording.

Examples:
- `Highest sensor CPU above 81.25°C for the previous 5 minutes.`
- `Highest sensor CPU below 79.50°C for the previous 5 minutes.`

## Tests
- `go test ./internal/alerts`


henrygd:
The direction should be in the title of the alert, so this may be a different issue. Are the titles omitted for your alert service?

Screenshot_20260916-204836

VACInc:
Hmmm looks like they definitely are...but seems like they're omitted on certain devices without explicitly pulling the notification down? Don't love that 🤦‍♂️

Screenshot_20260916_211454_ntfy

PR #2347, opened 2026-09-16
Use trend-aware SMART health classification for ATA media errors

VACInc:
## Description

This is a follow-up to #2275.

The current rule promotes a drive from `PASSED` to `WARNING` whenever ATA SMART attribute 5, 197, or 198 has any nonzero raw value. That catches real problems earlier than the firmware's overall SMART result, but it also turns old cumulative values into permanent alarms. Attribute 5 in particular normally does not return to zero, so one old reallocation can leave an otherwise stable drive yellow forever.

I've worked in the storage industry, and I would not normally keep a drive in a permanent alarm solely because a cumulative counter is nonzero. A nonzero value can absolutely justify investigation, especially for pending or uncorrectable sectors, but operational severity should be dynamic: how quickly the value is changing, how long it has persisted, whether it cleared and returned, and whether multiple media-error counters are deteriorating together.

This moves the final ATA health classification to the hub, where Beszel can retain enough history to evaluate those conditions over time. The thresholds below are an explicit operational policy, not a claim that ATA vendors define universal raw-value thresholds. SMART attributes and raw formats remain vendor-specific.

## Behavior

The hub now stores a compressed 366-day change-point history for ATA attributes:

- 5, reallocated sector count
- 197, current pending sector count
- 198, offline uncorrectable sector count

It calculates 24-hour, 7-day, 30-day, 90-day, and 365-day deltas where history exists, plus prior 7-day/30-day deltas, consecutive increasing/nonzero samples, persistence, clear-and-return behavior, and acceleration across three 7-day windows.

### Warning policy

- Attribute 5: raw >= 50; delta >= 2/24h, 3/7d, 5/30d, 10/90d, or 20/365d; three increasing samples; or doubled 7-day/30-day growth above the supplied minimums.
- Attribute 197: nonzero for 24 hours, raw >= 5, growth in the configured windows, two nonzero or increasing samples, or clear-then-return.
- Attribute 198: any new/non-stable nonzero value, 24-hour/7-day growth, or two nonzero samples.
- Combined warnings cover coexisting 5/197/198 values and joint 5/197 growth.

### Critical policy

- Attribute 5: delta >= 10/24h, 20/7d, or 50/30d; raw >= 500 while still growing; five increasing samples; tripled 7-day growth; or three-window acceleration.
- Attribute 197: raw >= 20, delta >= 5/24h or 10/7d, nonzero for 168 hours, three increasing samples, or deterioration combined with 5/198.
- Attribute 198: raw >= 10, delta >= 2/24h or 3/7d, two increasing samples, or deterioration combined with 5/197.
- Combined critical rules cover simultaneous deterioration, pending plus uncorrectable sectors, and accelerating reallocations with pending/uncorrectable sectors.

`FAILED` and `UNKNOWN` reported by smartctl still take precedence. Otherwise the hub emits `PASSED`, `WARNING`, or the new `CRITICAL` state. Alert severity and the SMART UI understand `CRITICAL`, and the detail sheet shows the exact trigger reasons.

## Data handling

- Uses raw values, not normalized SMART values.
- Clamps negative deltas to zero.
- Ignores historical windows that do not exist yet.
- Resets history when the serial number changes.
- Does not accumulate history when a serial number is unavailable.
- Compresses unchanged samples and keeps one anchor sample across retention pruning.
- Latches warnings until the requested clean windows exist.
- Preserves attribute 198's nonzero raw value as information after its trend warning clears.
- Keeps the agent-side warning behavior for mixed-version compatibility; the upgraded hub replaces that coarse classification with the trend result.

## Existing installations

Beszel currently stores only the latest `smart_devices` record, not a historical SMART time series. The migration therefore cannot reconstruct 7/30/90/365-day deltas. On first update, the hub preserves the one pre-migration snapshot and timestamp already present in each record, then accumulates compressed history going forward. Missing windows are not interpreted as zero growth.

## Verification

- `go test ./internal/entities/smart ./internal/hub/systems ./internal/alerts ./agent`
- `go vet -tags='testing no_ui' ./internal/entities/smart ./internal/hub/systems ./internal/alerts ./agent`
- Production UI build with Bun/Vite
- Agent and hub binaries built successfully
- Simulated the current 85-device fleet snapshot: the only immediate classification changes are the three stable drives with attribute 5 values of 1, 1, and 5 moving from the coarse `WARNING` back to `PASSED`.

The complete `make test` suite also passed during validation. A later repeat hit the existing `TestMultipleSystemsWithSameUniversalToken` status race; the same failure reproduces on the untouched fork main and does not involve the SMART paths changed here.

## References

- smartmontools `smartctl` manual: https://www.smartmontools.org/static/doxygen/smartctl.8.html
- Google, *Failure Trends in a Large Disk Drive Population*: https://static.googleusercontent.com/media/research.google.com/en/us/archive/disk_failures.pdf
- Backblaze, *What SMART Stats Tell Us About Hard Drives*: https://www.backblaze.com/blog/what-smart-stats-indicate-hard-drive-failures/


henrygd:
Thanks, sounds like an interesting approach. I'll review when I have time, just want to link #2296 and #2308 here for context. I've been leaning towards reverting the change rather than maintaining our own decision engine because it seems difficult to account for all vendor inconsistencies, but I'm open to what you've done here.

VACInc:
@henrygd on second thought, you're right.  If you were to implement anything it should really be user defined (which right now the biggest context lacking is "what kinda media is it?" even then there's a lack.

I would say the best enhancement route would be allowing users to create rules.  But that's not a small change.  I'm down to spin up some ideas if you're interested.  Or I can just shut this down if you want to revert back.

Down to help ease the load either way.

henrygd:
Maybe an env var on the agent like `SMART_WARNINGS` and `attr:value` comma separated?

```env
SMART_WARNINGS=5:20,197:5,198:10
```

Which would warn on attribute 5 over 20, 197 over 5, 198 over 10?

That approach should be fairly straightforward to implement.

VACInc:
Definitely the most straight forward option.  I feel like I'd want that plus the historical data so people can at least go back and check the history so they know it increased drastically or it has been a steady drip.

Gives them the flexibility to automate at their own will beyond the initial alert.

Thoughts?

PR #2357, opened 2026-09-19
feat: show pending package updates

### LINKED ISSUE #617: [Feature]: Info on pending package updates

Could we get a feature that allows us to quickly see how many updates are available on the individual machines?

Just an extra column in the overview page displaying a number of pending package updates, and maybe a list of packages within the server info page.

anandghegde:
## 📃 Description

Closes #617. Shows pending package updates per system, following the plan agreed in the issue. For now this ships counts only.

The agent detects the package manager and runs a read-only check every hour in the background, like the image update check. The last result is cached, so a slow package manager never delays metrics:

- apt: `apt-get -s dist-upgrade`. Counts `Inst` lines that upgrade an installed package. A line counts as a security update if its origin is a `-security` suite. I used `dist-upgrade` rather than `upgrade` because `upgrade` leaves out kept-back kernel updates.
- dnf (dnf4 and dnf5): `dnf -q -C check-update` for the total and `... --security` for the security count. `-C` reads only the system cache, which also works as a non-root user.
- zypper: `zypper --no-refresh -q list-updates` for the total, plus `list-patches --category security` for the security count. The security number is a count of patches, not packages.
- pacman: `checkupdates`. I changed this from `checkupdates -n`, which only reads checkupdates' own temp DB. That DB is empty unless something else keeps it synced, so `-n` always reported 0 in my tests. Plain `checkupdates` syncs a private copy and never touches pacman's DB.
- apk: `apk --no-network -u list`.

The counts go in system info as `pu: [total, security]`. apk and pacman have no security data, so for them the array has only the total. I put the counts in `Info` rather than `Details`, following the existing `sv` services counts. That way the dashboard gets them with no migration and no extra fetch, and an alert can use them later. The hub stores them without any code change.

The agent skips the check on non-Linux hosts and inside containers (`/.dockerenv` or `/run/.containerenv`), because a container's package database isn't the host's. `PACKAGE_UPDATES_INTERVAL` was trivial to add, so I included it. It takes a Go duration, and `0` disables the check.

## 📖 Documentation

A docs PR is still needed: the new env var, the list of supported package managers, and a note that the Docker agent doesn't report updates.

### ➕ Added

- Agent: background check for pending package updates (apt, dnf, zypper, pacman, apk) and a `PACKAGE_UPDATES_INTERVAL` env var
- Dashboard: an "Updates" column (total, plus security when known), sortable and hideable like the other columns
- System page: an update count in the info bar, with a tooltip for security updates

## Testing

- Parser unit tests use real outputs I captured in containers: Debian 12, Ubuntu 22.04, Rocky 9 (dnf4), Fedora 42 (dnf5), openSUSE Leap 15.5/15.6, Arch, and Alpine 3.20. They also cover edge cases: new-dependency `Inst` lines, dnf obsoletes, and a zypper run with no updates.
- I ran all the commands as a non-root user in those containers (except Fedora, whose image has no su/setpriv), because the systemd agent runs as `beszel`. They all worked.
- End to end: I ran the linux/arm64 agent as `nobody` in a Debian 12 container, connected over SSH to a locally built hub. The system record's `info.pu` showed `[44, 5]`, which matches `apt-get -s dist-upgrade` in that container.
- `go test -tags 'testing no_ui'` passes for the hub packages. It also passes for the agent on Linux, apart from `TestDirectoryIsWritable`, which fails there because the container runs as root. `bun run build` works, and tsc and biome show no new diagnostics in the files I changed.
 
beszel-updates-dashboard
beszel-updates-system

## Follow-ups

- A table on the system page listing each package with its current and available version, maybe with a major/minor/patch column. The apt, zypper, pacman and apk outputs already include both versions. dnf's `check-update` doesn't give the current version.
- A "pending updates above N" alert. It needs a new alert name in the collection and its own handling, so I kept it out of this PR.


svenvg93:
Thanks for your work on this! 
Just my two cents on the UI. I think the systems table should be readable at a glance without much text, so here's a simpler take on the Updates column:

- Show only the update count ("1 update" / "12 updates"), or "Up to date" when there are none.
- Use a colored dot for the type, the same way the Services column does:
  - 🔴 security updates pending
  - 🟡 regular updates only
  - 🟢 up to date
- Sort by dot color first, then by count, so the sort order matches what's on screen.

I'd also drop the updates item from the system page's info bar. It fits better in the table, and the info bar stays cleaner without it. As it should contain only static system information and the uptime.

I tried this locally and it looks tidy. Happy to make a PR to your branch to add the changes1 

Green = Up to date 
Scherm­afbeelding 2026-09-23 om 17 55 24

Amber = Normal updates
Scherm­afbeelding 2026-09-23 om 17 55 36

Red = Security updates
Scherm­afbeelding 2026-09-23 om 17 55 53

anandghegde:
Thanks @svenvg93, good call. Done in 2622b4331dec5d3ee5a854dfae7bf4176db01474.

- The Updates column shows only the count ("1 update" / "12 updates"), or "Up to date" when there are none. It uses the same dot as the Services column: red when security updates are pending, yellow for regular updates only, green when up to date. For apk and pacman there's no security data, so they show yellow or green.
- Sorting goes by dot color first, then by count, so it matches what's on screen. Systems with no update data sort last, like the other columns.
- I removed the updates item from the system page info bar.

PR #2362, opened 2026-09-20
fix(i18n): extract bulk monitor format hint

spatiumstas:
## 📃 Description

Adds translation keys for the new section.


henrygd (approved):
Thanks!

PR #2363, opened 2026-09-20
feat(ui): render country flag emoji

spatiumstas:
## 📃 Description

Adds support for Twemoji. Particularly useful on Windows systems.
Used font: https://github.com/mozilla/twemoji-colr

## 📷 Screenshots

<details>
  <summary>before</summary>

</details>

<details>
  <summary>after</summary>

</details>


svenvg93:
Out of curiosity; how is this supposed to work ? Like are all country code converted to flags automatically or is it only in a certain format that is converted?

spatiumstas:
> Out of curiosity; how is this supposed to work ? Like are all country code converted to flags automatically or is it only in a certain format that is converted?

This fix simply allows the flag emoji to be visible on systems that don't display them. For example, it displayed correctly in MacOS, but not in Windows. You still need to insert the flag emoji.

henrygd (commented):
Thanks, would the font file from here work, so we don't need to build it ourselves when it needs to be updated?

https://cdn.jsdelivr.net/npm/country-flag-emoji-polyfill@0.1.10/dist/

Does it need to be first in the order to load properly?

It would be nice if there was a way in CSS to gate it so it doesn't load in browsers / OS where rendering flags is already supported. Otherwise it loads for everyone if the settings page is opened, even if they don't need it. Not a deal breaker, though. We may just remove the flags from the language selector.

PR #2366, opened 2026-09-20
fix(ui): prevent premature truncation of network monitor system names

zfdang:
## Description

Network Monitors can truncate short system names even when their column has room. For example, the character-count heuristic selects `ali` over `z1` and `w1`, but `ali` is narrower in the bundled proportional font. The absolutely positioned label contributes no width, so the hidden placeholder limits every name to `ali`'s width.

Place the placeholder and visible name in the same CSS grid cell so both contribute to intrinsic sizing. This preserves the existing width cap and ellipsis for genuinely long names. Also include `longestSystemName` in the memo dependencies so existing cells resize when that store changes.

## Validation

- Reproduced the original truncation in Chrome using the actual system-cell component, bundled Inter font, and production styles.
- All 12 browser layout checks pass after the fix: mixed-width Latin names, CJK text, narrow cells, long-name truncation, and changes to the longest-name store without changing the displayed system.
- `npm run build` passes.
- `bun test` passes: 4 tests.
- `npx biome check src/components/network-monitors-table/network-monitors-columns.tsx` and `git diff --check` pass.
- `tsc --build --force --pretty false` reports the same 30 existing diagnostics on both the unmodified base (`97e6f64b`) and this branch; no new diagnostics.


zfdang:
Screenshot-2026-09-20-20-43-14

henrygd:
Thanks very much for your fix. There were multiple PRs open for this and I went with #2411 which fixes at the function source. So I will close this one but I appreciate your contribution in any case. Let me know if it isn't fixed for you in the next version and I will look at this again.

PR #2375, opened 2026-09-21
fix: preserve status alerts through pending recovery

spatiumstas:
## 📃 Description

Handles the `pending → up` transition when a system reconnects after its connection settings are edited. Active status alerts are preserved during `pending` and resolved once the system is confirmed online.

### 🔧 Fixed

- Status alerts remaining active after a system transitions from pending to online.


henrygd (approved):
Looks good to me. Thanks a lot!

PR #2377, opened 2026-09-21
feat(ui): change loss to a red line

svenvg93:
## 📃 Description

The packet loss chart now renders using a fixed red (`--destructive`) line color instead of the default per-monitor color cycling, making loss more visually distinct as a warning signal regardless of monitor count.

### ✏️ Changed

- Added an optional `color` override prop to `MonitorChart` to force a fixed line color instead of per-monitor color assignment
- `LossChart` now uses `var(--destructive)` for its line color


henrygd (approved):
Thank you!

PR #2382, opened 2026-09-21
feat: adds downtime duration to status recovery notifications

spatiumstas:
## 📃 Description

Adds the downtime duration to the "up" notification; the data is generated in the same way as in the web interface.

## 📷 Screenshots

photo_2026-09-22_00-35-38


henrygd:
Thanks, I think it might be better to put this in the body, so the alert would be like:

```md
Connection to Example Agent is up

Downtime duration: 8m

{link}
```

However, there's another huge PR that will probably conflict with this (#2317) so I'm hesitant to merge anything touching notifications until that's sorted out.

PR #2387, opened 2026-09-22
Fix: Convert intel gpu top from text to JSON

lukyjay:
## 📃 Description

Intel GPUs use intel_gpu_top, but we're processing text output. JSON is available so we should use that.

In my installation, Compute is reported as VideoEnhance. This bug is likely because text output is not consistent or reliable.

Changing to JSON makes this more resilient, and fixes the bug.

## 📖 Documentation

No changes proposed.
The documentation does quote the intel_gpu_top command which has been removed, but I suggest that remain in place since it's human readable (unlike JSON).

## 📷 Screenshots

No UX changes required. The back-end change fixes it automatically.

Screenshots below based on an LLM workload on the GPU (triggering 'compute')

Context showing source of truth:
image

Beszel before:
image

Beszel after:
image

## Testing
I do not have an intel dedicated GPU (only my iGPU and some nvidia cards). This change has been tested by my iGPU only. It would be great if someone with a dedicated GPU could also test to confirm the other attributes continue to work (e.g. my iGPU doesn't report power consumption).

## AI Disclosure
I am not familiar with the Go language so I had agents assist with this PR


henrygd:
Thanks, I made some updates. It's a little tricky because the JSON output differs in older and newer versions of `intel_gpu_top`. Would you mind pulling my changes and testing when you have a chance?

PR #2389, opened 2026-09-22
feat: Add option to define which DNS server a DNS monitor queries

### LINKED ISSUE #2381: [Feature]: Dns test against random sub-domains
### Welcome!

### Component

Hub & Agent

### Description

I love the new network features, one thing that makes them less usable is that I cannot easily make it check uncached DNS responses from upstream.

In my scenario, I use DoH, terminated to normal DNS (plus caching) in a container.

If I'd like to monitor actual recursive DNS speed, I'd need to use some sort of cachebuster-subdomain (e.g. xxxxxxxxx.example.org).

### Motivation / Use Case

It allows to monitor actual recursed DNS resolving without caching.

svenvg93:
## 📃 Description

DNS monitors always resolved through the agent's system resolver, with no way to test a specific DNS server. This adds an optional **DNS Server** field to DNS monitors so you can point a check at a particular resolver (e.g. `1.1.1.1` vs `8.8.8.8` vs an internal DNS server) instead of only ever testing whatever the host's OS resolver picks.

Leaving the field empty preserves the current behavior (system resolver), so this is fully backwards compatible with existing DNS monitors.

closes #2381

### ➕ Added

- Optional "DNS Server" field for DNS network monitors, letting you specify which resolver to query (`host` or `host:port`, defaults to port 53 when no port is given)
- Bulk Add now accepts a 5th CSV column for the DNS server: `target[,protocol[,port[,interval[,server]]]]`
- The monitor detail sheet shows the configured DNS server (with a server icon) alongside the protocol, next to where TCP monitors show their port

### ✏️ Changed

- Two DNS monitors on the same target with different servers are now treated as distinct monitors (same treatment TCP monitors already get for different ports)

## 📷 Screenshots

Scherm­afbeelding 2026-09-22 om 17 22 57


henrygd (approved):
Thank you!

PR #2390, opened 2026-09-22
fix(ui): reflect target response status in network monitor status dot

### LINKED ISSUE #2386: [Bug]: Network Monitor always displays a green status indicator, regardless of the target’s actual status.
### Welcome!

### Component

Hub

### Problem Description

I find the green dot next to the monitored target in the Network Monitor a bit misleading, as it suggests that the target is valid and online.
Image

### Expected Behavior

Change the color of the dot, depending on if there is a response.

### Steps to Reproduce

1. Go to Hub 
2. Add a Network Monitor with a nonsense target

### OS / Architecture

linux/amd64

### Beszel version

0.20.0

### Installation method

Docker

### Configuration

```yaml

```

### Hub Logs

```json

```

### Agent Logs

```shell

```

svenvg93:
## 📃 Description

Fixes the network monitor status dot always showing green regardless of whether the target actually responds. The dot next to a target in the Network Monitors table only reflected the parent system's status (Up/Down/Pending/Paused), so a monitor with an unreachable or nonsense target still showed green as long as the hub system it ran on was up. The dot now also factors in the monitor's own `loss1h`/`res` data, matching the thresholds already used by the "Loss 1h" column.

Closes #2386

### 🔧 Fixed

- Network Monitor target status dot now turns red/yellow based on the monitor's own packet loss instead of always showing green when the parent system is up.

## 📷 Screenshots
Scherm­afbeelding 2026-09-22 om 20 36 01
Scherm­afbeelding 2026-09-22 om 20 35 22


henrygd (approved):
Thanks for the fix! I made a small adjustment so the target dot reflects whether the monitor has a current response, rather than using the 1h packet-loss thresholds.

It now turns red when the monitor has already reported data but has no response, while new monitors stay neutral until their first result comes in.

PR #2391, opened 2026-09-22
fix(ui): preserve paused systems in network monitors table

svenvg93:
## 📃 Description

Pausing a system was clearing its entire `info` record, including the agent version. The network monitors table uses that version to decide whether a system supports network monitoring (added in agent 0.20.0), so once a system was paused its monitors would silently disappear from the table instead of just being marked inactive.

This PR preserves the agent version when clearing `info` on pause, and mutes (dims) the monitor's row in the network monitors table when its system is paused, matching the existing behavior in the main systems table.

## 🔧 Fixed

- Pausing a system no longer hides its rows from the network monitors table
- Network monitors table now visually mutes rows belonging to paused systems, consistent with the systems table


henrygd (approved):
Looks good, thanks!

PR #2392, opened 2026-09-22
fix(ui): disable Add Monitor button when no eligible systems exist

svenvg93:
## 📃 Description

The "Add Monitor" button (and its "Bulk Add" menu item) on the Network Monitors page was always clickable, even when there were no systems eligible to attach a monitor to. Clicking it opened a dialog with an empty/unusable system picker, and the mistake was only caught at submit time. This disables the trigger button up front and adds a tooltip explaining why.

## 📖 Documentation

N/A

### 🔧 Fixed

- "Add Monitor" and "Bulk Add" are now disabled when no systems support network monitors (agent ≥ 0.20.0), instead of silently failing after opening the dialog
- Added a tooltip ("Add a system first") explaining why the buttons are disabled in that state

## 📷 Screenshots

Scherm­afbeelding 2026-09-22 om 20 49 42


henrygd (approved):
Thanks Sven! I did remove the tooltip because a user could already have systems, but all of them could be running agents older than 0.20.0. I think the tooltip is not totally necessary, and removing it is cleaner than writing a longer explanation.

